Issue #30, opened 2021-03-24
Programatically validate the form


vsgoulart:
__Is your feature request related to a problem? Please describe__

In order to provide some feedback to the user by disabling the Task completion button after he/she claims a Task I would like to programmatically validated the form or to be able to retrieve the form errors even though they're not yet visible to the user.

__Describe the solution you'd like__

```js
const form = createForm({
  // ..
});

/*
  Option 1 - Programmatically validating the form:
  Calling this method triggers all validation functions
  and show the errors on the form
*/

form.validate();

// or

/*
  Option 2 - Retrieving the errors:
  The getState() method returns the validation errors
  even if the errors are not yet visible to the user
*/

const {errors} = form.getState();
```

__Describe alternatives you've considered__

Do nothing and just wait for the user to change the form or trigger the submission

__Additional context__

As we talked on this meeting this the reason for these features might not be desirable and I need to sync with the design team, but I'm creating the issue to register the discussion.

PR #36, opened 2021-03-25
feat: allow manual form validation

nikku:
Closes https://github.com/bpmn-io/form-js/issues/30


nikku:
Implemented via https://github.com/bpmn-io/form-js/pull/83

PR #37, opened 2021-03-25
Simplify editor properties

nikku:
Applies a number of simplifications to the properties panel.

---

**Screenshot**

image

---

Out of scope: `Key`, still being finalized.


philippfromme (approved):
👏🏻

PR #69, opened 2021-05-31
Add exporter and schemaVersion

nikku:
Closes https://github.com/bpmn-io/form-js/issues/9.

PR #71, opened 2021-05-31
Add viewer and editor UMD distributions.

nikku:
Makes for simple consumption of the used functionality.

---

Depends on https://github.com/bpmn-io/form-js/pull/69


nikku:
Closing this in favor of UMD distributions provided via `@bpmn-io/form-js` directly.

PR #88, opened 2021-06-17
test: verify import error during instantiation

nikku:
This ensures our story for handling import errors works, specifically the following:

```javascript
try {
  await createForm({ schema: someBrokenSchema });
} catch (error) {
  // blows up on error
}
```


philippfromme (approved):
👏🏻

Issue #87, opened 2021-06-17
Styling issues for rendered markdown


andreasgeier:
__Describe the Bug__

There exists a number of minor styling issues for markdown rendering:

* [x] `<br />` tags are inserted, preventing a proper paragraph layout :arrow_right: users should be able to structure text in multiple paragraphs (https://github.com/bpmn-io/form-js/pull/93)
* [x] **Demo only:** ~~links are not underlined :arrow_right: should be underlined~~
* [x] Images are placed in original size which may break the form layout :arrow_right: should apply a maximum width to images so that these do not break the layout (https://github.com/bpmn-io/form-js/pull/92)
* [x] Inconsistent spacing between paragraphs and lists :arrow_right: should have consistent spacing. (https://github.com/bpmn-io/form-js/pull/93)
* [ ] Unordered and ordered lists have different spacing on the left :arrow_right: should be aligned

__Steps to Reproduce__

- Open the [form-js demo](https://demo.bpmn.io/form)
- Open [test.form file](https://github.com/bpmn-io/form-js/files/6675309/text.form.txt)
- Enter markdown in textarea on properties panel to see the results mentioned in the bug description

__Expected Behavior__

See above.

__Context__

For reference a screenshot highlighting above issues:

![image](https://user-images.githubusercontent.com/58601/122523812-ba19a480-d017-11eb-96be-2fd025b9bd51.png)

---

For reference a screenshot with a HTML page:

![image](https://user-images.githubusercontent.com/58601/122526414-8f7d1b00-d01a-11eb-9525-3fdaa467adc7.png)

nikku:
For reference could you share the form file you used to reproduce this?

andreasgeier:
> For reference could you share the form file you used to reproduce this?

No, sorry. I hacked that using the form-js demo and the sandbox. But each topic is very easy to reproduce with just a small snippet of markdown.

nikku:
I'll take this as an opportunity to add a preview to our demo, and provide a diagram that allows us to reproduce the issue.

nikku:
Was able to reproduce most issues.

nikku:
@andreasgeier I've added plain HTML starters that allow you to use form-js in a raw manner: https://github.com/bpmn-io/form-js-examples/pull/1. The link styling issue you raised is due to style overrides in the bpmn.io demo (nothing form-js can do about it).

Will follow up with a preview on the demo, at some point.

nikku:
[...]

PR #92, opened 2021-06-18
fix: render images with max-width=100%

nikku:
Ensures embedded images do not break the form layout.

capture jZPiKi_optimized

---

Related to https://github.com/bpmn-io/form-js/issues/87


philippfromme (approved):
🆒

andreasgeier (approved):
✅ &nbsp; LGTM, max-width should do the job

PR #93, opened 2021-06-18
fix: properly generate paragraphs

nikku:
Should generate proper paragraphs from markdown now as the user would expect:

image

image

Had to workaround https://github.com/developit/preact-markup/issues/10 as well as https://github.com/developit/snarkdown/issues/75. Added a special rule to not wrap existing HTML tags (i.e. user-created `<p>` tags).

---

Related to https://github.com/bpmn-io/form-js/issues/87

### DIFF
--- a/packages/form-js-editor/test/spec/form.json
+++ b/packages/form-js-editor/test/spec/form.json
@@ -2,7 +2,7 @@
   "components": [
     {
       "type": "text",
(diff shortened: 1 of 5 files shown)


nikku:
Ready to review @philippfromme.

philippfromme (commented):
👏🏻

philippfromme (approved):

PR #130, opened 2021-07-21
feat(editor): add compact mode

nikku:
This allows users to configure a compact editor rendering mode:

screely-1626866387466

In compact mode the editor is geared towards smaller screen widths, compresses the palette and potentially provides further improvements in the future.

This approach does not add true responsiveness. Instead, integrators must decide on instantiation whether the editor should be rendered compact or not. An alternative would be to provide additional APIs to trigger that behavior / make it reactive via a generic _set property_ interface (cf. https://github.com/bpmn-io/form-js/blob/22a9ac92bc45bcb309df04770fa0fc004e49a628/packages/form-js-viewer/src/Form.js#L61).

### DIFF
--- a/packages/form-js-editor/assets/form-js-editor.css
+++ b/packages/form-js-editor/assets/form-js-editor.css
@@ -345,3 +345,23 @@
   pointer-events: none !important;
   height: 54px !important;
 }
+
+/**
+ * Compact editor overrides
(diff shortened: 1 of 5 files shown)


philippfromme (approved):
Works as expected. 👏🏻

PR #131, opened 2021-07-21
Improve playground, I/O wise

nikku:
This PR incorporates a number of significant improvements into the playground, mainly focused on getting things into the tool, and out again.

screely-1626856096544

### Added Features

* Add download, embed and file drop features
* Incorporate compact editing mode (https://github.com/bpmn-io/form-js/pull/130)
* Prepare for publishing and embedding (document, bundle, remove `private` flag)

### Try it out!

```sh
npx @bpmn-io/sr bpmn-io/form-js#playground-next -c "npm run start:playground"
```


packages/form-js-playground/src/Playground.js
```diff
@@ -34,12 +153,26 @@ function AppRoot(props) {
...
+  const [ data, setData ] = useState(props.data || {});
   const [ schema, setSchema ] = useState(props.schema);
 
   const [ resultData, setResultData ] = useState(props.data || {});
 
+  useEffect(() => {
+    props.onInit({
+      setSchema: setInitialSchema
+    });
+  });
```
pinussilvestrus, line 170: Is this effect intended to run _on every_ render (because it's missing dependencies)? Setting the initial schema could maybe be a first render task.

packages/form-js-playground/src/Playground.css
```diff
@@ -1,73 +1,158 @@
-.app-root {
+.fjs-pgl-root {
```
pinussilvestrus, line 1: Out of curiosity: what does `pgl` mean? I first thought it stands for "playground", but then it would rather name `.fjs-plg-*`.

packages/form-js-playground/rollup.config.js
```diff
@@ -0,0 +1,60 @@
...
+      plugins: [
+        [ '@babel/plugin-transform-react-jsx', {
+          'importSource': 'preact',
+          'runtime': 'automatic'
+        } ]
+      ]
+    }),
+    css({
+      output: 'assets/form-js-playground.css'
+    }),
```
pinussilvestrus, line 23: Nice ❤️

pinussilvestrus (commented):
That's really really cool 👏   I especially like the `embed` feature to quickly download and use my generated form. ❤️ 

I found a minor issue with the preview form related to the form component labels. When clicking one label, it does not control the input
Kapture 2021-09-24 at 11 23 06

In comparison, this does work in the downloaded form snippet

Kapture 2021-09-24 at 11 23 41

packages/form-js-playground/src/Playground.js
nikku: To my knowledge this is a once-only hook. But I'll investigate.

packages/form-js-playground/src/Playground.css
nikku: Good catch. Needs fix.

nikku:
Thanks for your review. I'll incorporate your comments.

The label thing is broke right now due to https://github.com/bpmn-io/form-js/issues/180, i.e. it needs an upstream fix that I do not have time for right now.

barmac (approved):
🟢  from my side.

nikku:
Let us get this out of the door then.

PR #132, opened 2021-07-21
feat(editor): debounce input per default

nikku:
This ensures that we debounce editor input per default, unless disabled.

This is standard behavior in editing environments and allows users to get good UX without further configuration.


philippfromme (approved):
👏🏻

Issue #141, opened 2021-07-28
Fix `POST_RELEASE` build


nikku:
__What should we do?__

Ensure the [`POST_RELEASE` build](https://github.com/bpmn-io/form-js/runs/3181561121?check_suite_focus=true) works, and works reliably:

* [ ] Correct demo update failing due to `npm install` failing :arrow_right: need to add a retry mechanism as it sometimes takes a minute or two for a published package to be discovered.
* [ ] Correct examples update failing

__Why should we do it?__

Preventing unneeded manual labor is why we've set it up in the first place. Fix it.

PR #151, opened 2021-08-20
ci: wait for published before executing post release steps

nikku:
Basically the idea is to wait in a loop for npm to make published packages available, which may take some time.

Maximum time we wait is 2:30 min.

Closes https://github.com/bpmn-io/form-js/issues/141


barmac (approved):
I've tested this manually and it works correctly. Well done.

PR #153, opened 2021-08-20
Properly update element ID in model and registry

nikku:
Closes https://github.com/bpmn-io/form-js/issues/152


MaxTru:
I can now successfully change the ID of an form and then change components on that form. But having done that, `undo` does not work correctly anymore:

* run `npm run start:editor`
* Change id of form
* Click on one component
* Change `key` of that component
* Perform `undo` multiple times
* Error occurs:
Screenshot_20210823_153337

MaxTru (changes requested):
See https://github.com/bpmn-io/form-js/pull/153#issuecomment-903773819

pinussilvestrus (commented):
The implementation looks good to me 👍  

I was able to reproduce the same as mentioned in this comment: https://github.com/bpmn-io/form-js/pull/153#issuecomment-903773819. Not only with the `key` property, but with any property for a form field after editing the parent forms ID. Without editing the ID beforehand, everything works perfectly fine.

nikku:
I can reproduce the following errors:

```
Uncaught TypeError: Cannot read property 'value' of null
    at Object.eval (webpack-internal:///./src/render/components/properties-panel/components/TextInput.js:31)
    at fire (webpack-internal:///../../node_modules/min-dash/dist/index.esm.js:450)
mocha.js:24291 Uncaught Error: Uncaught TypeError: Cannot read property 'value' of null (webpack-internal:///./src/render/components/properties-panel/components/TextInput.js:31)
    at commonjsGlobal.onerror (mocha.js:30195)
```

I'd presume these are the ones in the screenshot? Will need to investigate. Looks like an unrelated bug in our debounce logic, that simply surfaces now.

MaxTru:
> I can reproduce the following errors:
> 
> ```
> Uncaught TypeError: Cannot read property 'value' of null
>     at Object.eval (webpack-internal:///./src/render/components/properties-panel/components/TextInput.js:31)
>     at fire (webpack-internal:///../../node_modules/min-dash/dist/index.esm.js:450)
> mocha.js:24291 Uncaught Error: Uncaught TypeError: Cannot read property 'value' of null (webpack-internal:///./src/render/components/properties-panel/components/TextInput.js:31)
>     at commonjsGlobal.onerror (mocha.js:30195)
> ```
> 
> I'd presume these are the ones in the screenshot? Will need to investigate. Looks like an unrelated bug in our debounce logic, that simply surfaces now.

Yep, that is the one.

PR #157, opened 2021-08-27
Debounce debounce

nikku:
Some internal cleanups:

* Drop use of services within components but rely on hooks instead
* Prepare for cancelable debounced functions, to be added one day, maybe (https://github.com/bpmn-io/min-dash/pull/20)

### DIFF
--- a/packages/form-js-editor/src/render/hooks/useDebounce.js
+++ b/packages/form-js-editor/src/render/hooks/useDebounce.js
@@ -1,0 +1,25 @@
+import {
+  useMemo,
+  useEffect
+} from 'preact/hooks';
+
+import useService from './useService';
+
+
+export default function useDebounce(fn, dependencies=[]) {
+
+  const debounce = useService('debounce');
+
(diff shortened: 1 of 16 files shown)


packages/form-js-editor/src/render/hooks/useDebounce.js
```diff
@@ -0,0 +1,25 @@
...
+import useService from './useService';
+
+
+export default function useDebounce(fn, dependencies=[]) {
+
+  const debounce = useService('debounce');
+
+  const callback = useMemo(() => {
+    return debounce(fn);
+  }, [...dependencies, fn ]);
```
barmac, line 15:
```suggestion
  }, dependencies);
```

barmac: Otherwise it's flushed on each render if `fn` is defined within the render function.

barmac (approved):
I love it.

nikku:
@philippfromme Please have a look.

Issue #170, opened 2021-09-13
Integrate form icon


nikku:
__What should we do?__

Add [form icon](https://github.com/camunda/camunda-modeler-design/issues/28#issuecomment-916898760).

__Why should we do it?__

Make form a distinguishable entity.

PR #172, opened 2021-09-13
fix: correct form icon

nikku:
Closes https://github.com/bpmn-io/form-js/issues/170

----

Look and feel demo:

capture LjAhCN_optimized

### DIFF
--- a/packages/form-js-editor/src/render/components/palette/icons/Form.svg
+++ b/packages/form-js-editor/src/render/components/palette/icons/Form.svg
@@ -1,4 +1,4 @@
-<svg width="36" height="36" viewBox="0 0 54 54" xmlns="http://www.w3.org/2000/svg">
+<svg xmlns="http://www.w3.org/2000/svg" width="54" height="54" viewBox="0 0 54 54">
    <rect x="15" y="17" width="24" height="4" rx="1"/>
    <rect x="15" y="25" width="24" height="4" rx="1"/>
    <rect x="15" y="33" width="13" height="4" rx="1"/>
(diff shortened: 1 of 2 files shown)


MaxTru (approved):

nikku:
Note to design review: We currently work with a `54x54` icon grid in form-js land. 

Icon size shall be cleaned up / aligned once we migrate to the new properties panel infrastructure.

andreasgeier:
Pleas use this updated icon in a 54x54 viewbox: https://github.com/camunda/camunda-modeler-design/blob/main/icons/entities/forms-54x54.svg?short_path=d53b419

nikku:
> Pleas use this updated icon in a 54x54 viewbox: https://github.com/camunda/camunda-modeler-design/blob/main/icons/entities/forms-54x54.svg?short_path=d53b419

We're already using a scaled up version. Which visual effect would you like to accomplish?

andreasgeier:
I quickly pasted the icon (32x32 size in a 54x54 viewbox) on a screenshot and this is the result:

Screen Shot 2021-09-15 at 10 04 35

Please check, your implementation seems to be different.

nikku:
To properly integrate with our existing icons the form icon must be "54x54" in size on a "0 0 54 54" viewbox. We're doing some magic tricks to translate that image down to "36x36" dimensions in the panel.

This way it properly integrates with the [existing icons](https://github.com/bpmn-io/form-js/tree/master/packages/form-js-editor/src/render/components/palette/icons), for example [this one](https://github.com/bpmn-io/form-js/blob/master/packages/form-js-editor/src/render/components/palette/icons/Button.svg). To achieve the visual result you desire (https://github.com/bpmn-io/form-js/pull/172#issuecomment-919794464) the icon would need to have less padding.

### For Comparison

| your icon | existing icons |
| :--- | :--- |
| this is how your icon looks like | other icon another icon |

andreasgeier (approved):
I think the problem here is the "magic" of how this is implemented. In the future, we should avoid resizing the symbols outside of the canvas, otherwise, it doesn't make sense to provide pixel-sharp variants in different sizes.

I will accept your implementation as an intermediate step until we replaced this panel with the new one.

nikku:
For further clarification, this is the space we have available: 

image

PR #174, opened 2021-09-15
Fix CI scripts

nikku:
This should resolve our `POST_RELEASE` CI failures, e.g. [this one](https://github.com/bpmn-io/form-js/actions/runs/1198257674).

---

* Move `await-published` to own task (https://github.com/bpmn-io/form-js/commit/6d7a2cb71cdc1a68acab3f1b93787e47924731c5)
* Fix `await-published` to properly increment on wait
* Correct `update-example` and `update-website` hooks to work with strict error handling

### DIFF
--- a/.github/workflows/POST_RELEASE.yml
+++ b/.github/workflows/POST_RELEASE.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Wait for published
       env:
         PKG: '@bpmn-io/form-js@${{ env.TAG }}'
-      run: bash tasks/await-published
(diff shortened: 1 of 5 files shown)


nikku:
I'll port these changes over to our other toolkits if we find out that things work at `form-js`.

barmac (approved):
Looks good. I think it's much cleaner now.

nikku:
It actually works, too: https://github.com/bpmn-io/form-js/runs/3609845602?check_suite_focus=true

image

PR #175, opened 2021-09-15
fix(viewer): always render label

nikku:
Prevents components wrapped in labels and the required indicator from disappearing.

Closes https://github.com/bpmn-io/form-js/issues/168

----

capture hbE7uD_optimized


philippfromme (approved):
Very nice! 👏🏻

PR #179, opened 2021-09-16
Disallow toggling of `<radio />` element

### LINKED ISSUE #176: Do not allow unchecking of `<radio>` options
__Describe the Bug__

It is not in the nature of a `<radio>` element to be unchecked. To get that behavior users shall use a select (dropdown) or a list of checkboxes. [See also MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio).

Because of that I believe that radio should not have the check / uncheck behavior they currently exibit:

capture 31l8rY_optimized

__Steps to Reproduce__

1. Open form with a `<radio>` element defined, e.g. by executing `npm run start:viewer` in this repository
2. See the behavior displayed above

__Expected Behavior__

Radio options cannot be unset, unless a user explicitly resets the form.

__Environment__

- Host (Browser/Node version), if applicable: any
- OS: any
- Library version: v0.4.3

nikku:
There exist other form elements that can be used if non-radio / checkbox style interaction is desired.

capture kqPrHs_optimized

---

Closes #176

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Radio.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Radio.js
@@ -31,7 +31,7 @@ export default function Radio(props) {
   const onChange = (v) => {
     props.onChange({
       field,
-      value: v === value ? undefined : v
+      value: v
     });
   };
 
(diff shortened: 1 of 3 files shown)


andreasgeier (approved):
✅ &nbsp; LGTM

PR #202, opened 2021-11-15
chore(properties-panel): use more specific selectors

pinussilvestrus:
Instead of using the `fjs-properties-panel-textfield` class for all properties panel inputs, let's use more specific ones. That would make it easier to override certain field styles (e.g. during https://github.com/camunda/camunda-modeler/issues/2460).

We don't use `fjs-properties-panel-textfield` anywhere so far, so this change shouldn't break anything.


nikku (approved):
Makes a lot of sense.

PR #231, opened 2022-03-08
Initial implementation of File Form Field

### LINKED ISSUE #222: Add file upload and download capability
In general, I need to be able to both upload and download files from the Camunda database. This is currently supported by Camunda's embbedded forms capability. See 

https://docs.camunda.org/manual/latest/reference/forms/embedded-forms/controls/files/

To implement this, I would imagine that you would add a field type of **file**.  

If the field **is not** marked **readonly**, the user would be able to specify a file to upload. If a file has already been uploaded with the provided field name, then the user would be able to upload a new file.

If the field **is** marked **readonly**, then the user would only be able to download the previously uploaded file.

In my opinion, this is a critical feature - critical for me anyway. I've used this embedded forms capability often.

Note: the last I checked, a **readonly** property was not yet supported, but I was told by another Camunda forum user, that the feature had been implemented and would be included in the next Camunda release. Do you know if this is true? I think that **readonly** fields are an even more critical feature.

dfulgham:
Initial implementation of File field type

Implements and closes #222 

Implements a file input type using base64 encoding of file.


MaxTru:
Thanks @dfulgham for investing into this and this cool contribution.

Unfortunetly I have to reject this, since the solution does conceptually not fit into what we envision. We would discourage to store business data in the JSON itself, but more consider an approach, where files are uploaded to some storage and then logically references (for example: how images in GitHub comments are added). Also see [this comment](https://github.com/bpmn-io/form-js/issues/222#issuecomment-1068054359).

It would be great to understand from you, why you built this? Are you requiring this feature for your own custom application? If so, would you for example require a way to extend form-js with custom components?

dfulgham:
Hello @MaxTru 

I believe it should be up to the client application that is using the form.js library to determine how to use the javascript object submitted from the form, ie. whether to store the file property into a storage database rather than keep it as part of the JSON that might eventually go to a server or Camunda for example.  

I think that the form.js library should remain client/server solution agnostic as it is simply a javascript form library.

For example, what if I wanted to only process the file with an OCR library on the client before then sending the data to a database?  Having the business logic built into the form library would be too restrictive.

If needed I could have an option on the file component that would allow the object to be stored as Base64 encoding or a Blob.

For further debate, this is how Camunda handles File-Type Variables, A Form can submit them into process variables as Blobs. (https://docs.camunda.org/manual/7.14/user-guide/process-engine/variables/#file-values).

PR #239, opened 2022-03-18
feat: Provide API to the editor to decide which form field component …

### LINKED ISSUE #237: Provide API to the editor to decide which form field/component types are exposed in form creation

To allow extensibility in the core of form.js, the embedding client needs control over which form fields can be used when editing and creating forms. This allows a stable set of exposed form field types even when new field types are added to form.js.

This enables innovating on form.js by adding new components without the risk of exposing unwanted components to already embedding clients on a version bump.

On form editor instantiation, the developer should be able to provide a whitelist of form field/component types that should be exposed in the editor.

Something like
```js
const formEditor = new FormEditor({
  container: document.querySelector('#form-editor'),
  availableComponentTypes: ['textfield', 'number',  'checkbox', ...]
});
```

dfulgham:
Adds an API to restrict which component are allowed in the editor, handles issues when unallowed components are present in the schema.

Closes #237

### DIFF
--- a/packages/form-js-editor/src/FormEditor.js
+++ b/packages/form-js-editor/src/FormEditor.js
@@ -3,11 +3,11 @@ import Ids from 'ids';
 import { isString, set } from 'min-dash';
 
 import core from './core';
-
 import EditorActionsModule from './features/editor-actions';
 import KeyboardModule from './features/keyboard';
 import ModelingModule from './features/modeling';
 import SelectionModule from './features/selection';
+import { ALL_COMPONENTS } from './render/components/properties-panel/Util';
 
 const ids = new Ids([ 32, 36, 1 ]);
 
@@ -15,12 +15,18 @@ const ids = new Ids([ 32, 36, 1 ]);
  * @typedef { import('./types').Injector } Injector
  * @typedef { import('./types').Module } Module
  * @typedef { import('./types').Schema } Schema
- *
+ * @typedef { import('./types').ComponentTypes} ComponentTypes
+ * @typedef { import('./types').ComponentActions} ComponentActions
  * @typedef { import('./types').FormEditorOptions } FormEditorOptions
  * @typedef { import('./types').FormEditorProperties } FormEditorProperties
  *
+ *
+ *
+ *
  * @typedef { {
  *   properties: FormEditorProperties,
+ *   availableComponentTypes?: ComponentTypes[],
+ *   unavailableComponentAction: ComponentActions,
  *   schema: Schema
  * } } State
  *
@@ -28,6 +34,7 @@ const ids = new Ids([ 32, 36, 1 ]);
  * @typedef { (type:string, handler:Function) => void } OnEventWithOutPriority
  * @typedef { OnEventWithPriority & OnEventWithOutPriority } OnEventType
  */
+
 
 /**
  * The form editor.
@@ -60,8 +67,13 @@ export default class FormEditor {
 
     this._container.setAttribute('input-handle-modified-keys', 'z,y');
 
+    /**
+     * @type {FormEditorOptions}
+     */
     const {
       container,
+      availableComponentTypes=[],
+      unavailableComponentAction,
       exporter,
       injector = this._createInjector(options, this._container),
       properties = {}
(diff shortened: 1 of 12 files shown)


dfulgham:
How can I build my code coverage in my local environment so I can see where I am missing testing.

dfulgham:
@nikku I need some help with getting my local ENV setup for code coverage and get my tests written, if you have sometime to assist me I think I can get the coverage up and get this ready to merge?

nikku:
Happy to give you some insights into test coverage.

The following executes our tests with coverage enabled:

```
COVERAGE=1 npm test
```

I've created a PR that allows you to generate local test coverage, too (https://github.com/bpmn-io/form-js/pull/250). This will make the coverage report available in each packages `coverage` folder.


----

Generally speaking though we should aim to simply build test coverage that make sense (tests new features in a logical fashion).

What is more important for us is to test integration style (end-to-end, against real APIs, not mock components) over unit tests.

So what you'd need to do is to add test coverage:

* Verify viewer API + behavior (disallowed component => blow up during import)
* Verify modeler API + behavior (disallowed component => allow component to be removed but no more)
* No tests / modifications for playground should be needed I believe.

packages/form-js-editor/src/render/components/properties-panel/groups/UnallowedGroup.js
```diff
@@ -0,0 +1,20 @@
+import { Group } from '../components';
```
nikku, line 1: Cf. https://github.com/bpmn-io/form-js/pull/239/files#r869039594

packages/form-js-playground/src/Playground.js
```diff
@@ -1,22 +1,14 @@
-import JSONView from './JSONView';
```
nikku, line 1: No change shall be necessary here I believe?

packages/form-js-editor/test/spec/properties-panel/PropertiesPanel.spec.js
```diff
@@ -74,6 +66,39 @@ describe('properties panel', function() {
     expect(result.container.querySelectorAll('.fjs-properties-panel-group')).to.have.length(3);
   });
 
+  it('should render (readonly)', async function() {
```
nikku, line 69: Add a dedicated `describe` for the `component whitelist` feature.

Test

* All existing components can be addressed via well defined keys (to remove them)
* Import behavior on unknown component

nikku, line 69: Test cover on the viewer: 

* Import behavior (blow up on unsupported schema with a reasonable error message)

nikku (commented):
I think this goes into a good direction. Added a bunch of hopefully helpful hints.

packages/form-js-editor/src/FormEditor.js
```diff
@@ -3,24 +3,30 @@ import Ids from 'ids';
...
+ * @typedef { import('./types').ComponentActions} ComponentActions
  * @typedef { import('./types').FormEditorOptions } FormEditorOptions
  * @typedef { import('./types').FormEditorProperties } FormEditorProperties
  *
+ *
+ *
+ *
  * @typedef { {
  *   properties: FormEditorProperties,
+ *   availableComponentTypes?: ComponentTypes[],
```
nikku, line 28: My proposal is to simplify this API to `allowedComponents=string[]|(name:string) => boolean`.

nikku: Or for the initial case simply start with `allowedComponents=string[]`.

packages/form-js-editor/src/render/components/properties-panel/groups/ReadOnlyGroup.js
```diff
@@ -0,0 +1,20 @@
+import { Group } from '../components';
```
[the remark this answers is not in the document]
nikku: By bad, I think I got it. This is what we're showing to the user as a  "not allowed" placeholder, right?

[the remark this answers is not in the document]
dfulgham: Yes.

packages/form-js-playground/src/Playground.js
dfulgham: I'll revert this file, was just using it for testing the feature.

PR #245, opened 2022-04-13
fix(ValidationGroup): serialize regex pattern as <pattern>

MaxTru:
* In accordance with schema https://github.com/bpmn-io/form-js/blob/master/docs/FORM_SCHEMA.md?plain=1#L32
* In accordance with existing test cases https://github.com/bpmn-io/form-js/blob/master/packages/form-js-viewer/test/spec/core/Validator.spec.js#L24
* Fixes upstream camunda/camunda-modeler#2919

### DIFF
--- a/packages/form-js-editor/src/render/components/properties-panel/entries/DisabledEntry.js
+++ b/packages/form-js-editor/src/render/components/properties-panel/entries/DisabledEntry.js
@@ -7,7 +7,7 @@ export default function DisabledEntry(props) {
   } = props;
 
   const onChange = (value) => {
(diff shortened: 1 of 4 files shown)


nikku (changes requested):
I can confirm your finding. Fix wise the PR is :medal_sports:.

However, we're clearly missing test coverage here to verify that this works. The reason we were able to introduce this BUG is missing test coverage in the first place.

My proposal is: Let us add coverage (integration test style) to verify proper serialization on change, and proper read (on import).

philippfromme (approved):
✅

PR #248, opened 2022-05-05
chore(package): add `build:watch` script

pinussilvestrus:
In case you work in a specific library.

```sh
npm run build-watch -- @bpmn-io/form-js-editor
npm run build-watch -- @bpmn-io/form-js-viewer
npm run build-watch -- @bpmn-io/form-js-playground
```

Or to build (watch) the whole library:

```
npm run build:watch
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "scripts": {
     "all": "run-s lint build lint:types test test:distro",
     "build": "lerna run build --sort --stream",
+    "build:watch": "npm run build-watch -- @bpmn-io/form-js",
     "build-distro": "lerna run prepublishOnly --sort",
     "build-watch": "run-s \"lerna-build-watch -- {@}\" -- ",
     "clean": "del-cli \"{node_modules,dist}\" \"packages/*/{node_modules,dist}\"",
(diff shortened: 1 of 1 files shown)


nikku:
What is the use-case if we have `npm run start:xxx`?

pinussilvestrus:
The use case would be if I linked `form-js` to another project (e.g. Camunda Modeler), I don't want to build all the packages but only the one where I am making changes (+ hot reloading). The same goes if I run the playground ( `npm run start:playground`).

`npm run start:xxx` would work to, but I don't want to have karma running, only bundling my changes.

nikku:
> The use case would be if I linked form-js to another project (e.g. Camunda Modeler), I don't want to build all the packages but only the one where I am making changes (+ hot reloading). The same goes if I run the playground ( npm run start:playground).

I'd personally vote for having a global `build:watch`. Otherwise it is too easy to not pick up changes across sub-projects (i.e. wonder why the playground does not update as I fix stuff in the form-js-editor). 

But that is my two cents only :).

nikku:
I've added a global build:watch script via https://github.com/bpmn-io/form-js/pull/248/commits/e906a3e662cbc7b5271e6cb88f8373beed595cf1.

Lets merge this.

pinussilvestrus:
Thank you 🚀

Issue #123, opened 2021-07-06
Support for custom elements


vasicvuk:
### Is your feature request related to a problem? Please describe.

As a developer, it would be beneficial to have the ability to add custom elements to the palette in both the designer and viewer interfaces. Incorporating an upload component with custom behavior would be an example use case.

### Describe the solution you'd like

Ideally, a library consumer should be able to override the element registry, and hook their own element types into it. We would then expect to be able to define the structure required for the properties panel (and eventually a schema for component validation) as part of our new component implementation. 

### Describe alternatives you've considered

n/a

___

This is a pretty large scale refactoring, and it requires quite a few steps:

- [ ] Allow overriding the element registry
- [ ] Hook our palette into the element registry
    - [ ] Use the registry as a source instead of statically defining components, or create a modularized registry source
	- [ ] Handle icons, which as of now are statically defined.
- [ ] Extract the properties panel configurations into the components somehow (the biggest problem)
  - [ ] Option: Mirror [bpmn-js-properties-panel providers approach](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/render/BpmnPropertiesPanelRenderer.js#L116)
  - [ ] It should be easy to use existing properties panel entries
- [ ] Create a plan to gather feedback around the extension points
  - [ ] Idea: create [custom form playground demo](https://camunda-form-playground.netlify.app/) to act as a sandbox, e.g. via uploading my extension into the editor
  - [ ] Idea: use [hackdays project](https://github.com/pinussilvestrus/form-field-extension)
- [ ] Communicate how to create custom form elements
  - [ ] Cf. How we do this in [bpmn land via examples](https://github.com/bpmn-io/bpmn-js-example-custom-elements)
  - [ ] E.g. Add guide in our docs (especially a reference of `FormField.config` options)
  
Cf. the [recent spike](https://github.com/bpmn-io/form-js/pull/776) we conducted

![Image](https://github.com/camunda/team-hto/assets/9433996/d79c85fe-e862-4979-a7d2-887b7303bdc9)
![Image](https://github.com/camunda/team-hto/assets/9433996/6b79cbba-667c-49ee-93c0-70c61a1132ca)

nikku:
Thanks for opening this issue. Custom elements are on our mid-term roadmap.

VictorAITERA:
Hi, is there a way to develop our own component and add it to the left panel (FORM ELEMENTS LIBRARY) ?

[...]

PR #250, opened 2022-05-10
chore: allow local HTML coverage reports to be generated

nikku:
Related to https://github.com/bpmn-io/form-js/issues/123#issuecomment-1075594064.

----

You can now generate local HTML coverage reports via 

```
COVERAGE=1 npm test
```

### DIFF
--- a/packages/form-js-editor/karma.conf.js
+++ b/packages/form-js-editor/karma.conf.js
@@ -37,7 +37,7 @@ module.exports = function(karma) {
 
     coverageReporter: {
       reporters: [
-        { type: 'lcovonly', subdir: '.' },
+        { type: 'lcov', subdir: '.' }
       ]
     },
 
--- a/packages/form-js-playground/karma.conf.js
+++ b/packages/form-js-playground/karma.conf.js
@@ -33,7 +33,7 @@ module.exports = function(karma) {
 
     coverageReporter: {
       reporters: [
-        { type: 'lcovonly', subdir: '.' },
+        { type: 'lcov', subdir: '.' }
       ]
     },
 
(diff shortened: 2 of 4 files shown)


packages/form-js-editor/karma.conf.js
```diff
@@ -37,8 +37,10 @@ module.exports = function(karma) {
 
     coverageReporter: {
       reporters: [
-        { type: 'lcovonly', subdir: '.' },
-      ]
+        { type: 'lcovonly', subdir: '.' }
+      ].concat(process.env.HTML_COVERAGE ? [
+        { type: 'html', subdir: '.' }
+      ] : [])
```
pinussilvestrus, line 43: Maybe to clarify: why do we use `lcovonly` at all? In bpmn-js [we use `lcov`](https://github.com/bpmn-io/bpmn-js/blob/develop/test/config/karma.unit.js#L46) which offers HTML reporting as well. 

Is it to reduce the overhead when dealing with more packages?

pinussilvestrus (commented):
Nice stuff! Just one question for my understanding. Otherwise, let's merge 👍

packages/form-js-editor/karma.conf.js
nikku: No, there is literally no reason, other than that is how we do it.

If lcov also generates HTML reports, let us simply use it :).

nikku: Will double check and update PR accordingly.

nikku: Updated via https://github.com/bpmn-io/form-js/pull/250/commits/c25d14d0231dbea4b4cc08641669867d49b5e837.

PR #254, opened 2022-05-12
Migrate Properties Panel to latest framework version

pinussilvestrus:
Closes #249 

Screenshot of the new Forms Properties Panel

* [x] Integrate `@bpmn-io/properties-panel` framework
* [x] Migrate entries
* [x] Migrate test coverage
* [x] Remove old components (or do it later? breaking change?) 
  * [x] Remove preact components
  * [x] Remove styles
* [x] Consider: Have the properties panel as an own (feature) module (as we have in BPMN and DMN)
  * => proposal: Do not do it now, as it does not add any value to the migration itself. Personally, I like that everything just lives under the same tree, and I do not have to handle around different containers. If we do so, we should handle it in an orderly manner (also considering the Palette) (cf. https://github.com/bpmn-io/form-js/pull/254#issuecomment-1142083791)
  * ==> that might change though (cf. https://github.com/bpmn-io/form-js/pull/254#issuecomment-1135829669)
* [x] Resolve preact dependency (`src/components` vs. bundled components) (cf. https://github.com/bpmn-io/form-js/pull/254#issuecomment-1126029632)

To try out:

a) Playground

```sh
npx @bpmn-io/sr bpmn-io/form-js#249-migrate-properties-panel -c "npm run start:playground" --verbose
```

b) Camunda Modeler (artifacts)
- https://camunda-modeler-on-demand.s3.eu-central-1.amazonaws.com/update-forms-properties-panel/camunda-modeler-update-forms-properties-panel-linux-x64.tar.gz
- https://camunda-modeler-on-demand.s3.eu-central-1.amazonaws.com/update-forms-properties-panel/camunda-modeler-update-forms-properties-panel-mac.dmg
- https://camunda-modeler-on-demand.s3.eu-central-1.amazonaws.com/update-forms-properties-panel/camunda-modeler-update-forms-properties-panel-mac.zip
- https://camunda-modeler-on-demand.s3.eu-central-1.amazonaws.com/update-forms-properties-panel/camunda-modeler-update-forms-properties-panel-win-ia32.zip
- https://camunda-modeler-on-demand.s3.eu-central-1.amazonaws.com/update-forms-properties-panel/camunda-modeler-update-forms-properties-panel-win-x64.zip

PR #255, opened 2022-05-12
feat: support Checklist components

### LINKED ISSUE #196: Support multi-select checkbox

As a form design I can model a component which can have `n` options, where the later form-user is able to select `m` (`0<=m<=n`) entries. On submission, the `m` selected options are submitted as list.

Compare to:
* Radio Button component, but with the ability to select multiple entries (as form user)
* see form.io [`Select Boxes` component](https://help.form.io/userguide/forms/form-components#select-boxes)

image

n/a

__Additional context__

Raised as part of https://jira.camunda.com/browse/SUPPORT-12025 interaction
Child of https://github.com/bpmn-io/form-js/issues/200

Skaiir:
Closes #196 

**Code can be reviewed despite draft**, I'm just keeping this like this so that it doesn't get merged because we're missing an icon.

Note that the omission of `default values` is intentional, I realized this a little late but it would require creating a whole new component for the properties panel which from an effort to value ratio doesn't seem right a the moment. We'll follow this up in the future depending on user feedback, or if we ever build a multi-select component for the properties panel which would make this cheap.

image


pinussilvestrus:
> Note that the omission of default values is intentional, I realized this a little late but it would require creating a whole new component for the properties panel which from an effort to value ratio doesn't seem right a the moment

Makes sense as we anyway will have the properties panel migrated soon (cf. #254). Should we maybe create an issue for the default values to have it tracked?

Skaiir:
@pinussilvestrus Will do

Skaiir:
@andreasgeier Yes I agree, and have already ditched the term for checklist, but checkbox group may be better.

pinussilvestrus:
The [package-lock.json](https://github.com/bpmn-io/form-js/pull/255/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519) looks quite messed. Consider using `npm run reinstall` to hoist it properly 👍

packages/form-js-editor/src/render/components/properties-panel/PropertiesPanel.js
```diff
@@ -35,11 +36,11 @@ function getGroups(field, editField) {
     GeneralGroup(field, editField)
   ];
 
-  if (type === 'radio' || type === 'select') {
+  if (type === 'radio' || type === 'select' || type === 'checklist') {
     groups.push(ValuesGroup(field, editField));
   }
 
-  if (INPUTS.includes(type) && type !== 'checkbox') {
+  if (INPUTS.includes(type) && type !== 'checkbox' && type != 'checklist') {
```
pinussilvestrus, line 43: Could we move these to `Util` as well? I added something similar via the [properties panel migration](https://github.com/bpmn-io/form-js/pull/254/files#diff-f39d8b1f36be0837b1b467eed602c62a2ed76e484e25220efc988f5753d27fc4R45). Saying that, I think it's okay to refactor this via https://github.com/bpmn-io/form-js/pull/254.

packages/form-js-viewer/src/render/components/form-fields/Checklist.js
```diff
@@ -0,0 +1,92 @@
...
+  const {
+    description,
+    id,
+    label,
+    values
+  } = field;
+
+  const toggleCheckbox = (v) => {
+
+    var newValue = [...value];
```
pinussilvestrus, line 34:
```suggestion
    let newValue = [...value];
```

packages/form-js-viewer/src/render/components/form-fields/Checklist.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+  };
+
+  const { formId } = useContext(FormContext);
+
+  return <div class={ formFieldClasses(type, errors) }>
+    <Label
+      label={ label } />
+    {
+      values.map((v, index) => {
```
pinussilvestrus, line 54: Personally, I wouldn't use such small variable names if the variable is heavily used. But this probably comes from that `value` is already given via `prop`.

packages/form-js-viewer/src/render/components/form-fields/Checklist.js
```diff
@@ -0,0 +1,92 @@
...
+    id,
+    label,
+    values
+  } = field;
+
+  const toggleCheckbox = (v) => {
+
+    var newValue = [...value];
+
+    if (!newValue.includes(v)) {
```
pinussilvestrus, line 36: I assume this works because we can rely on that UI & data is always in sync, right? So we will technically never have the case.

* Checkbox is checked
* Value is not in the data

What I mean specifically: We do not react on the `event.target.checked` itself, but only toggling the data.

packages/form-js-viewer/test/spec/render/components/form-fields/Checklist.spec.js
```diff
@@ -0,0 +1,232 @@
...
+      expect(onChangeSpy).to.have.been.calledWith({
+        field: defaultField,
+        value: []
+      });
+    });
+
+  });
+
+
+
```
pinussilvestrus, line 154:
```suggestion


```

packages/form-js-viewer/test/spec/render/components/form-fields/Checklist.spec.js
```diff
@@ -0,0 +1,232 @@
...
+  afterEach(function() {
+    container.remove();
+  });
+
+
+  it('should render', function() {
+
+    // when
+    const { container } = createChecklist({
+      value: 'camunda-platform'
```
pinussilvestrus, line 32: Shouldn't `value` be an array?

packages/form-js-viewer/test/spec/render/components/form-fields/Checklist.spec.js
```diff
@@ -0,0 +1,232 @@
...
+  beforeEach(function() {
+    container = createFormContainer();
+  });
+
+  afterEach(function() {
+    container.remove();
+  });
+
+
+  it('should render', function() {
```
pinussilvestrus, line 28: Should we have a test case to verify it sets the `checkbox.checked` for the given `value`?

pinussilvestrus (commented):
Looks good to me 👏   Just some minor thoughts / comments

packages/form-js-editor/src/render/components/properties-panel/PropertiesPanel.js
Skaiir: I think we should do this later, reasoning being that there's currently 3 PRs that modify these and it'll be simpler to fix in one fell swoop than re-arrange the file structure at this stage.

packages/form-js-viewer/src/render/components/form-fields/Checklist.js
Skaiir: Yeah exactly, and it is something that we could have a look at again with the whole 'options' discussion we started.

packages/form-js-viewer/src/render/components/form-fields/Checklist.js
Skaiir: Yeah, the rendering state is directly tethered to the underlying data. The reason I did it this way was that I modeled this component from the radio buttons which do it like that.

Whether it's the right thing to do or not, 🤷 I'm not sure. We can also do it how it's done with the checkbox component and react to the checked state. I don't think it matters much.

philippfromme (approved):
Good job! 👍🏻

PR #262, opened 2022-05-23
feat: support Taglist component

### LINKED ISSUE #198: Support scalable (1.000 items +) multi-select component

As a form design I can model a component which can have `n` options, where the later form-user is able to select `m` (`0<=m<=n`) entries. 
The component is able to scope with significant options (e.g., 1.000 items)

Consider:
* Get options either via
  * Form Data (requires https://github.com/bpmn-io/form-js/issues/197)
  * External source / dynamically added via custom JS (see https://github.com/bpmn-io/form-js/issues/199)
* Scale as performance
* Scale as UX 
  * allow to view the 1.000items +
  * allow to filter
  * allow to type-ahead

n/a

__Additional context__

Raised as part of SUPPORT-12025 interaction

Skaiir:
Closes #198

Same as #255 : **Code can be reviewed despite draft**, I'm just keeping this like this so that it doesn't get merged because we're missing an icon.


package-lock.json
```diff
@@ -1016,244 +1016,6 @@
 				"to-fast-properties": "^2.0.0"
 			}
 		},
-		"@bpmn-io/snarkdown": {
```
pinussilvestrus, line 1019: Same as for the other component: https://github.com/bpmn-io/form-js/pull/255#issuecomment-1134578076

I think we can clean up this changes via `npm run reinstall`

packages/form-js-playground/karma.conf.js
```diff
@@ -48,7 +48,11 @@ module.exports = function(karma) {
         modules: [
           'node_modules',
           __dirname
-        ]
+        ],
+        alias: {
+          'react': 'preact/compat',
+          'react-dom': 'preact/compat'
+        }
```
pinussilvestrus, line 55: For what do we need this?

packages/form-js-playground/test/spec/Playground.spec.js
```diff
@@ -47,6 +47,7 @@ describe('playground', function() {
       approved: true,
       approvedBy: 'John Doe',
       product: 'camunda-cloud',
+      tags: ['tag1', 'tag2', 'tag3'],
```
pinussilvestrus, line 50:
```suggestion
      tags: [ 'tag1', 'tag2', 'tag3' ],
```

packages/form-js-viewer/assets/form-js.css
```diff
@@ -235,4 +235,95 @@
 
 .fjs-container .fjs-form-field-text a {
   color: var(--color-blue-205-100-45);
-}
\ No newline at end of file
+}
+
+.fjs-container .fjs-taglist {
```
pinussilvestrus, line 240: That could need a format (indentation looks broken)

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -0,0 +1,158 @@
...
+      }
+      break;
+    }
+  };
+
+  return <div class={ formFieldClasses(type, errors) }>
+    <Label
+      label={ label }
+      id={ prefixId(id, formId) } />
+    <div class={ 'fjs-taglist' + (disabled ? ' disabled' : '') }>
```
pinussilvestrus, line 102: Could we use `classnames` here? I think a good practice would be to structure the classes like
```js
class={ classnames('fjs-taglist', { 'fjs-taglist-disabled': disabled }) }
```

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -0,0 +1,158 @@
...
+  } = field;
+
+  const { formId } = useContext(FormContext);
+  const [filter, setFilter] = useState('');
+  const [selectedValues, setSelectedValues] = useState([]);
+  const [filteredValues, setFilteredValues] = useState([]);
+  const [isDropdownExpanded, setIsDropdownExpanded] = useState(false);
+  const [hasValuesLeft, setHasValuesLeft] = useState(true);
+  const [escapeClose, setEscapeClose] = useState(false);
+  const searchbar = useRef();
```
pinussilvestrus, line 42:
```suggestion
  const searchbarRef = useRef();
```

I think it is good practice, as we are not storing the DOM element, but the complete ref.

packages/form-js-viewer/src/render/components/form-fields/parts/DropdownList.js
```diff
@@ -0,0 +1,100 @@
...
+import useKeyDownAction from '../../../hooks/useKeyDownAction';
+
+export default function DropdownList(props)
+{
+
+  const {
+    keyEventsListener = window,
+    values = [],
+    getLabel = (v) => v,
+    onValueSelected = (v) => {},
```
pinussilvestrus, line 11: Consider having this default as constants 

```js
const DEFAULT_GETTER = (value) => value;
...
```

packages/form-js-viewer/src/render/hooks/useKeyDownAction.js
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from 'preact/hooks';
+
+export default function useKeyDownAction(targetKey, action, listenerElement = window) {
```
pinussilvestrus, line 3: If possible, let's have unit tests for this, as it is heavily used in the component.

Example: https://github.com/bpmn-io/properties-panel/blob/main/test/spec/hooks/useStickyIntersectionObserver.spec.js

packages/form-js-viewer/test/spec/render/components/form-fields/Taglist.spec.js
```diff
@@ -0,0 +1,532 @@
...
+  beforeEach(function() {
+    container = createFormContainer();
+  });
+
+  afterEach(function() {
+    container.remove();
+  });
+
+
+  it('should render', function() {
```
pinussilvestrus, line 28: Instead of having one big render test, we could check for individual parts

```js
it('should render');
it('should render tags');
it('should render tag');
....
```

pinussilvestrus: As you basically did with other parts (description) 👍

packages/form-js-viewer/src/render/components/form-fields/parts/DropdownList.js
```diff
@@ -0,0 +1,100 @@
+import { useState, useEffect, useCallback, useRef, useMemo } from 'preact/hooks';
```
pinussilvestrus, line 1: Same as for the hook: as it should be reusable, should we add unit tests for this one?

pinussilvestrus (changes requested):
Although it's a draft, I have already done a first code review (without going too much into detail, I will have a deeper look later).

I like the extensive test coverage of the Taglist component 👍

package-lock.json
Skaiir: This is actually what happens after a cleanup. I cannot run reinstall on my machine without problems sadly but I deleted the package json + node modules and that's what I get out of it. I think there were some leftover non cleaned up items in the file.

packages/form-js-playground/karma.conf.js
Skaiir: We added SVGs to the form viewer, which also cascades into the forms playground. The SVG parser needs some kind of compatibility aliases to work, as it's been done in the form-editor already. Why this wasn't an issue already because of the form-editor, I'm not so sure, but I haven't been able to build without this.

packages/form-js-viewer/assets/form-js.css
Skaiir: I somehow cannot see this anywhere but on Github. Locally I have only spaces no tabs... Strange.

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -0,0 +1,158 @@
...
+
+  const {
+    description,
+    id,
+    label,
+    values : options
+  } = field;
+
+  const { formId } = useContext(FormContext);
+  const [filter, setFilter] = useState('');
```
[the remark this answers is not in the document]
Skaiir: The spaced array rule has always been inconsistently applied I think. Even across a single project a believe, perhaps something we can look into across our solutions?

Skaiir:
This is now marked as ready for review to get this moving along

To still do:

- [ ] Unit tests for useKeyDownAction
- [ ] Unit tests for DropdownList

barmac:
https://user-images.githubusercontent.com/28307541/171395971-33cf1ee9-23df-41fb-a178-3031c470126a.mov

Is this expected that the dropdown shifts elements below? I'd rather say it should be above other elements just like the HTML5 `<datalist>`:

image

barmac:
For comparison, notice that Outlook search does not shift Outlook UI:
image

Skaiir:
@barmac Not expected this reverted for some reason, I'll investigate

Skaiir:
> > @barmac Not expected this reverted for some reason, I'll investigate
> 
> I think it makes sense to have a test case to verify this.

Honestly I'm not sure. The problem here is position:absolute somehow vanished during merging, but should we really be testing that CSS properties are set to the thing that they are? We don't really test visuals in general.

Skaiir:
@barmac Fair, I have added a test for this :). Let me know if you find some other stuff worth changing here.

packages/form-js-playground/test/spec/form.json
```diff
@@ -74,6 +74,57 @@
         }
       ]
     },
+    {
+      "key": "tags",
+      "label": "Taglist",
+      "type": "taglist",
+      "values": [
```
barmac, line 81: Do we have a test for _(1.000 items +)_ tag list?

Skaiir: We do not, but without dynamic data loading it's also not a covered scenario. Dynamic data might have me rethink some of the ways we handle values, so I would say let's wait on that PR (it's what I'm currently working on) to implement performance tests?

pinussilvestrus: Referring to the name of the closing issue (https://github.com/bpmn-io/form-js/issues/198) adding such a test right now would make sense to verify we really support this.

I would also be okay to wait for the dynamic data PR, but it's not necessary to wait for it. It should work regardless 👍

barmac:
This is a capture of how this behaves with 5000 items on a MacBook Pro 16" (2019):

https://user-images.githubusercontent.com/28307541/171809620-cb07f512-d713-4f25-ac70-40a2212c5ea7.mov

packages/form-js-playground/test/spec/form.json
barmac: > the dynamic data PR

In that case, probably the network will be the bottleneck and not the library.

barmac (approved):
LGTM

Skaiir:
> This is a capture of how this behaves with 5000 items on a MacBook Pro 16" (2019):
> Screen.Recording.2022-06-03.at.09.41.47.mov

Ooof this isn't right. I wouldn't have merged here, because this shows unreliability. Lag would be fine, but duplicate entries due to lag is broken.

barmac:
I haven't noticed the duplicates. Indeed, this needs to be fixed.

PR #270, opened 2022-06-21
feat: dynamic data loading

### LINKED ISSUE #197: Enable loading of values for all 'select-type' components via Form input data

As part of our dynamic data loading initiative for our form components, we would want to enable users to drive their select-able data from sources other than the form schema.

As a starting point, we'll enable users to pass in the data themselves from outside the form via `form input data`, this first ticket handles that functionality. Following up to this will be allowing the configuration of the type of data source via the properties panel, as it is currently blocked by https://github.com/bpmn-io/form-js/issues/249.

We want to support the rendering of 'select-type' component values via Input data, so that a form designer may dynamically build up these components. 

To this end, we'll need a few things:

- Create a differentiation in the schema for different data sources
- Allow our form components to interpret that differentiation and render appropriately
- Figure out how we want these dynamically driven form components to look in the preview when no data source is provided

___

SCHEMA EXAMPLE

**schema**
```js
{
  // values is ommited on purpose!
  //  "values": [
  //  {
  //   "label": "Value",
  //   "value": "value"
  //   }
  // ],
  "label": "Radio",
  "type": "radio",
  "id": "myRadio",
  "key": "radio",
  "valuesKey": "radioInputs"
}
```

**form data input**
```json
{
  "radio": "valA",
  "radioInputs": [
    {
      "label": "valA",
      "value": "valA"
    },
    {
      "label": "valB",
      "value": "valB"
    },
  ]
}
```

**Result**
Screenshot_20211108_203747

___

Implementation steps:
- [x] Agree on the final schema
- [x] Agree on UX for "empty" data, rendering nothing might be confusing to users
- [ ] Set up a whole new dynamic data form test case to work from
- [x] Enable components to be rendered dynamically
  - [x] Select
  - [x] Radio
  - [x] Checklist (depends on https://github.com/bpmn-io/form-js/pull/255)
  - [x] Taglist (depends on https://github.com/bpmn-io/form-js/issues/198)
- [x] Implement "empty" data solution

n/a

__Additional context__

Requires https://github.com/bpmn-io/form-js/issues/196
Raised during discussion of https://jira.camunda.com/browse/SUPPORT-12025
Also requested via https://jira.camunda.com/browse/SUPPORT-12306
Child of https://github.com/bpmn-io/form-js/issues/200

### LINKED ISSUE #256: Allowing configuration of values (options) source via the properties panel for 'select-type' components

As part of our dynamic data loading initiative for our form components, we would want to enable users to drive their select-able data from sources other than the form schema.

To this end, we need to enable configuration within the form schema to set this external source somehow. This will be used to allow values to be passed in from `input data`, but may be extended in the future to apply to `functions`, and others.

Sketch by @pinussilvestrus :

image

This toggle should drive the form schema and define the data source on it. How exactly this will be done still needs to be defined as part of https://github.com/bpmn-io/form-js/issues/197. 

__Additional context__

Blocking tickets: 
Depends on https://github.com/bpmn-io/form-js/issues/197
Depends on https://github.com/bpmn-io/form-js/issues/249

Parent: 
Child of https://github.com/bpmn-io/form-js/issues/200

Skaiir:
Closes #197 
Closes #256 

Draft for now to gather feedback. Please have a look around, this should work for all select type controls (taglist, checklist, select and radio).


packages/form-js-editor/src/render/components/properties-panel/entries/OptionsSourceEntry.js
```diff
@@ -0,0 +1,245 @@
...
+      value: option
+    }));
+  };
+
+  return SelectEntry({
+    element: field,
+    getOptions: getOptionsSourceOptions,
+    getValue,
+    id,
+    label: 'Options Source',
```
nikku, line 90:
```suggestion
    label: 'Options source',
```

packages/form-js-editor/src/render/components/properties-panel/entries/DefaultValueEntry.js
```diff
@@ -11,10 +11,10 @@ import { get } from 'min-dash';
 
 import { useService } from '../../../hooks';
 
-import { INPUTS } from '../Util';
+import { INPUTS, OPTIONS_INPUTS } from '../Util';
 
 
-export default function DefaultValueEntry(props) {
+export default function DefaultOptionEntry(props) {
```
nikku, line 17: Did we make the conscious decision to rename `values` to options? What is the benefit?

nikku (commented):
Look like this is going into a good direction!

A few comments / questions / remarks:

* Language remark: We should always use `First capital then lowercase` for labels, descriptions and titles.

* Seems like we move away from _Values_ to _Options_. Was that a conscious decision? If so, what is the rationale? I'd personally not want to mix both concerns (switching to options + adding support for dynamic loading in one PR / one commit).

* Given the current UI architecture we're not well equipped for lists inside existing groups. My proposal is therefor to move value configuration (either list of static values, or field to access the value) to a separate group. The group can have a meaningful name to clearly indicate what it does: 
    capture VyORz5_optimized

Skaiir:
@nikku 

I guess what mostly pushed me to going with options here is the fact that we are committing to 'valuesKey' on the form schema and that's why I wanted to take action now rather than later. But in any case sure I guess it's still better to be at least consistent in our schema, valuesKey, values, ect... Any changes may be considered down the line.
____

> Given the current UI architecture we're not well equipped for lists inside existing groups

Could you expand on this, I never saw a problem myself implementing this so maybe I'm missing some context.
____

Regarding the form data submit values, I need to confirm this still but I'm pretty sure that window is currently not an accurate representation of what gets sent out. For example when you disable a field it doesn't remove it from the output. I am pretty sure the actual form submission code iterates over all input fields and checks if they are indeed disabled. - This may be more something @pinussilvestrus might want to take on later since he is doing the playground integration.

packages/form-js-editor/src/render/components/properties-panel/entries/ValuesSourceEntries.js
```diff
@@ -0,0 +1,219 @@
...
+
+  return [{
+    id: 'input-values-key',
+    component: InputValuesKey,
+    label: 'Input values key',
+    description: 'Define which input property to populate the values from',
+    isEdited: isSelectEntryEdited,
+    editField,
+    field,
+  }];
```
pinussilvestrus, line 38:
```suggestion
  return [
    {
      id: 'input-values-key',
      component: InputValuesKey,
      label: 'Input values key',
      description: 'Define which input property to populate the values from',
      isEdited: isSelectEntryEdited,
      editField,
      field
    }
  ];
```

packages/form-js-editor/src/render/components/properties-panel/entries/ValuesSourceEntries.js
```diff
@@ -0,0 +1,219 @@
...
+  const {
+    editField,
+    field
+  } = props;
+
+  return [{
+    id: 'input-values-key',
+    component: InputValuesKey,
+    label: 'Input values key',
+    description: 'Define which input property to populate the values from',
```
pinussilvestrus, line 34: I'm fine with this, but I'm also not an expert in technical wording. Maybe @christian-konrad can give some feedback on this.

image

packages/form-js-editor/src/render/components/properties-panel/entries/ValuesSourceEntries.js
```diff
@@ -0,0 +1,219 @@
...
+  [VALUES_SOURCES.INPUT]: ['valuesKey'],
+};
+
+const VALUES_SOURCES_DEFAULTS = {
+  [VALUES_SOURCES.STATIC]: [],
+  [VALUES_SOURCES.INPUT]: '',
+};
+
+
+//* Helpers
```
pinussilvestrus, line 208: To be consistent

```suggestion
// helpers ///////////////////
```

packages/form-js-editor/src/render/components/properties-panel/entries/ValuesSourceEntries.js
```diff
@@ -0,0 +1,219 @@
...
+    element: field,
+    getValue,
+    id,
+    label,
+    setValue
+  });
+}
+
+
+//* CONFIG
```
pinussilvestrus, line 183: Maybe put this one on top of the file, as these are constants.

```suggestion
// config ///////////////
```

packages/form-js-editor/src/render/components/properties-panel/entries/ValuesSourceEntries.js
```diff
@@ -0,0 +1,219 @@
...
+      id,
+      label: entry.label,
+      entries: ValueEntry({
+        editField,
+        field,
+        idPrefix: id,
+        index,
+        validateFactory
+      }),
+      autoFocusEntry: true,
```
pinussilvestrus, line 101:
```suggestion
      autoFocusEntry: id + '-label'
```

Otherwise, it will always focus on the _first input of the page_.

packages/form-js-editor/src/render/components/properties-panel/groups/ValuesGroups.js
```diff
@@ -0,0 +1,52 @@
...
+      component: Group,
+      entries: InputKeyValuesSourceEntry(context)
+    });
+  }
+  else if (valuesSource === VALUES_SOURCES.STATIC) {
+    groups.push({
+      id,
+      label: 'Static values',
+      component: ListGroup,
+      entries: undefined,
```
pinussilvestrus, line 46: Why do we need this to undefined?

packages/form-js-editor/src/render/components/properties-panel/entries/ValuesSourceEntries.js
```diff
@@ -0,0 +1,219 @@
...
+    id: idPrefix
+  } = props;
+
+  const {
+    values
+  } = field;
+
+  const addEntry = (e) => {
+
+    e.preventDefault();
```
pinussilvestrus, line 54:
```suggestion
    e.stopPropagation();
```

Otherwise, it leads to this behavior

Kapture 2022-07-07 at 13 14 31

packages/form-js-editor/src/render/components/properties-panel/groups/ValuesGroups.js
```diff
@@ -0,0 +1,52 @@
+import { ValuesSourceSelectorEntry, StaticValuesSourceEntry, InputKeyValuesSourceEntry, getValuesSource, VALUES_SOURCES } from '../entries';
+
+import { Group, ListGroup } from '@bpmn-io/properties-panel';
+
+import {
+  VALUES_INPUTS
+} from '../Util';
+
+export default function ValuesGroups(field, editField) {
```
pinussilvestrus, line 9: Can you explain why we have a single file for all the groups? It's now harder to read as the logic for multiple groups is handled in a single file, not following the same structure as beforehand.

Shouldn't the previous `ValuesGroup` still work (maybe with some different labels and some configuration)?

packages/form-js-editor/src/render/components/properties-panel/entries/index.js
```diff
@@ -8,4 +8,5 @@ export { default as KeyEntry } from './KeyEntry';
 export { default as LabelEntry } from './LabelEntry';
 export { default as TextEntry } from './TextEntry';
 export { default as ValueEntry } from './ValueEntry';
-export { default as CustomValueEntry } from './CustomValueEntry';
\ No newline at end of file
+export { default as CustomValueEntry } from './CustomValueEntry';
+export * from './ValuesSourceEntries';
```
pinussilvestrus, line 12: Is this expected? It now also exports some helpers as `getValuesSource`, which I wouldn't expect to import from `entries`. What I'd rather expect

```js
import { getValuesSource, VALUES_SOURCES } from './entries/ValuesSourceEntries';
```

packages/form-js-editor/src/render/components/properties-panel/entries/ValuesSourceEntries.js
```diff
@@ -0,0 +1,219 @@
+import { get, isUndefined, without } from 'min-dash';
+
+import { SelectEntry, isSelectEntryEdited, TextFieldEntry } from '@bpmn-io/properties-panel';
+import ValueEntry from './ValueEntry';
+
+import { arrayAdd } from '../Util';
+import { useService } from '../../../hooks';
+
+export function ValuesSourceSelectorEntry(props) {
```
pinussilvestrus, line 9: Same as for the `groups`. This kind of breaks the current structures of one entry per file. Is there a specific reason why need to keep these entries together?

packages/form-js-editor/src/render/components/properties-panel/entries/DefaultValueEntry.js
```diff
@@ -56,11 +57,13 @@ export default function DefaultValueEntry(props) {
   if (type === 'radio' || type === 'select') {
     entries.push({
       ...defaultOptions,
-      component: DefaultValueMulti,
+      component: DefaultValueSingleSelect,
       isEdited: isSelectEntryEdited
     });
   }
 
+  // Todo: implement a multiselect equivalent (https://github.com/bpmn-io/form-js/issues/265)
```
pinussilvestrus, line 65: As a good practice, add your name to the note so the next person can get back to you for more context 👍 
```suggestion
  // todo(Skaiir): implement a multiselect equivalent (cf. https://github.com/bpmn-io/form-js/issues/265)
```

packages/form-js-editor/test/spec/properties-panel/PropertiesPanel.spec.js
```diff
@@ -530,7 +535,7 @@ describe('properties panel', function() {
               });
 
               // when
-              const input = screen.getByLabelText('Value', { selector: '#bio-properties-panel-Radio_1-value-0-value' });
+              const input = screen.getByLabelText('Value', { selector: '#bio-properties-panel-Radio_1-values-staticValue-0-value' });
```
pinussilvestrus, line 538: Is the ID change really needed or just a consequence of the re-arrangement of the entries?

It's not a breaking change as technical IDs are anyway just internal, but I'd like to prevent these changes (if not needed).

packages/form-js-editor/test/spec/properties-panel/PropertiesPanel.spec.js
```diff
@@ -575,6 +580,108 @@ describe('properties panel', function() {
 
       });
 
+
+      describe('dynamic values', function() {
```
pinussilvestrus, line 584: Given the current test structure, I think it's fine to have these tests in the `PropertiesPanel.spec.js`. In a perfect world, we would maybe have them in separate files, similar to how we do it in BPMN land.

packages/form-js-editor/test/spec/properties-panel/PropertiesPanel.spec.js
```diff
@@ -575,6 +580,108 @@ describe('properties panel', function() {
...
+            field
+          });
+
+          // assume
+          const input = screen.getByLabelText('Values source');
+
+          expect(input.value).to.equal('static');
+
+          // when
+          fireEvent.input(input, { target: { value: 'input' } });
```
pinussilvestrus, line 605: We could use `VALUES_SOURCES.INPUT` here, as I also wondered where the `input` value is coming from.

packages/form-js-editor/test/spec/properties-panel/PropertiesPanel.spec.js
```diff
@@ -575,6 +580,108 @@ describe('properties panel', function() {
 
       });
 
+
+      describe('dynamic values', function() {
+
+        it('source should be configurable', function() {
```
pinussilvestrus, line 586: Maybe rephrase, as it (also) tests a behavior.
```suggestion
        it('should be configurable - sets valuesKey', function() {

```

pinussilvestrus: The same would go for the other test case below.

packages/form-js-editor/test/spec/properties-panel/groups/GeneralGroup.spec.js
```diff
@@ -367,21 +367,115 @@ describe('GeneralGroup', function() {
     });
 
 
-    it('should render for INPUTS', function() {
+    describe('for singleSelect-like INPUTS', () => {
 
-      // given
-      for (const type of INPUTS.filter(i => ![ 'checklist', 'taglist' ].includes(i))) {
+      const singleSelectInputTypes = ['radio', 'select'];
```
pinussilvestrus, line 372:
```suggestion
      const singleSelectInputTypes = [ 'radio', 'select' ];
```

Mh, don't we lint that?

packages/form-js-editor/test/spec/properties-panel/groups/GeneralGroup.spec.js
```diff
@@ -367,21 +367,115 @@ describe('GeneralGroup', function() {
...
 
-        const field = { type };
+      it('should NOT render by default', () => {
```
pinussilvestrus, line 374: What means _default_ here? When `valuesKey` is set? Maybe we should name the test this way then.

packages/form-js-editor/test/spec/properties-panel/groups/GeneralGroup.spec.js
```diff
@@ -367,21 +367,115 @@ describe('GeneralGroup', function() {
...
+        }
+
+      });
+
+    });
+
+
+    describe('for all other INPUTS', () => {
+
+      const otherInputTypes = INPUTS.filter(i => !['checklist', 'taglist', 'radio', 'select'].includes(i));
```
pinussilvestrus, line 460: We can use 
```suggestion
      const otherInputTypes = INPUTS.filter(i => !VALUES_INPUTS.includes(i));
```

packages/form-js-playground/test/spec/Playground.spec.js
```diff
@@ -48,6 +48,20 @@ describe('playground', function() {
       approvedBy: 'John Doe',
       mailto: ['regional-manager', 'approver'],
       product: 'camunda-cloud',
+      dynamicDataTestData: [
```
pinussilvestrus, line 51: I like that the playground test data mostly follows a real-world use case here. Maybe we can come up with something less generic than `dynamicTestData`? Something about currencies (Euro, Dollar, Pound, ...)  

I'm also not super creative though 😅

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -41,26 +41,42 @@ export default function Taglist(props) {
...
+    }
+    else {
+      setSelectedValues([]);
+    }
+  }, [ JSON.stringify(values), options, loadState ]);
 
   useEffect(() => {
-    setFilteredValues(options.filter((o) => o.label && (o.label.toLowerCase().includes(filter.toLowerCase())) && !values.includes(o.value)));
+    if (loadState === LOAD_STATES.LOADED)
+    {
```
pinussilvestrus, line 66:
```suggestion
    if (loadState === LOAD_STATES.LOADED) {
```

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
```diff
@@ -0,0 +1,61 @@
...
+  } = field;
+
+  const [ valuesGetter, setValuesGetter ] = useState({ values: [], state: LOAD_STATES.LOADING });
+  const initialData = useService('form')._getState().initialData;
+
+  useEffect(() => {
+
+    let values = [];
+
+    if (valuesKey !== undefined) {
```
pinussilvestrus, line 38: Why not simply?

```suggestion
    if (valuesKey) {
```

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
```diff
@@ -0,0 +1,61 @@
...
+
+      const safeInitialData = initialData || {};
+      const keyedValues = safeInitialData[ valuesKey ];
```
pinussilvestrus, line 41: Just personal taste

```suggestion
      const keyedValues = (initialData || {})[ valuesKey ];
```

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
```diff
@@ -0,0 +1,61 @@
...
+
+      if (keyedValues && Array.isArray(keyedValues)) {
+        values = keyedValues;
+      }
+    }
+    else if (staticValues !== undefined) {
+      values = Array.isArray(staticValues) ? staticValues : [];
+    }
+    else {
+      setValuesGetter({ values, state: LOAD_STATES.ERROR });
```
pinussilvestrus, line 51: Is this expected in any case? If not, maybe we should even throw a proper error then.

If it's expected, maybe the `error` state alone is not enough context for the API consumer to handle this properly. At least an error message would be needed.

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
```diff
@@ -0,0 +1,61 @@
...
+ * @property {(LOAD_STATES)} state - The values data's loading state, to use for conditional rendering
+ */
+
+/**
+ * A hook to load values for single and multiselect components.
+ *
+ * @param {Object} field - The form field to handle values for
+ * @return {ValuesGetter} valuesGetter - A values getter object providing loading state and values
+ */
+export default function(field) {
```
pinussilvestrus, line 25: I like this approach. Even though it might not be needed _right now_, this will help us when adding something like lazy data loading in the future. It also defines the data loading in a central place for all form fields 👍 Something we could have for conditional rendering as well.

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
pinussilvestrus: And if it's expected, should we have a test case for this? E.g. testing what's happening when a form field neither defines static values nor dynamic input values.

packages/form-js-editor/test/spec/properties-panel/groups/GeneralGroup.spec.js
pinussilvestrus: Damn, we still use `0.12` of our eslint rules, so the [bracket rules](https://github.com/bpmn-io/eslint-plugin-bpmn-io/commit/9dba9c9b29d28ed47ea7a471bd252e59231066d3) don't apply yet.

pinussilvestrus (changes requested):
Functional-wise, it works like a charm, so nice job 👏 

I have some general understanding questions and some remarks regarding the general structure. Generally, I'd like to keep things structure-wise (groups, entries) whenever possible, or at least understand why structure changes are needed 👍

packages/form-js-editor/src/render/components/properties-panel/PropertiesPanel.js
```diff
@@ -7,7 +7,7 @@ import {
   CustomValuesGroup,
   GeneralGroup,
   ValidationGroup,
-  ValuesGroup
+  ValuesGroups
```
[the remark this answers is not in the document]
nikku: It was a conscious decision, driven by me: https://github.com/bpmn-io/form-js/pull/270#pullrequestreview-1016796831.

packages/form-js-editor/src/render/components/properties-panel/entries/DefaultValueEntry.js
```diff
@@ -147,7 +150,7 @@ function DefaultValueNumber(props) {
   });
 }
 
-function DefaultValueMulti(props) {
+function DefaultValueSingleSelect(props) {
```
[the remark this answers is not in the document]
Skaiir: @pinussilvestrus Because this component now only makes sense with singleselect type values controls (radio, select). We have plans to also make something available for multiselects (checklist/taglist), but since it requires some properties panel changes and a new control it'll come later.

Lmk if you find something wrong with the reasoning :)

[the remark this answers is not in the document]
pinussilvestrus: Makes sense to me 👍

packages/form-js-editor/src/render/components/properties-panel/PropertiesPanel.js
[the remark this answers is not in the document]
pinussilvestrus: Got it! Thanks for the explanation, I agree 👍

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
Skaiir: I guess this can be argued, but the presence of the key itself is how we determine what type of source we use. I'm re-enforcing it here, but this is mostly on the front end I suppose so I don't know exactly how justified I am. 

Adding this would make us a tiny bit more redundant (we could have valuesKey=null, values=[someSet] still work), but at the same time we're agreeing that we don't accept two types of sources on the form spec, this is how the rest of the codebase works so I'm not sure.

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
Skaiir: It's not expected in any case, except if someone messes with a form definition manually.

If they accidentally remove both, then push it to our ecosystem and it gets picked up by Tasklist. 

It might also hit if we load in that broken diagram into form-js. This scenario is worth testing. I know the properties panel is somewhat fault tolerant to it, but here we wouldn't want to be since we're in an unallowed design state. At the same time we also don't want to throw an actual exception since we'd still want people to be able to edit their way out of this problem via the properties panel. 
This is a linting concern also in a way. But we need some state to allow this to not crash the app but also make the user aware that something is wrong. I think we'll need to do more stuff with those error states in the future (for example if we're loading from an API, and it's down). So I would say, we keep this, but test it.

packages/form-js-editor/src/render/components/properties-panel/entries/ValuesSourceEntries.js
```diff
@@ -0,0 +1,219 @@
...
+      value: valueSource
+    }));
+  };
+
+  return SelectEntry({
+    element: field,
+    getOptions: getValuesSourceOptions,
+    getValue,
+    id,
+    label: 'Values source',
```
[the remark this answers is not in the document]
Skaiir: I'll just go with hiding the label since value source is explicit enough

packages/form-js-editor/src/render/components/properties-panel/groups/ValuesGroups.js
Skaiir: I honestly do not know why but the linting won't let me do without explicitly setting that. It's happened in a few place. Maybe you know about it?

pinussilvestrus:
One thing I also noticed: Should we [update the `schemaVersion`](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-viewer/src/index.js), as we adding `valuesKey` as addition? And maybe adding it to [`FORM_SCHEMA.md`](https://github.com/bpmn-io/form-js/blob/develop/docs/FORM_SCHEMA.md)?

packages/form-js-editor/src/render/components/properties-panel/groups/ValuesGroups.js
Skaiir: I've done this simply because the groups are very tightly bound. Looking at this, you can instantly tell that there's a logical link between the three groups in the code. You see the conditional group choosing code.

If you we separated the code out into multiple groups we'd have:

ValuesSourceGroup => always visible
StaticValuesGroup => visible if valuesSource == static
DynamicValuesGroup => visible if valuesSource == dynamic

I think that at a glance, grouping those in a single file tells the story of what these are about better.

packages/form-js-editor/src/render/components/properties-panel/entries/ValuesSourceEntries.js
Skaiir: Here I absolutely agree, these components should be separate and I'm currently doing just that

packages/form-js-editor/src/render/components/properties-panel/entries/index.js
Skaiir: Going to split all the assisting code into its own util, so that it may be used by entries and groups without issue

packages/form-js-editor/test/spec/properties-panel/PropertiesPanel.spec.js
Skaiir: I do want to keep these somewhat logical so I'll still rename but to something that makes more sense :)

packages/form-js-editor/src/render/components/properties-panel/groups/ValuesGroups.js
Skaiir: It seems that the @type here is enough, in any case it's a bit clunky but better than using undefined for sure.

docs/FORM_SCHEMA.md
```diff
@@ -52,7 +58,7 @@ A form is defined as JSON.
       "key": "reset",
       "label": "Reset",
       "type": "button"
-    }
+    },
```
pinussilvestrus, line 61:
```suggestion
    }
```

Skaiir:
@pinussilvestrus Had to rewrite history for this but it came out okay, everything is added.

pinussilvestrus (approved):
Great job 👏

PR #284, opened 2022-07-18
chore(playground): move components to separate files

pinussilvestrus:
Just as a cleanup activity. 

* Change structure of Playground to have components in separate files
* Prepares the playground to changes that will come (cf. https://github.com/bpmn-io/internal-docs/issues/527).


philippfromme (approved):
Makes sense. 👍🏻

PR #290, opened 2022-07-25
Palette module

### LINKED ISSUE #280: Provide Palette as own feature module
__What should we do?__

As part of delivering the next version of the `form-js-playground`, the Palette should be renderable in different contexts. For now, it's [part of the `form-js-editor`](https://github.com/bpmn-io/form-js/blob/334b6a1d7f34403da9385f1d57d78eaca80bd6e2/packages/form-js-editor/src/render/components/FormEditor.js#L293).

The goal should be to make this components an actual feature module to be configured and used independently from the editor implementation.

__Why should we do it?__

The Palette should not be hard-coded inside the form-js-editor, but should be more flexible. 

pinussilvestrus:
Closes #280 

* Provides Palette as own module
* Render Palette in Editor per default
* Add Palette individually in Playground

Run playground
```sh
npm run start:playground
```

Run editor
```sh
npm run start:editor
```

image


pinussilvestrus:
Note: Dedicated test coverage is currently missing for the Playground. I plan to tackle this as part of the properties panel integration (cf. #291).

marstamm (approved):
Great improvement!

Issue #288, opened 2022-07-21
Move component stress tests out of the main form definition & add flag to start a stress tested instance


Skaiir:
__What should we do?__

Currently we have a test file with 10000 value entries for the Taglist. This should ideally not interrupt regular use cases, and we should stress test more components than just the Taglist. Hence I suggest we should move this into their own form definition and stress test all components that should work with larger datasets.

See: https://github.com/bpmn-io/form-js/blob/739e52f9093a08f68244ec9060f92123e0879397/packages/form-js-viewer/test/spec/form.json#L370

Also now that dynamic data works, we should use dynamic data instead of massive form definitions to make this happen, avoids us having 10000 lines of values in our form definitions.

__Why should we do it?__

This would let us run stress tests on our components easily as a developer by simply running `npm run start:viewer -- --stressTest`, but keeping it optional. It would also make the codebase more succinct.

PR #293, opened 2022-07-28
feat: stop taglist duplicate entries during slowdowns

### LINKED ISSUE #268: Taglist behaving inconsistently during performance issues.
__Describe the Bug__

Follow up to https://github.com/bpmn-io/form-js/issues/198

When lagging due to a very large amount of values, the tag-list control [sometimes creates duplicate entries](https://github.com/bpmn-io/form-js/pull/262#issuecomment-1145681061). 

__Steps to Reproduce__

1. Use a tag-list with a very large size (~5000)
2. Select new entries

__Expected Behavior__

We shouldn't be getting any duplicate entries.

Skaiir:
Closes #268

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Taglist.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Taglist.js
@@ -1,4 +1,4 @@
-import { useContext, useEffect, useRef, useState } from 'preact/hooks';
+import { useContext, useEffect, useMemo, useRef, useState } from 'preact/hooks';
 import useValuesAsync, { LOAD_STATES } from '../../hooks/useValuesAsync';
 
 import { FormContext } from '../../context';
@@ -38,10 +38,9 @@ export default function Taglist(props) {
 
   const { formId } = useContext(FormContext);
   const [ filter, setFilter ] = useState('');
-  const [ selectedValues, setSelectedValues ] = useState([]);
-  const [ filteredValues, setFilteredValues ] = useState([]);
+  const [ filteredOptions, setFilteredOptions ] = useState([]);
   const [ isDropdownExpanded, setIsDropdownExpanded ] = useState(false);
-  const [ hasValuesLeft, setHasValuesLeft ] = useState(true);
+  const [ hasOptionsLeft, setHasOptionsLeft ] = useState(true);
   const [ isEscapeClosed, setIsEscapeClose ] = useState(false);
   const searchbarRef = useRef();
 
@@ -50,43 +49,43 @@ export default function Taglist(props) {
     values: options
   } = useValuesAsync(field);
 
+  // We cache a map of option values to their index so that we don't need to search the whole options array every time to correlate the label
+  const valueToOptionMap = useMemo(() => Object.assign({}, ...options.map((o, x) => ({ [o.value]:  options[x] }))), [ options ]);
 
   // Usage of stringify is necessary here because we want this effect to only trigger when there is a value change to the array
   useEffect(() => {
     if (loadState === LOAD_STATES.LOADED) {
-      const selectedValues = values.map(v => options.find(o => o.value === v)).filter(v => v !== undefined);
-      setSelectedValues(selectedValues);
+      setFilteredOptions(options.filter((o) => o.label && o.value && (o.label.toLowerCase().includes(filter.toLowerCase())) && !values.includes(o.value)));
     }
     else {
-      setSelectedValues([]);
+      setFilteredOptions([]);
     }
-  }, [ JSON.stringify(values), options, loadState ]);
+  }, [ filter, JSON.stringify(values), options, loadState ]);
 
   useEffect(() => {
-    if (loadState === LOAD_STATES.LOADED) {
-      setFilteredValues(options.filter((o) => o.label && o.value && (o.label.toLowerCase().includes(filter.toLowerCase())) && !values.includes(o.value)));
-    }
-    else {
-      setFilteredValues([]);
-    }
-  }, [ filter, JSON.stringify(values), options ]);
-
-  useEffect(() => {
-    setHasValuesLeft(selectedValues.length < options.length);
-  }, [ selectedValues.length, options.length ]);
+    setHasOptionsLeft(options.length > values.length);
+  }, [ options.length, values.length ]);
 
   const onFilterChange = ({ target }) => {
     setIsEscapeClose(false);
     setFilter(target.value);
   };
 
-  const selectValue = (option) => {
-    setFilter('');
-    props.onChange({ value: [ ...values, option.value ], field });
+  const selectValue = (value) => {
+    if (filter) {
+      setFilter('');
+    }
+
+    // Ensure values cannot be double selected due to latency
+    if (values.at(-1) === value) {
+      return;
+    }
+
+    props.onChange({ value: [ ...values, value ], field });
   };
 
-  const deselectValue = (option) => {
-    props.onChange({ value: values.filter((v) => v != option.value), field });
+  const deselectValue = (value) => {
+    props.onChange({ value: values.filter((v) => v != value), field });
   };
 
   const onInputKeyDown = (e) => {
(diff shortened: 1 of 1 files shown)


Skaiir:
Note that the actual fix is at line 80, but I've generally removed some unnecessary complexity from the control as well which was enabled by the sanitation code that we recently merged through.

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -50,43 +49,43 @@ export default function Taglist(props) {
...
-  const selectValue = (option) => {
-    setFilter('');
-    props.onChange({ value: [ ...values, option.value ], field });
+  const selectValue = (value) => {
+    if (filter) {
+      setFilter('');
+    }
+
+    // Ensure values cannot be double selected due to latency
+    if (values.at(-1) === value) {
```
pinussilvestrus, line 80: I think we don't guarantee support for IE anymore, so this might be fine.

https://caniuse.com/?search=Array%3A%20at

Just to make sure, is there maybe a safer choice instead of using `.at` ?

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -120,13 +119,13 @@ export default function Taglist(props) {
       id={ prefixId(id, formId) } />
     <div class={ classNames('fjs-taglist', { 'disabled': disabled }) }>
       {!disabled && loadState === LOAD_STATES.LOADED &&
-        selectedValues.map((sv) => {
+        values.map((v) => {
           return (
             <div class="fjs-taglist-tag" onMouseDown={ (e) => e.preventDefault() }>
               <span class="fjs-taglist-tag-label">
-                {sv.label}
+                {valueToOptionMap[v] ? valueToOptionMap[v].label : `unexpected value{${v}}`}
```
pinussilvestrus, line 126: Do we have a test case for this? If not, let's add it.

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -50,43 +49,43 @@ export default function Taglist(props) {
...
 
+  // We cache a map of option values to their index so that we don't need to search the whole options array every time to correlate the label
+  const valueToOptionMap = useMemo(() => Object.assign({}, ...options.map((o, x) => ({ [o.value]:  options[x] }))), [ options ]);
 
   // Usage of stringify is necessary here because we want this effect to only trigger when there is a value change to the array
   useEffect(() => {
     if (loadState === LOAD_STATES.LOADED) {
-      const selectedValues = values.map(v => options.find(o => o.value === v)).filter(v => v !== undefined);
-      setSelectedValues(selectedValues);
+      setFilteredOptions(options.filter((o) => o.label && (o.label.toLowerCase().includes(filter.toLowerCase())) && !values.includes(o.value)));
```
pinussilvestrus, line 58: This looks extremely complicated; any chance to split it up (or have a helper function)?

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -50,43 +49,43 @@ export default function Taglist(props) {
     values: options
   } = useValuesAsync(field);
 
+  // We cache a map of option values to their index so that we don't need to search the whole options array every time to correlate the label
+  const valueToOptionMap = useMemo(() => Object.assign({}, ...options.map((o, x) => ({ [o.value]:  options[x] }))), [ options ]);
```
pinussilvestrus, line 53: Maybe my eyes trick me because the expression is pretty long, but are we trying to assign an array (`.map` result) to an empty object?

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
pinussilvestrus, line 80: I can't find any test case for this?

pinussilvestrus (commented):
Looks good performance-wise 👍 Let's consider some test coverage for this.

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
Skaiir: Yeah so this basically just turns ({a: x}, {b: y}) into {a: x, b: y}. It's just a short way to create an object from an array.

Skaiir: There might be a better way to do this but I'm not certain.

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
Skaiir: Well this is a bit of a strange one, but basically it's a state you shouldn't be able to get into unless there is a code issue. I will change this actually, we can instead make the sanitation bullet proof and check if the label is present there. I think that makes more sense.

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
Skaiir: Yeah I've switched over to the good ol' a[a.length-1]

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
Skaiir: I have moved it out and made it easier to read👍

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
Skaiir: Actually I've removed this entirely to match behavior with all our other controls. If no label is defined we just don't display a label. Perhaps we may want to do something else in the future, but let's be consistent before we set a behavior in stone. I've raised it as a follow up: https://github.com/bpmn-io/form-js/issues/296

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
Skaiir: Could we realistically rely on test cases for this? This is after all an issue that only occurs due to large re-renders. Could we replicate the lagging conditions under test? The bug was not consistent in the first place so I see a test like this could very likely stay green even if the issue is back.

Skaiir: I'm planning on creating a custom run script to stress test our components that are supposed to handle large datasets, tracked here: https://github.com/bpmn-io/form-js/issues/288. Perhaps this could be something better to handle these kinds of cases. It's not automatic testing but it would help.

pinussilvestrus: If we can't reliably reproduce it by setting such a value in the existing TagList specs, we could consider to extract the `selectValue` functionality, export it and have a dedicated unit test in the TagList specs.

```js
// internal implementation of the component
const _selectValue = (value) => {
 return selectValue(value, values, setFilter, props.onChange);
}

// helper, which we test
export function selectValue(value, values, setFilter, onChange) {
  // ...
}
```

We should consider it only if it makes sense and it's worth the effort. My gut feeling just notified me because we fixed something in the PR and didn't provide a test case for it. 😅

pinussilvestrus:
@Skaiir what is missing to get this one merged?

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
Skaiir: @pinussilvestrus Where would you have the helper function? I actually think this would be the most robust thing to do, albeit a bit intrusive. I guess the question is, do we want to complicate things to this extent to test the rather simple condition? In a way, it does also disrupt the code. 

I personally think it's too disruptive.

Skaiir: @pinussilvestrus Where would you have the helper function? I actually think this would be the most robust thing to do, albeit a bit intrusive. I guess the question is, do we want to complicate things to this extent to test the rather simple condition? In a way, it does also disrupt the code. 

I personally think it's too disruptive.

pinussilvestrus: Agree, It's ok to keep things simple (as they are) in this case 👍

PR #297, opened 2022-07-30
test(viewer): add basic a11y unit tests

pinussilvestrus:
Adds the same unit test styles as we already have in `properties-panel` and `bpmn-js-properties-panel`. I think this helps us a lot to verify that our form fields validate against basic standards.


marstamm (approved):
And it succeeds with no changes to the components themselves. Great Stuff!

PR #298, opened 2022-08-04
Flexible playground components

### LINKED ISSUE #292: Playground components are freely attachable and detachable
__What should we do?__

In order to make the playground components, they need to be freely attachable and detachable (as other components, e.g. the palette). 

This was already spiked via https://github.com/bpmn-io/form-js/pull/253.

__Why should we do it?__

pinussilvestrus:
Closes #292 

---

This makes it possible to attach all playground components individually. This enables a mode for the playground to render nothing (by not providing `container` config) but only orchestrate the different instances. This is a requirement to create a product-specific playground version (cf. https://github.com/bpmn-io/internal-docs/issues/527).

This also makes the "actions" (download, embed) optional. 

**Overview components**

image

**Demo**

Check out [this demo](https://form-playground-demo.netlify.app/demo/) to get a feeling of what you can do with the API changes.

### DIFF
--- a/packages/form-js-playground/src/Playground.js
+++ b/packages/form-js-playground/src/Playground.js
@@ -8,10 +8,11 @@ import { PlaygroundRoot } from './components/PlaygroundRoot';
 
 /**
  * @typedef { {
- *  container: Element,
+ *  container?: Element,
  *  schema: any,
  *  data: any,
  *  editor?: { inlinePropertiesPanel: Boolean }
+ *  actions?: { display: Boolean }
  * } } FormPlaygroundOptions
  */
 
@@ -36,7 +37,9 @@ export default function Playground(options) {
 
   container.classList.add('fjs-pgl-parent');
 
-  parent.appendChild(container);
+  if (parent) {
+    parent.appendChild(container);
+  }
 
   const handleDrop = fileDrop('Drop a form file', function(files) {
     const file = files[0];
@@ -50,6 +53,16 @@ export default function Playground(options) {
       }
     }
   });
+
+  const withRef = function(fn) {
+    return function(...args) {
+      if (!ref) {
+        throw new Error('Playground is not initialized.');
+      }
+
+      fn(...args);
+    };
+  };
 
   container.addEventListener('dragover', handleDrop);
 
@@ -93,4 +106,28 @@ export default function Playground(options) {
   this.destroy = function() {
     this.emit('destroy');
   };
+
+  this.attachEditorContainer = withRef(function(node) {
+    return ref.attachEditorContainer(node);
+  });
+
+  this.attachPreviewContainer = withRef(function(node) {
+    return ref.attachFormContainer(node);
+  });
+
+  this.attachDataContainer = withRef(function(node) {
+    return ref.attachDataContainer(node);
+  });
+
+  this.attachResultContainer = withRef(function(node) {
+    return ref.attachResultContainer(node);
+  });
+
+  this.attachPaletteContainer = withRef(function(node) {
+    return ref.attachPaletteContainer(node);
+  });
+
+  this.attachPropertiesPanelContainer = withRef(function(node) {
+    return ref.attachPropertiesPanelContainer(node);
+  });
 }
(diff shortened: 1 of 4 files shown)


pinussilvestrus:
That's still requiring https://github.com/bpmn-io/form-js/pull/291 to be merged, but feel free to already have a look at the changes.

pinussilvestrus:
Ready for review now.

pinussilvestrus:
@Skaiir @barmac @philippfromme, maybe any chance to get a review for this? 🙂

barmac:
I missed this. Thanks for notice.

packages/form-js-playground/src/Playground.js
```diff
@@ -36,7 +37,9 @@ export default function Playground(options) {
 
   container.classList.add('fjs-pgl-parent');
 
-  parent.appendChild(container);
+  if (parent) {
```
barmac, line 40: This means that we should update the type definition so that `container` is optional.

barmac: Here: https://github.com/bpmn-io/form-js/pull/298/commits/5c712c143cf4e9a5b8b3b16a5562a3a46f0e5b96#diff-a7a6e2acf8e80a6b7daf321655518c1fc00e22b1b9570b93ce805db0937db1dcR11

barmac: (I can't add suggestion via GH)

packages/form-js-playground/test/spec/Playground.spec.js
```diff
@@ -207,4 +299,143 @@ describe('playground', function() {
...
+      {
+        name: 'data',
+        attachFn: 'attachDataContainer',
+        selector: 'cm-editor'
+      },
+      {
+        name: 'result',
+        attachFn: 'attachResultContainer',
+        selector: 'cm-editor'
+      },
```
barmac, line 334: Should these two have the same selector?

packages/form-js-playground/test/spec/Playground.spec.js
```diff
@@ -207,4 +299,143 @@ describe('playground', function() {
...
+          playground[attachFn](parent);
+        });
+
+        const expectedContainer = domQuery(`.${selector}`, parent);
+
+        // then
+        expect(expectedContainer).to.exist;
+
+        // cleanup
+        document.body.removeChild(parent);
```
barmac, line 389: Note that this won't fire if test case fails, so it can produce false positives. For generated test cases, it would perhaps make sense to declare a `describe` block for each template with `beforeEach` and `afterEach`.

packages/form-js-playground/test/spec/Playground.spec.js
```diff
@@ -207,4 +299,143 @@ describe('playground', function() {
...
+        });
+      });
+
+      // then
+      forEach(components, ({ selector, parent }) => {
+        expect(domQuery(`.${selector}`, parent)).to.exist;
+      });
+
+      // cleanup
+      document.body.removeChild(testContainer);
```
barmac, line 436: Same goes here for "it does not clean up when test case fails".

barmac (changes requested):
Let's make sure the test cases don't cause us pain in the future, and also make sure TS users can consume new API :)

packages/form-js-playground/test/spec/Playground.spec.js
pinussilvestrus: I thought about adding a custom class name to each Code Mirror instance. However, it was not possible (at least not clearly documented). That's why I keep things as they are; you can still handle/style each editor via parent selectors.

packages/form-js-playground/test/spec/Playground.spec.js
pinussilvestrus: Good catch! Done 👍

packages/form-js-playground/test/spec/Playground.spec.js
barmac: I was afraid that having the same selector in the tests can make us think data container is attached while it's the result container. If that's not the case, we can ignore it.

pinussilvestrus: As we now remove every container after each test, it shouldn't be the case.

pinussilvestrus: I think we are fine.

barmac (approved):
Let's merge it!

PR #299, opened 2022-08-05
docs: add build and run instructions

pinussilvestrus:
Similar to what we have in other projects.


marstamm (approved):
Makes sense, let's add it

PR #302, opened 2022-08-05
feat: use template engine mustache to render textfield

### LINKED ISSUE #247: Be able to bind inputs to readonly text

As a user I want to bind form input to readonly text, potentially formatted using Markdown.

I can model a field to bind a single variable to a readonly text representation.

We shall consider how we deal with these special _readonly_ variables; shall we still include them in the result? What SAFE properties do we want to guarantee? Cf. [blog post](https://bpmn.io/blog/posts/2022-form-js-default-values-improved-submit.html).

Re-use the existing text component to allow variable binding. This would allow to mix and match a number of variables.

__Additional context__

Originally captured in [this conversation](https://camunda.slack.com/archives/C034N0E5YCX/p1650414431837809).

----

Requested via SUPPORT-12403

Related to https://github.com/bpmn-io/form-js/issues/200

jonathanlukas:
Closes #247 

Markdown coming from the text area will now be rended against a template context if available.

[Mustache](https://github.com/janl/mustache.js/) is used as template engine which is also stated in the description of the text area now.

### DIFF
--- a/packages/form-js-editor/src/render/components/properties-panel/entries/TextEntry.js
+++ b/packages/form-js-editor/src/render/components/properties-panel/entries/TextEntry.js
@@ -51,7 +51,7 @@ function Text(props) {
 
   return TextAreaEntry({
     debounce,
-    description: 'Use Markdown or basic HTML to format.',
+    description: 'Use Markdown or basic HTML to format. You may also use Mustache Template Language.',
     element: field,
     getValue,
     id,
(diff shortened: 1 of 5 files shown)


nikku:
Thanks for your PR @jonathanlukas!

Could you add a test case to show how the binding is meant to be used?

On our end we need to clarify what template language we want to use for templating, and support that language full stack.

Mustache is a strong player due to it's safety and simplicity, but it is also fairly _opinionated_ (cf. [docs](https://github.com/janl/mustache.js)).

barmac:
Closing this as a stalled PR. Feel free to comment once you have time to look into this again.

PR #306, opened 2022-08-16
chore: rename 'values' to 'options'

### LINKED ISSUE #304: Rename the values properties to options for selection and tag fields
__What should we do?__

Rename all occurrences of "value" to "option" where actually select options (or choices) are meant.

E.g. 
Bildschirmfoto 2022-08-08 um 17 53 11

to

Bildschirmfoto 2022-08-08 um 17 55 31

respectively

Bildschirmfoto 2022-08-08 um 17 59 36

__Why should we do it?__

The value(s) of a field is/are the finally selected item(s) (the outputs), not the selectable ones. Additionally, we should stay close to the HTML standard (https://developer.mozilla.org/de/docs/Web/HTML/Element/option)

smbea:
Closes #304

We have many internal references to "values" in variables, functions and component names. I think we can keep those to make this easier and only change the displaying values

### DIFF
--- a/packages/form-js-editor/src/render/components/properties-panel/groups/ValuesGroups.js
+++ b/packages/form-js-editor/src/render/components/properties-panel/groups/ValuesGroups.js
@@ -26,7 +26,7 @@ export default function ValuesGroups(field, editField) {
   const groups = [
     {
       id: valuesSourceId,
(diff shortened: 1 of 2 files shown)


Skaiir (approved):
LGTM 👍

PR #307, opened 2022-08-17
feat: add component stress testing command

### LINKED ISSUE #288: Move component stress tests out of the main form definition & add flag to start a stress tested instance

Skaiir:
Closes #288


nikku (commented):
I like this change a lot!

image

package.json
```diff
@@ -27,7 +27,8 @@
     "start": "npm run start:pkg -- @bpmn-io/form-js",
     "start:viewer": "npm run start:pkg -- @bpmn-io/form-js-viewer",
     "start:editor": "npm run start:pkg -- @bpmn-io/form-js-editor",
-    "start:playground": "npm run start:pkg -- @bpmn-io/form-js-playground"
+    "start:playground": "npm run start:pkg -- @bpmn-io/form-js-playground",
+    "stress": "cross-env SINGLE_START=stress npm run dev -- @bpmn-io/form-js"
```
nikku, line 31: Should we rename this to `start:viewer:stress` or something?

nikku (changes requested):
Let's rename the `stress` command to follow existing patterns.

Skaiir:
@nikku Do you mean to start:viewer:stress? I think I did that.

nikku:
Perfect, did miss your change!

nikku (approved):

PR #308, opened 2022-08-17
fix: use outlines for field focus states

### LINKED ISSUE #267: Minor fix: use outlines instead of borders for input field focus states
__Describe the Bug__

Currently we use a border which we increase to 2px width when a field is focused. While the styling has been set up so that this shouldn't be much different from an outline, in some cases it causes tiny changes in page layout which make some items move around by a fraction of a pixel. This is particularly visible when tabbing around:

firefox_hvYq2SeIZq

A simple solution here would be to use `outline` instead.

__Steps to Reproduce__

- `npm run start`
- Tab around
- Watch some elements move

__Expected Behavior__

Focusing elements shouldn't affect page layout. 

__Environment__

- Host Firefox
- OS: W10 x64
- Library version: 0.7.1

Skaiir:
Closes #267

### DIFF
--- a/packages/form-js-viewer/assets/form-js.css
+++ b/packages/form-js-viewer/assets/form-js.css
@@ -189,9 +189,7 @@
 .fjs-container .fjs-button[type='reset']:focus,
 .fjs-container .fjs-textarea:focus,
 .fjs-container .fjs-select:focus {
-  outline: none;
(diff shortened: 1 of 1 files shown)


nikku (approved):
I verified it works! :muscle:

PR #313, opened 2022-08-18
feat: add mac and windows CI checks

### LINKED ISSUE #312: Add windows and mac CI checks

When I encounter a CI bug on Windows locally I cannot tell if it is due to a local problem or something else broke it. Just to improve dev experience and to cover all cases, we should also run form-js through windows and mac CI.

Using GHW, we should pass form-js through CI and on both windows and mac and not only linux.

n/a

Skaiir:
Closes #312

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,7 +5,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ ubuntu-latest ]
+        os: [ macos-latest, ubuntu-latest, windows-latest ]
         node-version: [ 14 ]
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


nikku:
Why would we want to do that? What additional insights do we expect to gain?

Skaiir:
@nikku When I am developing on Windows and I come across an error during a build, it's completely unclear to me if this is a me issue or if a previous change from someone else caused them. With CI, I can safely know that it's going to be on my end. This change forces us to support windows development and will avoid us accidentally breaking development for these other OSes, an issue which I've encountered many times across some of our linux-only CI repos.

nikku:
Fair enough! :)

nikku (approved):
Good change! :heart: :window:

Issue #305, opened 2022-08-08
Explanatory tooltip for new dynamic input options


christian-konrad:
__Is your feature request related to a problem? Please describe__

Users will not understand the new dynamic input options feature, even not when using the playground, since there is no explanation of how to define the input values.

__Describe the solution you'd like__

A tooltip explains how the schema of the input attribute looks like.

![explanation_tooltip_propertiespanel](https://user-images.githubusercontent.com/56470362/183469663-a4a7564f-54ab-4589-9f18-3c9f44dd1b33.png)

__Describe alternatives you've considered__

__Additional context__

christian-konrad:
I could imagine having this as a generic feature for the properties panel.

pinussilvestrus:
As a first step, would that also work by extending the description below the input field?

christian-konrad:
Related to: https://github.com/bpmn-io/form-js/issues/303

christian-konrad:
@pinussilvestrus @smbea having it in the description is totally fine for now.

christian-konrad:
Related to https://github.com/camunda/team-hto/issues/57

Skaiir:
https://github.com/bpmn-io/form-js/pull/316 was Bea's original description contribution, just going to post this here as I'm closing the original ticket (https://github.com/bpmn-io/form-js/issues/303).

christian-konrad:
@Skaiir what's the progress here?

pinussilvestrus:
This can be fixed once https://github.com/bpmn-io/properties-panel/pull/258 is merged and integrated.

Skaiir:
Related: https://github.com/bpmn-io/form-js/issues/784

Issue #303, opened 2022-08-08
Missing value attribute in process variables as tag list options breaks the tag list


christian-konrad:
__Describe the Bug__

When I feed process variables as tag list options, I must specify a list of key-value pairs, where I must provide both a `value` and a `label` for each select option. If I miss a `value`, it breaks the tag list: the value is still offered as a choice, but if I select it, it won't appear in the list of tags, but adds a `null` value to the output.

Additionally, the hint in the properties panel does not help me to understand in which format I need to provide the select options.

__Steps to Reproduce__

Best to be done in the playground.

1. Provide a list of maps that contains at least one select option entry that does not have a `value` attribute
2. Assign this map to a tag list component as the "value" input (actually the options input, name is confusing here)
3. Select the item in the tag list and check the output

Also see the video.

https://user-images.githubusercontent.com/56470362/183446847-ee3864b4-a91a-4c38-b648-85b00e7864f7.mov

__Expected Behavior__

In addition to fix the faulty behavior itself, we should add a hint in the properties panel on how to configure it: https://github.com/bpmn-io/form-js/issues/305

__Environment__

- Host (Browser/Node version), if applicable: [e.g. MS Edge 18, Chrome 69, Node 10 LTS]
- OS: [e.g. Windows 7]
- Library version: [e.g. 2.0.0]

PR #315, opened 2022-08-18
fix(viewer): filter out invalid dynamic taglist options

smbea:
Related to #303

For the properties panel hint: #305 is for adding a more sophisticated solution, but for now we decided to improve the existing description:

Screenshot 2022-08-18 at 14 41 32

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Taglist.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Taglist.js
@@ -64,7 +64,7 @@ export default function Taglist(props) {
 
   useEffect(() => {
     if (loadState === LOAD_STATES.LOADED) {
-      setFilteredValues(options.filter((o) => o.label && (o.label.toLowerCase().includes(filter.toLowerCase())) && !values.includes(o.value)));
+      setFilteredValues(options.filter((o) => o.label && o.value && (o.label.toLowerCase().includes(filter.toLowerCase())) && !values.includes(o.value)));
     }
     else {
       setFilteredValues([]);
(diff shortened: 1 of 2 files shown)


packages/form-js-editor/src/render/components/properties-panel/entries/InputKeyValuesSourceEntry.js
```diff
@@ -10,12 +10,18 @@ export default function InputKeyValuesSourceEntry(props) {
     id
   } = props;
 
+  const schema = '[{"label": "dollar", "value": "$"}]';
+
+  const description = <div>
+    Define which input property to populate the values from. The property must follow this schema: <code>{schema}</code>.
```
pinussilvestrus, line 16: Should we maybe add the description as mentioned in the issue (https://github.com/bpmn-io/form-js/issues/305)? So having "The property must follow ..." as another paragraph and add some indentation to the code example. 

image

pinussilvestrus: I like this one much more 👍 

Ideally, this should be done by the core (description) styles, yeah, to align it across our tools.

pinussilvestrus: We could also split this PR into two parts (bug fix, description updates) and, make progress with the bug fix faster + deliver the style + description updates in another PR.

pinussilvestrus (approved):
👍 ✅

PR #316, opened 2022-08-19
feat(editor): improve dynamic input field description

smbea:
Related to #303

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -693,9 +693,9 @@
       "link": true
     },
     "node_modules/@bpmn-io/properties-panel": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/@bpmn-io/properties-panel/-/properties-panel-1.2.0.tgz",
-      "integrity": "sha512-J6zya4IvRIeuefNUaWdg2sG89WB5shZX5KkbuZG0/lX4Vr4TLs2FWDMXV57hoKwnRlrPBEv18uueJfNxYvc3EQ==",
+      "version": "1.4.0",
+      "resolved": "https://registry.npmjs.org/@bpmn-io/properties-panel/-/properties-panel-1.4.0.tgz",
+      "integrity": "sha512-M5WsC08dFjN2cvhCNrCgy7ci7ZGERBDoX97FuhTkPfCDn7ZNQc3XweQLbz7puJy7fFGEN4eitvfo6CxnIBfHAw==",
       "dependencies": {
         "@bpmn-io/feel-editor": "^0.7.0",
         "classnames": "^2.3.1",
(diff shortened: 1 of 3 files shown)


pinussilvestrus:
For transparency (why this is still in draft): We are lacking support in `properties-panel` to properly display `code` elements in descriptions.

Context: https://github.com/bpmn-io/form-js/pull/315#discussion_r950004707

pinussilvestrus:
I went on simply adding the missing parts via https://github.com/bpmn-io/properties-panel/pull/219. @christian-konrad @RomanKostka feel free to share your feedback; I did not modify anything specific to the existing properties panel styles.

image

nikku:
Not exactly sure why this is necessary at the core. You can always override stuff, and once you feel it is a mature solution (for a given context) _then_ we could ship a solution at the core. 

But either way awesome that you pick up this low hanging fruit @pinussilvestrus :medal_sports: .

christian-konrad:
@pinussilvestrus I also created this small PR https://github.com/bpmn-io/form-js/pull/543, so probably we should adjust the message

pinussilvestrus:
> @pinussilvestrus I also created this small PR #543, so probably we should adjust the message

We can adjust the message once we settle on something in the context of #543. This should not block us here.

pinussilvestrus:
This is ready for review now.

vsgoulart (commented):
Found this issue we discussed through Slack
it seems to not be happening on `develop`, so it could potentially be some CSS leaking from the properties panel update

image

vsgoulart (commented):
It seems to be working well, but I think the validation is not ideal.
If I add a value like `[{"label": "bar", "value": "foo"}]` it complains because of the spaces even though it's a valid value. It's specially confusing since if you just copy and paste the value from the description it shows an error

Maybe we should use a [function](https://github.com/camunda/tasklist/blob/master/client/src/modules/utils/isValidJSON.tsx) like Tasklist to valid this JSON array?

pinussilvestrus:
Where did you paste the JSON? Note that the "Input values key" only accepts a variable name, and that variable needs to follow this structure, not the key.

image

Should we update the description to make that clearer? /cc @RomanKostka @christian-konrad

vsgoulart:
> Where did you paste the JSON? Note that the "Input values key" only accepts a variable name, and that variable needs to follow this structure, not the key.
> 
> image
> 
> Should we update the description to make that clearer? /cc @RomanKostka @christian-konrad

Ah, ok. I misinterpreted it :sweat_smile:

vsgoulart (approved):
I think  changing the label would be better, but if you think it's not necessary the rest looks ok to me

pinussilvestrus:
Thanks for the review! I'd let it open for @christian-konrad and @RomanKostka whether we want to further improve the description + labels.

RomanKostka (approved):
nice!
> I think changing the label would be better, but if you think it's not necessary the rest looks ok to me
i agree and would also vote for the singular label.
Besides that it looks fine for me

pinussilvestrus:
Regarding the label: Personally I'd still be in favor of the plural. You are defining *options* with the key, not a *single value*.

PR #333, opened 2022-09-19
chore(CHANGELOG): update to v0.9.0

pinussilvestrus:
This minor release includes mostly `form-js-playground` changes. As there is no schema update, we do not need to inform other product teams specifically. 

/cc @Skaiir @CatalinaMoisuc @christian-konrad

### DIFF
--- a/packages/form-js/CHANGELOG.md
+++ b/packages/form-js/CHANGELOG.md
@@ -7,6 +7,20 @@ All notable changes to [form-js](https://github.com/bpmn-io/form-js) are documen
 ___Note:__ Yet to be released changes appear here._
 
 
+## 0.9.0
+
+### Editor
+
+* `CHORE`: show palette action names in compact mode ([`fc41baf9`](https://github.com/bpmn-io/form-js/commit/fc41baf97b4739ac8f688ea4dda3290f7d6dee8b))
+
+### Playground
+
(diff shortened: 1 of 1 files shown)


Skaiir:
@pinussilvestrus Actually this might involve some cross team concerns. It will trigger an auto-update on the web-modeler side and as long as they don't have linting this might cause trouble as the 0.8 features would not be functional. Although technically that's also true of many versions of form-js prior to that one. 

There also was a properties panel issue when updating. @CatalinaMoisuc what do you think?

pinussilvestrus:
We shouldn't be blocked from releasing core libraries because products use auto-updating. Also, I think the web modeler already stopped while auto-upgrading is not possible: https://github.com/camunda/web-modeler/issues/2383

Skaiir:
@pinussilvestrus True, just thought this might need some communication but fair enough it was properly blocked, I see 👍

nikku (approved):

pinussilvestrus:
Released as v0.9.0: https://github.com/bpmn-io/form-js/commit/10ea539989ea857adc7511e5f1a80a99c7ec96b1.

pinussilvestrus:
I strongly agree on giving a heads-up when we have breaking changes. For this release, there aren't any, as the forms playground is not used in the Web Modeler. :+1:

PR #343, opened 2022-09-28
deps: upgrade to `properties-panel@0.21.0`

Skaiir:
Related to https://github.com/camunda/camunda-modeler/pull/3133

Tested this pretty thoroughly by copying over the dependencies. I was not able to replicate the headers bug in the modeler integration, so this should cover the fix


marstamm (approved):
Can no longer reproduce, lets go

PR #349, opened 2022-10-06
fix(editor): ensure valueSource is undoable

marstamm:
related to https://github.com/camunda/camunda-modeler/issues/3215


pinussilvestrus (approved):
Works for me 👍

PR #351, opened 2022-10-10
fix(editor): do not order values alphanumerical

### LINKED ISSUE #350: Different order of checklist items in form and properties panel
__Describe the Bug__

The order of checklist items in the form and the properties panel is different. While they appear in the form in the order of how they've been added, they appear in reverse order in the properties panel, and when collapsing the section and reopening it, they appear in lexicographical order.

static-forms-order

__Steps to Reproduce__

See video:

https://user-images.githubusercontent.com/56470362/194884814-6d8c9b79-45a2-408e-8958-b154395bdef7.mov

__Expected Behavior__

Order in the form and the panel is always the same.

__Environment__

- Camunda Modeler 5.4.1

pinussilvestrus:
Closes #350

... because it doesn't make sense. We have to display them as they occur.

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/StaticValuesSourceEntry.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/StaticValuesSourceEntry.js
@@ -70,6 +70,7 @@ export default function StaticValuesSourceEntry(props) {
 
   return {
     items,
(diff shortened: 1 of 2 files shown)


barmac (approved):
I hope one day we make `shouldSort: false` the default value.

PR #369, opened 2022-10-21
fix: allow click through taglist input in editor

### LINKED ISSUE #365: Allow clicking through taglist pseudo-input in editor
__Describe the Bug__

When clicking on the taglist pseudo-input area _(highlighted red below)_, the click is intercepted which makes selecting it in the editor harder than for most control.

image

__Steps to Reproduce__

1. drag a taglist into the editor
2. try and select it by clicking in the above selected area (doesn't work)
3. drag a text input into the editor
4. try and select it the same way (works)

__Expected Behavior__

The click events should be entirely passed through when clicking on any part of the control

Skaiir:
Closes #365

Pretty trivial fix, won't provide tests for this one as it's just not worth it

### DIFF
--- a/packages/form-js-editor/assets/form-js-editor.css
+++ b/packages/form-js-editor/assets/form-js-editor.css
@@ -112,6 +112,7 @@
 }
 
 .fjs-editor-container .fjs-input:disabled,
+.fjs-editor-container .fjs-taglist-input:disabled,
 .fjs-editor-container .fjs-select:disabled {
   pointer-events: none;
 }


Skaiir:
The reason we use a different class for taglists is that `.fjs-input` comes with a lot of styling baggage that we don't want in the preview. It would look okay in the editor but that's because I just render it as a regular disabled input there for simplicity :).

pinussilvestrus (approved):
It makes totally sense to have another selector for this type of input. It's just something different (cf. https://github.com/bpmn-io/form-js/pull/369#issuecomment-1286749037)!

PR #376, opened 2022-10-27
fix(viewer): use blue with more contrast

pinussilvestrus:
Discovered while working on https://github.com/camunda/team-hto/issues/28. Links and submit buttons have too less contrast against a white background. I propose to increase that a bit. /cc @RomanKostka @christian-konrad 

It doesn't make a big impact visually, but a big impact a11y wise :+1: 

_Before_

image

image

_After_

image

image

### DIFF
--- a/packages/form-js-viewer/assets/form-js.css
+++ b/packages/form-js-viewer/assets/form-js.css
@@ -12,6 +12,7 @@
   --color-grey-225-10-95: hsl(225, 10%, 95%);
   --color-grey-225-10-97: hsl(225, 10%, 97%);
 
+  --color-blue-205-100-40: hsl(205, 100%, 40%);
   --color-blue-205-100-45: hsl(205, 100%, 45%);
   --color-blue-205-100-50: hsl(205, 100%, 50%);
   --color-blue-205-100-80: hsl(205, 100%, 80%);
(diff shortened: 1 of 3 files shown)


packages/form-js-viewer/assets/form-js.css
```diff
@@ -36,8 +37,8 @@
   --color-borders: var(--color-grey-225-10-55);
   --color-borders-disabled: var(--color-grey-225-10-75);
   --color-warning: var(--color-red-360-100-45);
-  --color-accent: var(--color-blue-205-100-45);
-  --color-accent-dark: var(--color-blue-205-100-45);
+  --color-accent: var(--color-blue-205-100-40);
```
Skaiir, line 40: Accent and accent dark have the same shade, is this intentional? If not we should merge the tag.

pinussilvestrus: It was the same before, not sure whether it's intentional though. From the naming I assume it was meant to be `darker`, so maybe just let's use `--color-accent-dark: var(--color-blue-205-100-35);`

_Not focussed_
image

_Focussed_
image

pinussilvestrus: Alternative: remove the border completely and keep the browser default on focus for buttons.

pinussilvestrus: Just found out that we are setting the border-color to prevent using the default `--color-borders` for other inputs, which is a grey tone. 

Let's simply remove the `--color-accent-dark` which doesn't have any value and simply use `--color-accent` to keep things as they are.

Cf. https://github.com/bpmn-io/form-js/pull/376/commits/e45eba75d7705fbe5cdf50e251c45339d3fc4019

Skaiir (approved):

PR #379, opened 2022-10-31
feat: <textarea> form element implementation

### LINKED ISSUE #283: Enable users to configure multiline text inputs

As part of our initiative to enable more use cases in form-js, we want to allow users to enter larger chunks of text via a multi-line text component.

There are two main ways we can consider implementing this:

1. simply make the regular text-field multi-line capable via configuration 
2. create an entirely new component for multi-line inputs

Solution 1 is the fastest way to do this, the downside relative to solution 2 would be that multi-line is a different html component, and so it might make sense to have that difference represented in the palette as well. It also make the base component for text a tiny bit more complex, but honestly it's minor.

I would prefer solution 1, it avoids quite a bit of overhead and I also think it makes sense for a form developer.

__Additional context__

[Product hub issue](https://github.com/camunda/product-hub/issues/97)

Skaiir:
Closes #283


package-lock.json
```diff
@@ -1199,18 +1199,6 @@
         "js-yaml": "bin/js-yaml.js"
       }
     },
-    "node_modules/@eslint/eslintrc/node_modules/minimatch": {
```
pinussilvestrus, line 1202: How are these changes related? Do we have any dependency updates?

packages/form-js-editor/src/features/palette/components/Palette.js
```diff
@@ -5,6 +5,10 @@ const types = [
     label: 'Text Field',
     type: 'textfield'
   },
+  {
+    label: 'Text Area',
```
pinussilvestrus, line 9: Maybe we should update our field labels to [use sentence case](https://camunda.com/brand/writing-style-guide/).

> Text field
> Text area

packages/form-js-editor/src/features/properties-panel/Util.js
```diff
@@ -41,7 +41,8 @@ export const INPUTS = [
   'radio',
   'select',
   'taglist',
-  'textfield'
+  'textfield',
+  'textarea',
```
pinussilvestrus, line 45:
```suggestion
  'textarea'
```

packages/form-js-editor/src/features/properties-panel/entries/DefaultValueEntry.js
```diff
@@ -72,6 +74,14 @@ export default function DefaultOptionEntry(props) {
     });
   }
 
+  if(type === 'textarea') {
```
pinussilvestrus, line 77: `npm run lint -- --fix` should do the trick here 👍

packages/form-js-viewer/src/render/components/form-fields/Textfield.js
```diff
@@ -68,4 +68,4 @@ Textfield.type = type;
 Textfield.label = 'Text Field';
 Textfield.keyed = true;
 Textfield.emptyValue = '';
-Textfield.sanitizeValue = ({ value }) => (isArray(value) || isObject(value)) ? null : String(value);
\ No newline at end of file
+Textfield.sanitizeValue = ({ value }) => (isArray(value) || isObject(value)) ? '' : String(value);
```
pinussilvestrus, line 71: Can you describe why this change is needed? Do we have a test case for this?

packages/form-js-viewer/src/render/components/form-fields/Textarea.js
```diff
@@ -0,0 +1,89 @@
...
+  const { required } = validate;
+
+  const onInput = ({ target }) => {
+
+    props.onChange({
+      field,
+      value: target.value
+    });
+
+    autoSizeTextarea(target);
```
pinussilvestrus, line 42: When I first type in something, the text area got's expanded immediately, although it's not needed. Not sure whether it's a [playground demo](https://demo-textarea--camunda-form-playground.netlify.app/) issue though.

Kapture 2022-11-01 at 13 33 57

packages/form-js-viewer/src/render/components/form-fields/Textarea.js
```diff
@@ -0,0 +1,89 @@
...
+
+  };
+
+  const autoSizeTextarea = (textarea) => {
```
pinussilvestrus, line 46: I assume auto-resize is our general strategy here right? So we accept that it's not configurable (e.g., to have a fixed height only).

packages/form-js-viewer/src/render/components/form-fields/Textarea.js
```diff
@@ -0,0 +1,89 @@
...
+
+  const { formId } = useContext(FormContext);
+
+  return <div class={ formFieldClasses(type, errors) }>
+    <Label
+      id={ prefixId(id, formId) }
+      label={ label }
+      required={ required } />
+    <textarea class="fjs-textarea"
+      disabled={ disabled }
```
pinussilvestrus, line 70: The auto resize is not working on disabled text areas (the scroll bar is visible).

image

packages/form-js-viewer/test/spec/render/components/form-fields/Textarea.spec.js
```diff
@@ -0,0 +1,255 @@
...
+} from '../../../../TestHelper';
+
+import { WithFormContext } from './helper';
+
+const spy = sinon.spy;
+
+let container;
+
+
+describe('Textarea', function() {
```
pinussilvestrus, line 20: Don't we have any test coverage for the auto resize behavior?

packages/form-js-viewer/src/render/components/form-fields/Textarea.js
pinussilvestrus: The same happens when using larger input data for a text area.

image

pinussilvestrus (changes requested):
Looks good so far functionality and code wise, just some minor comments 👍 Let's also ensure we test the auto resize behavior (if possible).

package-lock.json
Skaiir: They are not, but I did run an audit and fix as we had critical vulnerabilities.

packages/form-js-viewer/src/render/components/form-fields/Textarea.js
Skaiir: We can add this as a visual configuration in the future, but for the initial version of this, I don't think it's worth it.

package-lock.json
pinussilvestrus: Ah alright. Let's mark this in an additional commit.

packages/form-js-viewer/src/render/components/form-fields/Textfield.js
Skaiir: As I built out the `textArea`, I noticed that I was sanitizing the `textField` to a null value. This is however incorrect as we use empty string as our empty state for `textField`, never null. 

This addresses it and yes we do have a test case for it: https://github.com/bpmn-io/form-js/blob/183e8cd911da92c9bafe38e0206e06e27b8d03cc/packages/form-js-viewer/test/spec/import/Importer.spec.js#L326

packages/form-js-viewer/src/render/components/form-fields/Textarea.js
Skaiir: Behavior is now very consistent

packages/form-js-viewer/test/spec/render/components/form-fields/Textarea.spec.js
Skaiir: There now is.

pinussilvestrus (approved):
Looks good to me 🎉

pinussilvestrus:
For some reason firefox is broken: https://github.com/bpmn-io/form-js/actions/runs/3387857007/jobs/5638190490#step:6:128

I will investigate whether this is related to this PR or not.

pinussilvestrus:
It's not related to this branch, so let's merge and fix at another place.

PR #385, opened 2022-11-04
Image view component with expression properties

### LINKED ISSUE #383: Implement image view component

Implement image view component.

image

* [ ] Add the possibility to declare image `src`
  * [ ] from process variables (FEEL expression) or static value as a (data) URI or link (`source`)
  * [ ] <del>uploaded into the form definition as default value via properties panel (`defaultValue`)</del>
* [ ] Add the possibility to declare `alt` text (similar to `source`)
* [ ] Display a default image in the editor
* [ ] Add image container to make layout configurable in future iterations
  * [ ] fit container to given image dimensions 
* [ ] Update [`getSchemaVariables`](https://github.com/bpmn-io/form-js/blob/a9113d5a38511fb1875636ccfe3487130f9ba10d/packages/form-js-viewer/src/util/index.js#L73) to detect `source` and `alt` variables
* [ ] ...

Child of https://github.com/camunda/team-hto/issues/42

pinussilvestrus:
Closes #383 
Depends on https://github.com/bpmn-io/form-js/pull/403

* Provide image `src` via `source` (static or via FEEL)
* Provide image `alt` text (static or via FEEL)

[Demo](https://demo-image-view--camunda-form-playground.netlify.app/)
[Docs updates](https://github.com/camunda/camunda-platform-docs/compare/form-image-component-docs?expand=1) (to be updated after latest changes)
[Lint rule updates](https://github.com/camunda/form-linting/pull/22)

Kapture 2022-11-17 at 15 35 16


Skaiir:
Sorry I didn't comment about this earlier, the component went through too fast 😆 

On `DefaultValue` and `Source`:

IMO we are confounding concepts a little bit here. Default value across all our proper form fields refers to an initial state which can be modified via interaction with the field. This is different from the way it's being used here, and I think we should keep the concept of `DefaultValue` tightly bound to that of `Value`.

The way things stand right now: `DefaultValue` is a static definition of an image, and `Source` is a dynamic one, based on input variables. The sort of relationship we're designing here is the same as the one we required for the dynamic data story, but seems to be basing itself on the `Value` / `DefaultValue` pairs found on form fields.

But for options, static and dynamic are treated with the same level of importance. They live in the same group and are conceptually put on the same level.

Here we have a major separation between the two ways of defining the source, I think this shouldn't be the case. 

**Some ideas:**

What we ideally would have is feel evaluation and one single property to define the image source, whether it's static or dynamic.

`source: "http://...."` OR `source:  "=someFeelExpression"`

But this isn't quite there yet.

- What we could do in the meantime is have two properties for this (as it currently is done)
 	- source: takes in a static http link (what is currently being called default value)
 	- sourceRef: takes in a variable (which itself contains the link)
- We could then, without breaking the backwards compatibility, feel-enable the source property some time in the future

**My two main points:**

(1) Let's really analyze the schema commitments we're willing to make at this point, feature toggles are always best avoided. @christian-konrad definitely would like to hear your input on this here because I'm pretty sure you foresee feel-evaluation as becoming pretty much ubiquitous in the semi-near future.

(2) I think `defaultValue`, `value`, `key` are all intermingling concepts which should only exist in relation to `form fields`. Image isn't a form field, it's just an element, and so we shouldn't be associate it with these already established concepts. 

This is just general thoughts on the properties. I'll still review the implementation. Let me know what you think.

packages/form-js-editor/src/features/properties-panel/PropertiesPanelHeaderProvider.js
```diff
@@ -30,6 +31,10 @@ export const PropertiesPanelHeaderProvider = {
       return textToLabel(field.text);
     }
 
+    if (type === 'image') {
+      return null;
```
Skaiir, line 35: Perhaps the `altText` would make sense here?

Skaiir: image

packages/form-js-editor/src/features/properties-panel/entries/AltTextEntry.js
```diff
@@ -0,0 +1,60 @@
+import { get } from 'min-dash';
```
Skaiir, line 1: I think this is an oversight: If the `source` can be defined through input data, then we also need a mechanism to define `altText` through input data, because the whole point of using input data is the flexibility of not knowing what the image will be. From an a11y standpoint I don't think we can make due without it.

packages/form-js-editor/src/features/properties-panel/entries/DefaultValueEntry.js
```diff
@@ -29,7 +29,11 @@ export default function DefaultOptionEntry(props) {
   const entries = [];
 
   // Only make default values available when they are statically defined
-  if (!INPUTS.includes(type) || VALUES_INPUTS.includes(type) && !field.values) {
+  if (
+    !INPUTS.includes(type) &&
+    type !== 'image' ||
```
Skaiir, line 34: This is one of the symptoms of those intermingling concepts I was referring to in my large initial point on `defaultValue`, it's codebase structure I think we shouldn't try to hack our way around. Value, defaultValue ect are all related to the lifecycle of form field state which image is not.

Note that `Text` doesn't need an exception, because it creates its own property (text) to represent its value. I think we should do the same.

packages/form-js-viewer/src/render/components/form-fields/Image.js
```diff
@@ -0,0 +1,62 @@
...
+
+Image.create = function(options = {}) {
+  return {
+    ...options
+  };
+};
+
+Image.type = type;
+Image.keyed = false;
+Image.sanitizeValue = ({ value }) => (isArray(value) || isObject(value)) ? '' : String(value);
```
Skaiir, line 62: `sanitizeValue` is only run for form fields, hence it's dead code here.

packages/form-js-viewer/src/render/components/form-fields/Image.js
```diff
@@ -0,0 +1,62 @@
...
+    alt,
+    id
+  } = field;
+
+  const { source } = useImageSource(field);
+  const safeSource = safeImageSource(source);
+
+  const { formId } = useContext(FormContext);
+
+  // todo(pinussilvestrus): what to display if image broken?
```
Skaiir, line 34: Should we answer that question or have we already accepted something for now?  Is this PR aimed at dev or the sprint branch, because that changes the standards :D.

Skaiir (changes requested):
For demo purposes on the sprint branch, this would work totally fine, but I think there are still some questions to be answered before we can commit to a dev release and to a schema on this one.

pinussilvestrus:
Good catch about the default value! I like the `source` vs. `sourceRef` approach, let's go for it.

packages/form-js-viewer/src/render/components/form-fields/Image.js
pinussilvestrus: It's an oversight (sorry!), let's remove the comment. We agreed to simply display the browser default on broken images.

pinussilvestrus:
I updated the target branch to `design-sprint` so we can also finalize things after the sprint week. Still, let's fix the obvious things before merging this PR (avoid using defaultValue).

packages/form-js-editor/src/features/properties-panel/entries/AltTextEntry.js
pinussilvestrus: Good point. Maybe let's follow the same direction here:
* alt: static value (and potentially FEEL later)
* altRef: key to bind input data

packages/form-js-editor/src/features/properties-panel/PropertiesPanelHeaderProvider.js
pinussilvestrus: I like that, let's go for it 👍

pinussilvestrus:
An interesting question regarding the `source`/`alt` vs. `sourceRef`/`altRef` approach. Do we need a selection as we have for the select values to specify the source type? Or is it sufficient enough for images to simply have two inputs and always use the `sourceRef` (process variable) over `source` (static value as for now, before we introduce FEEL)?

image

packages/form-js-editor/src/features/properties-panel/entries/DefaultValueEntry.js
pinussilvestrus: As proposed, let's ditch that one completely 👍

packages/form-js-editor/src/features/properties-panel/entries/AltTextEntry.js
Skaiir: It makes sense to me, and we could even get rid of altRef in the future if we figure out a safe mechanism to do migrations. Then it would be expressed as a feel expression instead, same as with source.

pinussilvestrus:
@Skaiir @RomanKostka this is ready for review again. I also [updated the demo](https://demo-image-view--camunda-form-playground.netlify.app/) to make it easier to review 👍

Skaiir:
image

The description is cool but I think this will be rather added with the tooltips. I added something similar to the number field and Christian suggested we'd rather not have these. Check with him to confirm. 

---

image

As mentioned on Slack, we might want to display the alt text when the link is completely broken. 

---

image

Would be nice to align the placeholder, the style below should do the trick

.fjs-container .fjs-image-placeholder {
    height: 64px;
    margin: 2px 0;
}

pinussilvestrus:
Thanks for the feedback!

> The description is cool but I think this will be rather added with the tooltips. I added something similar to the number field and Christian suggested we'd rather not have these. Check with him to confirm.

I synced with @christian-konrad and we agreed to keep the description (but shorter) until we have tooltips to display in this case.

image

> As mentioned on Slack, we might want to display the alt text when the link is completely broken.

Also synced with @christian-konrad on this one. We do not display the alt text with the placeholder but rather consider lint rules to confirm something went wrong with the image source (no source, XSS malformed).

> Would be nice to align the placeholder, the style below should do the trick

Added, thanks!

RomanKostka (approved):
looks good from UI/UX perspective - thanks! 🎉

Skaiir (approved):
Let's merge, nice work :)

PR #387, opened 2022-11-04
Renovate/bpmn io properties panel 0.x

pinussilvestrus:
Also adds `didi` to the editor. With this properties panel version it starts to report missing types. 

As the `viewer` also has the `didi` dependency, I wonder why the `editor` doesn't have it already. 

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
         "cross-env": "^7.0.3",
         "del-cli": "^4.0.0",
         "diagram-js": "^10.0.0",
+        "didi": "^9.0.0",
         "eslint": "^8.22.0",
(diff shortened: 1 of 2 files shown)


packages/form-js-editor/package.json
```diff
@@ -44,6 +44,7 @@
   "dependencies": {
     "@bpmn-io/form-js-viewer": "^0.9.9",
     "array-move": "^3.0.1",
+    "didi": "^9.0.0",
```
barmac, line 47: Perhaps it should go to devDependencies as we use it for types only: https://github.com/bpmn-io/form-js/blob/56f33dba5a6826aaef6b59d81f65c3b68bd78f77/packages/form-js-viewer/src/util/injector.js#L1

Issue #402, opened 2022-11-14
As a form designer, I want to use any data given to form as variables in the condition.


barmac:

barmac:
This needs some significant changes in the library as we don't allow to pass "inputData" to the form editor instance yet.

pinussilvestrus:
An option could be to provide a `config.demoData`  to the formEditor, which can be then used by the `useVariables` hook: https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/src/features/properties-panel/hooks/useVariables.js#L10

Alternatively we could consider a `demoDataProvider` module and provide this via `additionalModules`.

PR #403, opened 2022-11-14
Add conditional rendering

### LINKED ISSUE #374: Implement conditional rendering of forms
__What should we do?__

* [x] As a form designer, I want to define a condition (FEEL unary test) under which field is displayed. 
* [x] As a form designer, I want to have editing support for the condition.
* [x] As a form designer, I want to use other fields' values as variables in the condition. 
* [x] As a user, I want to see only fields which have no condition or for which condition is met.
* [x] As a user, I want to submit only fields which have no condition or for which condition is met.

__Why should we do it?__

We want to allow users to decide whether field should be displayed based on given data.

### LINKED ISSUE #401: `getSchemaVariables` should include variables mentioned in conditions
> As a form designer, I want to use any data given to form as variables in the condition. 

Some context to this, this will require changes to the `getSchemaVariables(schema)` method. C8 will only provide the bare minimum amount of data to the form, and this method is a way for us to tell them what data we rely on. This means that for all feel expressions, we need to somehow extract all the variable references as well. This is crucial for the feature.

I had a look at the WIP code and I noticed that you actually make use of that method to define the legal variables in the feel editor. Even though it wasn't the initial intention for it, it makes a lot of sense. But something which will be needed now is to actually complete the loop and extract variables from your feel expressions in the method. If you want to sync on this, just drop me a message on Monday.

_Originally posted by @Skaiir in https://github.com/bpmn-io/form-js/issues/374#issuecomment-1304045580_

### LINKED ISSUE #415: Use "hideIf" logic for field condition
Based on https://github.com/camunda/product-hub/issues/56#issuecomment-1315173662, I will need to reverse the logic so that we use a `"hideIf"` field. This will also require changes in the properties panel. Let's do that in a separate issue.

_Originally posted by @barmac in https://github.com/bpmn-io/form-js/issues/403#issuecomment-1318835164_

barmac:
This PR implements conditional rendering via a `Field#condition` property. Condition needs to be a FEEL unary test which is parsed and evaluated by [feelin](https://github.com/nikku/feelin). It is defined via the properties panel. Form designer can use any variables passed to the form object to define the condition.

Closes #374
Closes #401
Closes #415

---

Follow-up issues:
* https://github.com/bpmn-io/form-js/issues/402


barmac:
@nikku FEEL free to provide your feedback :)

barmac:
Try this out via `npm run start:playground`.

packages/form-js-viewer/src/Form.js
```diff
@@ -318,7 +319,7 @@ export default class Form {
    * @private
    */
   _emit(type, data) {
-    this.get('eventBus').fire(type, data);
+    return this.get('eventBus').fire(type, data);
```
barmac, line 322:
```suggestion
    this.get('eventBus').fire(type, data);
```

packages/form-js-viewer/test/spec/condition.json
```diff
@@ -0,0 +1,23 @@
...
+        "min": 0,
+        "max": 1000
+      }
+    },
+    {
+      "label": "Text Field",
+      "type": "textfield",
+      "id": "Field",
+      "key": "text",
+      "condition": "=amount > 2"
```
barmac, line 17: Consider renaming.

packages/form-js-playground/rollup.config.js
```diff
@@ -71,6 +74,17 @@ export default [
...
-    ])
+    ]),
+    onwarn
   }
-];
\ No newline at end of file
+];
+
+// TODO(@barmac): remove once https://github.com/moment/luxon/issues/193 is resolved
+function onwarn(warning, warn) {
```
nikku, line 83: Sure that we want to ignore this? Can the built bundle still be consumed (conditions be evaluated)?

packages/form-js-viewer/src/core/ConditionChecker.js
```diff
@@ -0,0 +1,79 @@
...
+    if (!isString(condition) || !condition.startsWith('=')) {
+      return false;
+    }
+
+    try {
+      const result = unaryTest(condition.slice(1), data);
+
+      return result;
+    } catch (error) {
+      return false;
```
nikku, line 52: Consider not swallowing the error, but logging it (or emitting it via `eventBus.fire(`error`, ...)`.

packages/form-js-viewer/src/core/ConditionChecker.js
```diff
@@ -0,0 +1,79 @@
...
+    if (!condition) {
+      return true;
+    }
+
+    if (!isString(condition) || !condition.startsWith('=')) {
+      return false;
+    }
+
+    try {
+      const result = unaryTest(condition.slice(1), data);
```
nikku, line 48: I guess at this point we don't want to enable context sensitive parsing, right? Our inputs don't contain spaces and other special chars.

packages/form-js-viewer/src/core/ConditionChecker.js
```diff
@@ -0,0 +1,79 @@
+import { unaryTest } from 'feelin';
+import { isString } from 'min-dash';
+
+export class ConditionChecker {
```
nikku, line 4: Awesome that this is pluggable. :heart:

packages/form-js-viewer/src/render/hooks/useCondition.js
```diff
@@ -0,0 +1,16 @@
+import useService from './useService.js';
+
+/**
+ *
+ * @param {{ expression?: string } | undefined} condition
```
nikku, line 5: What is the result of this hook?

Return type is missing; unclear if it returns "true" to show or to hide.

nikku (commented):
Great addition (service + hook). Just found some nit-picks you may act upon.

packages/form-js-playground/rollup.config.js
nikku: ==> Let's add an integration test for this in form-js.

packages/form-js-editor/src/features/properties-panel/entries/ConditionEntry.js
```diff
@@ -0,0 +1,95 @@
...
+  const previousVariables = usePrevious(variables) || [];
+
+  const variablesChanged = !areEqual(variables, previousVariables);
+
+  if (variablesChanged) {
+    variablesRef.current = variables.map(key => ({ name: key }));
+  }
+
+  return variablesRef.current;
+}
```
pinussilvestrus, line 81: Should we move this under `/hooks`? I can imagine we will need this more often in the future (similar to `useCondition` in the `viewer`)

packages/form-js-viewer/src/core/ConditionChecker.js
```diff
@@ -0,0 +1,79 @@
...
+   * Check if given condition is met.
+   *
+   * @param {string} condition
+   * @param {import('../types').Data} [data]
+   *
+   * @returns {boolean}
+   */
+  check(condition, data = {}) {
+    if (!condition) {
+      return true;
```
pinussilvestrus, line 40: Should we allow to specify this `defaultValue`? I could imagine other cases where we do not default a condition to `true`, for example: `readonly` => default `false`.

packages/form-js-viewer/package.json
```diff
@@ -41,6 +41,7 @@
     "@bpmn-io/snarkdown": "^2.1.0",
     "classnames": "^2.3.1",
     "didi": "^9.0.0",
+    "feelin": "^0.41.0",
```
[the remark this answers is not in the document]
barmac: We can do that as soon as we want to implement something in the library.

packages/form-js-editor/src/features/properties-panel/entries/ConditionEntry.js
barmac: Hmm I think I could simplify this a bit...

barmac: OK it's now simplified and exported from hooks.

pinussilvestrus: I think we have to be careful here. The `useService` hook that is used in the new `useVariables` hook now needs a `FormEditorContext` to work correctly. Note that the properties panel can also be rendered outside of the editor. We use our own `useService` therefore: https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/src/features/properties-panel/hooks/usePropertiesPanelService.js

We either need the `formEditor` to be provided as a parameter for this hook or, alternatively, move the `useVariables` to the properties panels hooks.

barmac: Yeah this is probably why the tests started to fail after the change.

barmac: OK it should be fine now.

packages/form-js-viewer/src/core/ConditionChecker.js
barmac: Not sure what you mean. I need to pass data in order to correctly evaluate the condition, right?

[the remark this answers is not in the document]
barmac: I added a comment.

packages/form-js-viewer/src/core/ConditionChecker.js
barmac: Implemented via https://github.com/bpmn-io/form-js/pull/403/commits/848b703812998668b302397f73c8b416abbb094a

packages/form-js-playground/rollup.config.js
barmac: What would you expect to test? Something like https://github.com/bpmn-io/form-js/pull/403/commits/3c0647231ddad54f2a2ad53389ead50c976538e9  ?

packages/form-js-viewer/src/core/ConditionChecker.js
barmac: This sounds to me more like a form field concern: it knows whether it should display or not.

barmac:
Based on https://github.com/camunda/product-hub/issues/56#issuecomment-1315173662, I will need to reverse the logic so that we use a `"hideIf"` field.

Skaiir (commented):
Everything looks quite good, just got a few cleanup suggestions :).

Skaiir:
@barmac I'm not sure we should do the reversing in a separate issue. If I understand correctly you'd want to commit the current state once cleaned up to the `develop` and then in another commit invert the behavior right? 

The problem with that is that we should really avoid changes to the form schema, especially if we know they are going to be temporary, and **_especially_** if they break backwards compatibility.

I understand we don't want to be blocking this feature, but since this would break the backward compatibility of the schema, we'll still be blocked from creating new releases until the inversion has been committed to `develop`.

barmac:
Thanks for your feedback. It feels like we should have a conversation about this before the PR is merged. I'll set this up.

christian-konrad:
@barmac can't run it in the playground: 

https://user-images.githubusercontent.com/56470362/203605197-ae483dc5-05c6-41d3-948f-412d5f15cd3c.mov

barmac:
> @barmac can't run it in the playground:
> 
>  Bildschirmaufnahme.2022-11-23.um.17.56.42.mov

Sorry, the branch is wip now. I'll let you know when it's ready again.

barmac:
This is now ready for review. It implements the "hide if" logic and includes `getSchemaVariables` implementation for hide conditions.

packages/form-js-editor/src/features/properties-panel/entries/ConditionEntry.js
```diff
@@ -0,0 +1,69 @@
...
+    editField,
+    field,
+    id
+  } = props;
+
+  const debounce = useService('debounce');
+
+  const variablesList = useVariables();
+
+  // keep stable reference until https://github.com/bpmn-io/properties-panel/issues/196 is fixed
```
pinussilvestrus, line 40: That might be resolved now via https://github.com/bpmn-io/form-js/commit/3a51c379ea398a3ae916cec1a0aed8b0c6521cdd

pinussilvestrus (commented):
Looks good to me! Just one comment, otherwise, let's merge and bring this out of the door 👍

packages/form-js-editor/src/features/properties-panel/entries/ConditionEntry.js
barmac: Indeed! Let's check this in action.

barmac: OK done

pinussilvestrus (approved):
great work!🥇

PR #410, opened 2022-11-16
removed dependency on initialData

Skaiir:
The main change is that we no longer have an import step for the data. We do however still have to pass it through an initialization step to apply stuff like form field value defaults and input sanitation. 

But now instead of having to specify all the values that get passed, we just pass all the values by default, and then for form fields we overwrite with defaults when it makes sense.

### DIFF
--- a/packages/form-js-viewer/test/spec/import/Importer.spec.js
+++ b/packages/form-js-viewer/test/spec/import/Importer.spec.js
@@ -315,6 +315,32 @@ describe('Importer', function() {
     }));
 
 
+    it('should import data not defined on the form', inject(async function(form) {
+
+      // given
+      const data = { randomExternalDataBool: true, randomExternalDataInt: 9 };
+
+      // when
+      await form.importSchema(defaultValues, data);
+
+      // then
+      expect(form._getState().data).to.eql({
+        randomExternalDataBool: true,
+        randomExternalDataInt: 9 ,
+        creditor: 'Max Mustermann GmbH',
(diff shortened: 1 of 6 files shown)


packages/form-js-viewer/test/spec/import/Importer.spec.js
```diff
@@ -315,6 +315,32 @@ describe('Importer', function() {
     }));
 
 
+    it('should import data not defined on the form', inject(async function(form) {
```
nikku, line 318: Let's make this an integration test on the actual form instance (`Form.spec`).

packages/form-js-viewer/test/spec/render/components/FormField.spec.js
```diff
@@ -246,23 +246,6 @@ describe('FormField', function() {
     });
 
 
-    it('should use initial data to check condition', function() {
```
nikku, line 249: Should we not keep this around, i.e. as an integration test in `Form.spec`?

barmac: It's already covered in Form.spec: https://github.com/bpmn-io/form-js/pull/403/files#diff-f7f5270dd58d99d4b13f56770a7bcd44bee6b5d7a2f7d467fc4cf29b8f6d5093R1020-R1049

barmac:
I am merging this to my PR, so that the review of the feature can be consolidated.

PR #414, opened 2022-11-17
feat: Implement phone and email validations

vsgoulart:
Closes camunda/team-hto#52, camunda/team-hto#53


pinussilvestrus (commented):
There is a failing test case that prevents us from merging: https://github.com/bpmn-io/form-js/actions/runs/3486984421/jobs/5834087642#step:6:486

If this is fixed, we can merge it 👍

vsgoulart:
@pinussilvestrus I think this is ready for review again

packages/form-js-editor/src/features/properties-panel/groups/ValidationGroup.js
```diff
@@ -9,16 +9,33 @@ import {
...
+  phone: {
+    value: 'phone',
+    label: 'Phone',
+  },
+};
 
 export default function ValidationGroup(field, editField) {
   const { type } = field;
+  const validate = get(field, [ 'validate' ], {});
+  const isCustomValidation = [ undefined, VALIDATION_TYPE_OPTIONS.custom.value ].includes(validate?.validationType);
```
pinussilvestrus, line 38: Just to raise alternatives: instead of changing the schema by adding `validationType` we could detect the type by the pattern. But I guess this is too much magic, so having the `validationType` is the most explicit way 👍

PR #422, opened 2022-11-22
fix(taglist): make tag remove accessible via keyboard

### LINKED ISSUE #413: Pills in the taglist component are not accessiblie
__Describe the Bug__

When using that taglist component if a user selects a value the pills of the selected values in the taglist component are not accessible.

image

__Steps to Reproduce__

1. Create a form with a taglist and render it with the form viewer
2. Select a value in the taglist
3. Now with keyboard only try to remove a value
4. Selected values can't be removed with keyboard only (no outline and pills are not clickable/focusable)

__Expected Behavior__

Values should be able to be removed with keyboard only and the pills should be focusable with an outline

__Environment__

- OS: [e.g. Windows 7]

pinussilvestrus:
Closes #413

This makes the tag remove pills actual buttons. I also added another a11y test to verify the focus state (dropdown list open).

Kapture 2022-11-22 at 14 18 34

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Taglist.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Taglist.js
@@ -1,4 +1,5 @@
 import { useContext, useEffect, useMemo, useRef, useState } from 'preact/hooks';
+
 import useValuesAsync, { LOAD_STATES } from '../../hooks/useValuesAsync';
 
 import { FormContext } from '../../context';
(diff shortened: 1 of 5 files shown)


packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -106,6 +106,7 @@ export default function Taglist(props) {
       setIsEscapeClose(true);
       break;
     case 'Enter':
+      e.preventDefault();
```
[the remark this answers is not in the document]
vsgoulart: Event bubbling maybe? But the event handler seems to not affect that

vsgoulart (commented):
@pinussilvestrus I noticed that the focus path is a bit broken. When you have to focus on the list options and tab to the next element, if you tab back you have focus on the previous pill and not on the taglist input which open the available options.

Do you think we should fix that?

pinussilvestrus:
Seems like the focus got lost after the removal of a tab. I'll have a look 👍

pinussilvestrus:
I updated the PR, so the focus path should be clean now. Let me know if you see anything else.

Kapture 2022-11-24 at 11 42 38

Note that I keep the dropdown list unfocussable. That's the state beforehand to ensure it works correctly with up-down navigation (similar to the native select dropdown).

vsgoulart:
https://user-images.githubusercontent.com/1022916/203797299-ad056c50-21bf-4595-bb5e-9a19ff956fbe.mp4

@pinussilvestrus Found another issue. If you select one of the radio buttons on the field before and press enter the last pill is deleted

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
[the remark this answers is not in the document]
pinussilvestrus: Root cause: that's related to a missing [`type` attribute](https://www.w3schools.com/tags/att_button_type.asp) (default is `submit`).

pinussilvestrus:
I forgot to add the `type` attribute for the remove button. Should work now 👍

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -113,22 +114,44 @@ export default function Taglist(props) {
...
-          return (
-            <div class="fjs-taglist-tag" onMouseDown={ (e) => e.preventDefault() }>
-              <span class="fjs-taglist-tag-label">
-                {valueToOptionMap[v] ? valueToOptionMap[v].label : `unexpected value{${v}}`}
-              </span>
-              <span class="fjs-taglist-tag-remove" onMouseDown={ () => deselectValue(v) }><CloseIcon /></span>
-            </div>
-          );
-        })
+        <div class="fjs-taglist-tags">
```
vsgoulart, line 135: I'll have to check if this doesn't break the styling on Tasklist. We have to setup a system for changes on components in order to not break customization

vsgoulart (approved):

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
pinussilvestrus: I agree with being careful with this kind of change. I introduced this just to ensure the tags are encapsulated from the search input. I can also revert this if you see any problem in Tasklist 👍

vsgoulart: I just rebased locally with `develop` and tested it on Tasklist. This PR should be good to merge.
I just need to fix the focus state, but it's unrelated to this PR

pinussilvestrus: thanks for the confirmation 👍

PR #423, opened 2022-11-22
feat: retrieve variables mentioned in conditions

### LINKED ISSUE #401: `getSchemaVariables` should include variables mentioned in conditions

barmac:
This implements support for `field#condition` in `getSchemaVariables`. Note that it doesn't allow to shadow keywords, i.e. `=external` won't work. Until we know the context, we can't tell that a keyword is a variable, and `getSchemaVariables` is used to **create** the context.

Closes #401

### DIFF
--- a/packages/form-js-viewer/src/util/feel.js
+++ b/packages/form-js-viewer/src/util/feel.js
@@ -1,0 +1,24 @@
+import { parseUnaryTests } from 'feelin';
+
+/**
+ * Retrieve variable names from given FEEL unary test.
+ *
+ * @param {string} unaryTest
+ * @returns {string[]}
+ */
+export function getVariableNames(unaryTest) {
+  const tree = parseUnaryTests(unaryTest);
+  const cursor = tree.cursor();
+
+  const variables = new Set();
+  do {
+    const node = cursor.node;
+
+    if (node.type.name === 'VariableName') {
+      variables.add(unaryTest.slice(node.from, node.to));
+    }
+
+  } while (cursor.next());
+
+  return Array.from(variables);
+}
--- a/packages/form-js-viewer/src/util/index.js
+++ b/packages/form-js-viewer/src/util/index.js
@@ -1,3 +1,5 @@
+import { getVariableNames } from './feel';
+
 export * from './injector';
 export * from './form';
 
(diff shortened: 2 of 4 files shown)


packages/form-js-viewer/src/util/feel.js
```diff
@@ -0,0 +1,24 @@
...
+
+  const variables = new Set();
+  do {
+    const node = cursor.node;
+
+    if (node.type.name === 'VariableName') {
+      variables.add(unaryTest.slice(node.from, node.to));
+    }
+
+  } while (cursor.next());
```
pinussilvestrus, line 21: Couldn't we also do `tree.iterate` here? Maybe it's safer? I don't know much about the insights.

```js
tree.iterate({
  enter: (node) => {
    if (node.name === 'VariableName') {
      variables.add(unaryTest.slice(node.from, node.to));
    }
  }  
});
```

packages/form-js-viewer/src/util/index.js
```diff
@@ -92,11 +94,21 @@ export function getSchemaVariables(schema) {
       variables = [ ...variables, key ];
     }
 
-    if (valuesKey && !variables.includes(valuesKey)) {
+    if (valuesKey) {
       variables = [ ...variables, valuesKey ];
     }
 
-    return variables;
+    if (condition) {
```
pinussilvestrus, line 101: In the future, we would need to specify any other property here that could be an expression or unary test, right?

barmac: Yeah or have a separate `if` with logic to handle that property.

packages/form-js-viewer/src/util/feel.js
barmac: I tried both and liked cursor-based solution more. In `tree.iterate`, you have to know that iteration is done synchronously, while with cursor it's explicit.

packages/form-js-viewer/src/util/index.js
```diff
@@ -69,19 +71,19 @@ export function clone(data, replacer) {
...
 
-  return schema.components.reduce((variables, component) => {
+  const variables = schema.components.reduce((variables, component) => {
 
     const {
       key,
       valuesKey,
-      type
+      type,
+      condition
```
pinussilvestrus, line 86: todo: rename to `conditional.hide`.

PR #429, opened 2022-11-24
chore: add @bpmn-io/properties-panel to @bpmn-io/form-js-editor

vsgoulart:
With the alpha version we started to get the following warning on Tasklist:
```
./node_modules/@bpmn-io/form-js-editor/dist/index.es.js
Module not found: Can't resolve '@bpmn-io/feel-editor' in '/home/vinicius/repos/tasklist/client/node_modules/@bpmn-io/form-js-editor/dist'
```

This is happening because `@bpmn-io/properties-panel` is not bundling `@bpmn-io/feel-editor` anymore. To fix that we're just adding `@bpmn-io/properties-panel` as a dependency

### DIFF
--- a/packages/form-js-editor/package.json
+++ b/packages/form-js-editor/package.json
@@ -43,6 +43,7 @@
   },
   "dependencies": {
     "@bpmn-io/form-js-viewer": "^0.10.0-alpha.0",
+    "@bpmn-io/properties-panel": "^0.25.0",
     "array-move": "^3.0.1",
     "dragula": "^3.7.3",
     "ids": "^1.0.0",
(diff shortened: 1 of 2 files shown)


packages/form-js-editor/package.json
```diff
@@ -43,6 +43,7 @@
   },
   "dependencies": {
     "@bpmn-io/form-js-viewer": "^0.10.0-alpha.0",
+    "@bpmn-io/properties-panel": "0.25.0",
```
pinussilvestrus, line 46: Do we need to pin it? We don't do that elsewhere if there is no need.

vsgoulart: We don't have to, it's a habit I have to guarantee more stable installations
But I can change it to keep consistent

pinussilvestrus: Let's do so until we see the need to pin it 👍

pinussilvestrus (approved):
Let's try this out

PR #432, opened 2022-11-25
feat: add `key` validation to `input values key`

### LINKED ISSUE #428: Inconsistent validation for `key` and `valuesKey`
__Describe the Bug__

In the properties panel, we do not allow spaces in form field keys.

image

We don't do the same for the "Input values key".

image

__Steps to Reproduce__

1. Open forms editor
2. Select a multi-select element (e.g. `taglist`)
3. Enter variable name with spaces in `Key` => spaces are not allowed
4. Enter variable name with spaces in `Input values key` => spaces are allowed

__Expected Behavior__

Be consistent when declaring variable keys in form fields.

__Environment__

- Host (Browser/Node version), if applicable: Chrome 107.0.5304.110
- OS: macOS 11.7
- Library version: `develop`

Skaiir:
Closes #428

@RomanKostka 

firefox_zn2fH4YaeX

Is this okay behavior? This property is required to be set always, and while I understand that preemptively erroring things in a form might be a UX problem, I think the clarity is more important to a form designer. This is a pattern we've applied for the bpmn editor as well.

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/InputKeyValuesSourceEntry.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/InputKeyValuesSourceEntry.js
@@ -1,5 +1,5 @@
 import { TextFieldEntry, isTextFieldEntryEdited } from '@bpmn-io/properties-panel';
-import { get } from 'min-dash';
+import { get, isUndefined } from 'min-dash';
 import { useService } from '../hooks';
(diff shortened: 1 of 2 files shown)


RomanKostka (approved):
Thanks!
for me it is fine from UX side.
Especially when having the same behavior in the bpmn editor.

pinussilvestrus (approved):

Issue #437, opened 2022-11-29
Remove form fields via keyboard


pinussilvestrus:
__Is your feature request related to a problem? Please describe__

Removing elements requires me to click the delete action via the context pad. As a heavy keyboard, I often wish to do that via keyboard, as you know from many other editors (e.g., bpmn-js).

__Describe the solution you'd like__

Make it possible to remove form fields in the editor via `DELETE` or `BACKSPACE` key.

__Describe alternatives you've considered__

Do not do anything.

__Additional context__

Inspiration from how it is done in [diagram-js editors](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/keyboard/KeyboardBindings.js#L164).

Issue #173, opened 2021-09-14
Improve editor element selection and keyboard navigation


nikku:
__Is your feature request related to a problem? Please describe__

The form editor at this stage is not keyboard aware: While there is a visual selection that selection does not confirm to the actual (browser) selection. Tabbing with the keyboard focuses embedded links rather than the actual form fields.

__Describe the solution you'd like__

Improve accessibility by better integrating with native browser focus handling.

* [ ] Selection on the canvas confirms with the keyboard selection
* [ ] Navigation between form elements is possible via tabbing

__Describe alternatives you've considered__

Ignore this topic.

__Additional context__

This screencapture shows the current state of keyboard navigation:

![capture bOICWW_optimized](https://user-images.githubusercontent.com/58601/133254651-c72be077-cbd5-47cd-b463-3d32bcc39b0e.gif)

2manoj1:
Are form-js support keyboard - On Select remove form component by pressing delete key?

pinussilvestrus:
We improved a lot via https://github.com/bpmn-io/form-js/pull/634 (navigation and selection via keyboard). Let's close this one and wait for feedback.

PR #438, opened 2022-11-29
Remove form field via keyboard

pinussilvestrus:
Related to #437 

Check it out via https://demo-437-remove-via-keyboard--camunda-form-playground.netlify.app/.

What immediately comes through: to make it fly, we'd need navigation via keyboard (cf. #173). 

Kapture 2022-11-29 at 16 32 07


pinussilvestrus:
I'd love to hear your feedback @Skaiir @vsgoulart @RomanKostka @christian-konrad

pinussilvestrus:
We should probably simply limit the keyboard binding of the editor to the editor container (via keyboard config). It's currently binded to the document.

vsgoulart:
@pinussilvestrus I like it a lot! I would say this is a very power user feature, so I would give it a lower priority if we compare to other features. But if the implementation is very easy I think it's worth it to merge.

Do we have any shortcuts cheatsheet in the playground?

pinussilvestrus:
> Do we have any shortcuts cheatsheet in the playground?

I don't think so. [We have a module](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/src/features/keyboard/FormEditorKeyboardBindings.js) that takes care of these. The playground itself doesn't have any other keyboard bindings (yet). The panel toggles are implemented in the Modeler products separately.

pinussilvestrus:
Feedback [via Slack](https://camunda.slack.com/archives/C046XS3GGCR/p1669737663905269?thread_ts=1669736130.558029&cid=C046XS3GGCR) from @Skaiir 

> If there is one thing I would say it's that it's maybe a little bit too easy for the deletion to get captured by the main window. Basically the editor is always active, and unless we're focusing something that will handle the keyboard input, the delete action will trigger. A good example of this problem can be seen when pressing delete when focusing a taglist tag.

nikku:
From my point of view the overall interaction pattern (including playgrounds and stuff) is blocked by https://github.com/bpmn-io/diagram-js/issues/661.

To properly handle keyboard shortcuts we must be browser focus aware and move from explicit keyboard bindings to implicit bindings.

Cf. https://github.com/bpmn-io/diagram-js/pull/662#issuecomment-1218494067 how that works (implicit keyboard focus) in the playground.

Generally I 100% support keyboard navigation. The core at least should invite both clickclick and power users.

PR #444, opened 2022-12-01
chore: adjusted test img svg to render in firefox

Skaiir:
Height-only svgs don't render in firefox, made it width-only instead.


pinussilvestrus (approved):
🐎

PR #446, opened 2022-12-01
Added h6 to the avilable html tags - issue #412

### LINKED ISSUE #412: H6 header disappears in markdown text view
__Describe the Bug__

We don't have the equivalent of an h6 with `######`—the header just disappears.

__Steps to Reproduce__

Create a text view with a h6 header—and it will disappear:

Bildschirmfoto 2022-11-16 um 16 32 35

__Expected Behavior__

Show that header

__Environment__

- Latest playground as of Nov 16

adrianmarinwork:
Closes #412

### DIFF
--- a/packages/form-js-viewer/src/render/components/Sanitizer.js
+++ b/packages/form-js-viewer/src/render/components/Sanitizer.js
@@ -7,6 +7,7 @@ const ALLOWED_NODES = [
   'h3',
   'h4',
   'h5',
+  'h6',
   'span',
   'em',
   'a',


packages/form-js-viewer/src/render/components/Sanitizer.js
```diff
@@ -7,6 +7,7 @@ const ALLOWED_NODES = [
   'h3',
   'h4',
   'h5',
+  'h6',
```
pinussilvestrus, line 10: Maybe let's update our [existing test case](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-viewer/test/spec/render/components/form-fields/Text.spec.js#L41) to also test for the new `h6` element.

pinussilvestrus (commented):
Looks really promising, thanks for your contribution ❤️

I added one idea regarding the test coverage. Let's also update the commit message to fit our guidelines

```text
chore(viewer): add h6 to the available HTML tags 

Closes #412
```

PR #450, opened 2022-12-02
chore(viewer): add h6 to the available HTML tags

### LINKED ISSUE #412: H6 header disappears in markdown text view

adrianmarinwork:
Closes #412

### DIFF
--- a/packages/form-js-viewer/test/spec/render/components/form-fields/Text.spec.js
+++ b/packages/form-js-viewer/test/spec/render/components/form-fields/Text.spec.js
@@ -80,7 +80,7 @@ Some _em_ **strong** [text](#text) \`code\`.
     const formField = container.querySelector('.fjs-form-field');
 
     expect(formField).to.exist;
-    expect(formField.innerHTML).to.eql(`<div class="markup"><div xmlns="http://www.w3.org/1999/xhtml"><h1>h1</h1><h2>h2</h2><h3>h3</h3><h4>h4</h4>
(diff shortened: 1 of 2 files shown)


pinussilvestrus (approved):
Looks good to me! Thank you very much 🎉

packages/form-js-viewer/test/spec/render/components/form-fields/Text.spec.js
```diff
@@ -46,6 +46,8 @@ describe('Text', function() {
 ## h2
 ### h3
 #### h4
+##### h5
+###### h6
```
pinussilvestrus, line 50: We need to adjust the `expect` part of this test as well: https://github.com/bpmn-io/form-js/pull/450/files#diff-f782522d93784395284f1b8e2c8b29c8ff07908fcb77815c9f527d9aa8222e0dR83

You can execute the tests locally as well via `npm run test` or `npm run dev -- '@bpmn-io/form-js-viewer`

pinussilvestrus (changes requested):
I was a bit too fast and didn't wait for the CI 😅 Let's fix the test and merge afterward!

adrianmarinwork:
Okay, I've updated the expected value of the test with the new HTML tags!

pinussilvestrus:
Works now, thanks! 👍

PR #452, opened 2022-12-05
fix(viewer): hidden fields do not affect other fields

### LINKED ISSUE #431: When a component is hidden, it's data should not affect other components
__What should we do?__

When a component is hidden (i.e. as the result to an expression), it's current data state (potentially determined by previous user input, another expression, or both) should not affect other components as long as it is hidden.

__Why should we do it?__

Components that are hidden do not impact the form output. Either the value is not submitted, or the original one (passed via input data). But currently, it affects the conditions and expressions of other fields. According to "no magic", we don't want fields hidden to the user to impact the behavior of the form. This can lead to (in the perspective of the user) non-deterministic, non-idempotent, unclear behavior (since the users misses a part of the information that controls the behavior).

Hidden form field content therefore should **not** impact other fields (via expressions/conditions).
To use such fields for intermediate, but hidden calculations (i.e. using formula fields), is also discouraged. We should rather introduce variables that are not bound to form fields (cf. https://github.com/camunda/product-hub/issues/697).

See this video to understand the issue:

https://user-images.githubusercontent.com/56470362/203988251-22ac708f-01cb-49f6-b4ed-77dc1628efbe.mov

pinussilvestrus:
Closes #431

* makes sure data from conditionally hidden fields do not affect other conditions/expressions
* instead, uses `initialData` as the fallback

[Checkout the demo](https://demo-431-hidden-fields-data--camunda-form-playground.netlify.app/).

### DIFF
--- a/packages/form-js-viewer/src/render/hooks/useCondition.js
+++ b/packages/form-js-viewer/src/render/hooks/useCondition.js
@@ -9,11 +9,18 @@ import useService from './useService.js';
  * @returns {boolean} true if condition is met or no condition or condition checker exists
  */
 export function useCondition(condition, data) {
+  const initialData = useService('form')._getState().initialData;
   const conditionChecker = useService('conditionChecker', false);
 
   if (!conditionChecker) {
     return null;
   }
 
-  return conditionChecker.check(condition, data);
+  // make sure we do not use data from hidden fields
+  const filteredData = {
+    ...initialData,
+    ...conditionChecker.applyConditions(data, data)
+  };
+
+  return conditionChecker.check(condition, filteredData);
 }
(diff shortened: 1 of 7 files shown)


pinussilvestrus:
@christian-konrad @RomanKostka @vsgoulart @Skaiir I added [a demo](https://demo-431-hidden-fields-data--camunda-form-playground.netlify.app/) with the branch. My goal would be to test this one like crazy to check all edge cases are covered. Feel free to have a look, break it and report back 👍

pinussilvestrus:
@bpmn-io/modeling-dev fyi. Feel free to try as well 👍

pinussilvestrus:
Sending this under review now, as I wasn't able to break it.

christian-konrad (approved):
No way to crash it :) lgtm

pinussilvestrus:
Just merging this now to get it out of the door. The demo is still available, so feel free to report afterward 👍

PR #457, opened 2022-12-09
Playground Input Data Autocomplete

### LINKED ISSUE #295: Playground: Auto-populate used variables in (initial) Input Data

Given we can extract the defined variables from the schema (cf. #294), we could potentially auto-populate the keys in the input data container once the form is initialized. We could also suggest the variable keys in the code editor.

image

[`getSchemaVariables`](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-viewer/src/util/index.js#L81) can be a helper here.

Once a a form is initialized, automatically add used variables to the input data, so I don't have to type it manually. Also, consider show all variables as suggestions to also cover added variables after the first initialization. 

Populate each variable after it got added via a form field or expression. This could lead to some unwanted magic though. 

__Additional context__

Related to https://github.com/bpmn-io/form-js/issues/103

pinussilvestrus:
* Incorporates proper tab key handling inside JSON editor
* Adds autocomplete of variables keys (fetched via `getSchemaVariables`)

Closes #295 

Find a demo here: https://demo-295-populate-input-data--camunda-form-playground.netlify.app/

image


christian-konrad:
See comment here: https://camunda.slack.com/archives/C04F4AR9GAD/p1670839419313839

pinussilvestrus:
I added 81580643944ab8245c28f9d8dce7f7286c0eb4d0 to make sure it also triggers immediately when the cursor is inside a property name.

pinussilvestrus:
This is ready for review now.

pinussilvestrus:
We use the built-in [JSON linter](https://github.com/codemirror/lang-json) for this, meaning we would need to build our own. It's doable, but:

The [built-in one](https://github.com/codemirror/lang-json/blob/main/src/lint.ts) uses `JSON.parse`, which indicates the error on the following line. This is standard behavior, as Monaco is doing the same (the editor we use in the Connectors Editor).

image

RomanKostka:
@pinussilvestrus that the error is displayed in the following line is okay.
I was wondering about the content of the error message.

current msg:
image

... in JSON at **position 74.**

that leads to confusion because the position 74 is not visible for anyone.
The user only sees lines:
image

Therefore the error message would be easier understood by the user if the rephrase it like following and tell the user the line the error occurs:

'Expected ... in line 4'

christian-konrad:
@RomanKostka @pinussilvestrus I found this:

https://codemirror.net/5/demo/lint.html

Bildschirmfoto 2022-12-14 um 10 36 28

Might be helpful. I would consider this a separate issue (i.e. spring cleaning?)

christian-konrad (approved):

pinussilvestrus:
> @RomanKostka @pinussilvestrus I found this:
> 
> https://codemirror.net/5/demo/lint.html
> 
> Bildschirmfoto 2022-12-14 um 10 36 28
> 
> Might be helpful. I would consider this a separate issue (i.e. spring cleaning?)

The demo is about Codemirror 5 I think, so it might not be compatible with the version we use (v6). But for sure we can investigating overriding the default linting. Feel free to open another issue for this :+1:

packages/form-js-playground/test/spec/JSONEditor.spec.js
```diff
@@ -0,0 +1,187 @@
...
+}
+
+function getCm(editor) {
+  return editor._view || editor;
+}
+
+/**
+ * Copied over from @bpmn-io/feel-editor.
+ */
+async function expectEventually(fn) {
```
vsgoulart, line 170: We could consider using React Testing Library, they have utilities like this too

vsgoulart (approved):
One thing I noticed is that when you press tab with the autocompletion on it doesn't select the highlighted option like on most editors.

Is that easy to implement? I think it's more natural

pinussilvestrus:
> One thing I noticed is that when you press tab with the autocompletion on it doesn't select the highlighted option like on most editors.
> 
> Is that easy to implement? I think it's more natural

I like this idea, but implementing could be quite hard as we'd need to override the keymaps of the CodeMirror editor. I also think we should not differ too much from the other code editor look & feel in our toolkits (compared to the FEEL editor, where tab also navigates outside). If we introduce certain behaviors, we should make sure, we do it everywhere. /cc @christian-konrad

packages/form-js-playground/test/spec/JSONEditor.spec.js
pinussilvestrus: Good catch. We use Preact Testing Library which seems not to offer something like this. I think we should not mix up just because of some utilities. Also does not feel pretty natural to use React Testing utilities in a Preact library.

vsgoulart: Yeah, what I meant by React Testing Library was using the Preact counterpart.
It seems like the async hooks are actually part of the the `dom-testing-library` package, so it should be ok to use in Preact project https://testing-library.com/docs/dom-testing-library/api-async and they have examples in the Preact docs https://preactjs.com/guide/v10/preact-testing-library

pinussilvestrus: Thanks for the follow-up 👍 I'll see whether I can contribute a slight adjustment here on top.

PR #463, opened 2022-12-14
fix: align editor and viewer spacing

### LINKED ISSUE #424: Align all selection margins
__Describe the Bug__

To avoid more conflicts in the CSS than we are already going to have, I've decided to take on the alignment of selection margins as part of a separate ticket here.

The issue in question is best described visually:

image
image
image
image

There is inconsistency in the paddings applied to various fields. We would like to homogenize this a little bit. 

__Steps to Reproduce__

1. Select various components of a form in the editor
2. Notice the variation in padding, especially bottom padding

__Expected Behavior__

All padding top and bottom should be equal in all scenarios. 

__Related__

https://github.com/bpmn-io/form-js/issues/406

Skaiir:
Closes #424

### DIFF
--- a/packages/form-js-viewer/assets/form-js.css
+++ b/packages/form-js-viewer/assets/form-js.css
@@ -228,19 +228,9 @@
 .fjs-container .fjs-input-group {
   display: flex;
   width: 100%;
-  margin: 4px, 0;
+  margin: 4px 0;
   border: var(--border-definition);
   border-radius: 3px;
-}
-
-.fjs-container .fjs-input-group.disabled {
-  background-color: var(--color-background-disabled);
-}
-
-.fjs-container .fjs-input-group.disabled,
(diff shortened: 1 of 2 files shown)


Skaiir:
It's not perfect but it's a lot better now :)

image

image

I also made all selection margins 4px, which was what the original ticket was about. Everything looks much better.

image
image

packages/form-js-viewer/assets/form-js.css
```diff
@@ -567,7 +558,7 @@
 
 .fjs-container .fjs-image-placeholder {
   height: 64px;
-  margin: 2px 0;
+  margin: 4px 0;
```
vsgoulart, line 561: We could save the default spacing in a CSS var

vsgoulart (approved):

PR #473, opened 2022-12-27
chore(playground): adjust data panel names

nikku:
Rename playground data panels to `Form Input` and `Form Output`:

image


Skaiir (approved):
As per the new playground, makes sense to me.

PR #475, opened 2022-12-27
feat(playground): render properties panel to the right

nikku:
Ensures a uniform UX across our editing tools (always rendered to the right).

image

### DIFF
--- a/packages/form-js-playground/src/components/PlaygroundRoot.css
+++ b/packages/form-js-playground/src/components/PlaygroundRoot.css
@@ -35,10 +35,6 @@
   grid-template-rows: 70% 30%;
 }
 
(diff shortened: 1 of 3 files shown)


Skaiir:
I don't know how much we care about keeping the old playground up to date, but this makes sense considering it's matching the new playground style. Though in this case we can't collapse the window so it's a bit more mouse travel to go from the editor to the properties panel, but not a huge deal.

Skaiir (approved):

nikku:
> I don't know how much we care about keeping the old playground up to date, but this makes sense considering it's matching the new playground style. 

@Skaiir We do care having a playground (end-to-end usable thing) available in core modeling (bpmn.io). Without that the form toolkit is much less accessible.

PR #476, opened 2022-12-27
feat: simplify checklist icon

nikku:
Makes checklist icon more compact so it better fits into the existing context (palette + properties panel).

capture zqbE1j_optimized


Skaiir:
I think it's a strict improvement but I'll wait on Roman for this one.

RomanKostka (changes requested):
nice!
one thing i would like to add:
less stroke width for the checklist labels.
I created a new svg for that:
Checklist

nikku:
Updated checklist icon. Works for me as well!

capture dg2P0Q_optimized

RomanKostka (approved):
thanks!
looks good from my point of view.

nikku:
@Skaiir Awaiting your final review.

Skaiir (approved):

PR #477, opened 2022-12-29
chore(lerna): do not use nx

pinussilvestrus:
By [bumping to `lerna@6`](https://github.com/bpmn-io/form-js/pull/471), we now enable the new `nx` task scheduler per default (cf. https://github.com/lerna/lerna/blob/main/CHANGELOG.md#600-2022-10-12).

Although it seems not to have a negative impact, I'd like to make this an actual decision and take a bit of time to understand the consequences. This one disables `nx` until we make this decision.

Also, pinging @bpmn-io/modeling-dev to make them aware. I think we should roll this out across a couple of more bpmn-io projects in case we decide to do so.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,6 @@
 {
   "version": "0.10.1",
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
-  "useWorkspaces": true
+  "useWorkspaces": true,
+  "useNx": false
 }


pinussilvestrus:
From what I understood, the main advantage of using `nx` with Lerna is the [caching support](https://nx.dev/concepts/how-caching-works) of quite everything task related. Depending on how well the support is made for GitHub actions, it might also speed up CI

Docs: https://nx.dev/recipes/adopting-nx/lerna-and-nx

nikku:
I'd say go ahead and see what happens.

No changes API wise, so the upgrade _should_ be smooth.

vsgoulart:
I think it's better to leave the defaults and only change in case we see problems. As Nico said, nothing changed API wise

pinussilvestrus:
👍 then let's go for it. I'll close accordingly.

PR #488, opened 2023-01-09
fix(playground): props panel no longer squashes

### LINKED ISSUE #487: Dev playground layout stretches strangely under certain circumstances.
__Describe the Bug__

The new layout of the dev internal playground stretches beyond the monitor size in some cases. Most likely related to [this commit](https://github.com/bpmn-io/form-js/commit/8289216fb5cc5ccc24241a84f1a57ca35098a622).

image

__Steps to Reproduce__

1. Add a text or number field
2. Include a very large prefix or suffix
3. The editor and preview screens slowly start taking over the properties panel, until they eventually completely squash it out of the view

__Expected Behavior__

The properties panel should neither be squashed nor pushed out of view.

Skaiir:
Closes #487

image

### DIFF
--- a/packages/form-js-playground/src/components/PlaygroundRoot.css
+++ b/packages/form-js-playground/src/components/PlaygroundRoot.css
@@ -29,7 +29,7 @@
 }
 
 .fjs-pgl-main {
-  width: calc(100% - 111px);
+  width: calc(100% - 361px);
   height: 100%;
 
   display: grid;
@@ -79,9 +79,13 @@
   position: relative;
   display: flex;
   height: 100%;
+  width: 250px;
   overflow-y: auto;
   background-color: var(--color-properties-container-background);
-  --properties-panel-width: 200px;
+}
+
+.fjs-pgl-properties-container .fjs-properties-container {
+  --properties-panel-width: 250px;
 }
 
 .fjs-pgl-properties-container .bio-properties-panel {


packages/form-js-playground/src/components/PlaygroundRoot.css
```diff
@@ -79,9 +79,13 @@
...
   display: flex;
   height: 100%;
+  width: 250px;
   overflow-y: auto;
   background-color: var(--color-properties-container-background);
-  --properties-panel-width: 200px;
+}
+
+.fjs-pgl-properties-container .fjs-properties-container {
+  --properties-panel-width: 250px;;
```
vsgoulart, line 88:
```suggestion
  --properties-panel-width: 250px;
```

vsgoulart, line 88: Do we really want to limit the scope? Why can't we define it on `:root`?

vsgoulart (changes requested):

packages/form-js-playground/src/components/PlaygroundRoot.css
pinussilvestrus: We are not defining it, we are overriding the one [specified in form-js-editor](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/assets/form-js-editor.css#L30).

Being specific helps us not to interfere with potential variables from other properties panel libraries. One alternative could be to adjust our CSS variables to be more specific in their names.

Skaiir: The reason I've limited the scope is simply that this just did not work previously. The property is used (and defined) for .fjs-properties-container, so it needs to be overridden on a selector of more specificity. Honestly, this kind of CSS variable definition is something that feels very unpredictable and hard to trace. I think in general we probably shouldn't define variables anywhere but at the root, that way the behavior is a bit more predictable.

vsgoulart (approved):
The adornment text is still overflowing, but as we talked over Slack @Skaiir mentioned he will fix it in another PR

PR #490, opened 2023-01-10
fix: trim increment number value

Skaiir:
Related to https://github.com/camunda/camunda-modeler/issues/3382

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/NumberEntries.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/NumberEntries.js
@@ -91,7 +91,13 @@ function NumberArrowStep(props) {
     return value;
   };
 
-  const setValue = (value) => editField(field, [ 'increment' ], value);
+  const clearLeadingZeroes = (value) => {
+    if (!value) return value;
+    const trimmed = value.replace(/^0+/g, '');
+    return (trimmed.startsWith('.') ? '0' : '') + trimmed;
+  };
+
+  const setValue = (value) => editField(field, [ 'increment' ], clearLeadingZeroes(value));
 
   const decimalDigitsSet = decimalDigits || decimalDigits === 0;
 
--- a/packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
+++ b/packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
@@ -2360,6 +2360,106 @@ describe('properties panel', function() {
           });
 
 
+          it('should clear', function() {
+
+            // given
+            const editFieldSpy = spy();
+
+            const field = defaultValues.components.find(({ key }) => key === 'amount');
+
+            createPropertiesPanel({
+              container,
+              editField: editFieldSpy,
+              field
+            });
+
+            // assume
+            const input = screen.getByLabelText('Increment');
+
+            // when
+            fireEvent.input(input, { target: { value: '' } });
+
+            // then
+            expect(editFieldSpy).to.have.been.calledWith(field, [ 'increment' ], undefined);
+
+          });
+
+
+          it('should trim leading zeroes', function() {
+
+            // given
+            const editFieldSpy = spy();
+
+            const field = defaultValues.components.find(({ key }) => key === 'amount');
+
+            createPropertiesPanel({
+              container,
+              editField: editFieldSpy,
+              field
+            });
+
+            // assume
+            const input = screen.getByLabelText('Increment');
+
+            // when
+            fireEvent.input(input, { target: { value: '0005.1000' } });
+
+            // then
+            expect(editFieldSpy).to.have.been.calledWith(field, [ 'increment' ], '5.1000');
+
+          });
+
+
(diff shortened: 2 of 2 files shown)


packages/form-js-editor/src/features/properties-panel/entries/NumberEntries.js
```diff
@@ -91,7 +91,7 @@ function NumberArrowStep(props) {
     return value;
   };
 
-  const setValue = (value) => editField(field, [ 'increment' ], value);
+  const setValue = (value) => editField(field, [ 'increment' ], value && Big(value).toString());
```
pinussilvestrus, line 94: It seems like this is not side-effect free. When typing rather low and starting to type the digits, the `.` got erased rather quickly. 

Kapture 2023-01-10 at 14 02 52

Skaiir: Very nice catch, there are three ways we can fix this if we want to keep trimming

1) We don't treat `45.` as a valid value, and wait for the user to enter a more complete entry:

4 is entered => 4 is serialized
5 is entered => 45 is serialized
. is entered => nothing happens
5 is entered => 45.5 is serialized

The downside of this is, if someone were to enter `45.` quickly, because the field is debounced, we may have no value at all.

2) We serialize the dot in the schema

This avoids the above mentioned downside, but may be a little weird in terms of schema to store something like `23.`

3) We implement something at the properties panel field level to deal with this exact scenario

Downside here is implementation time, I'm not sure how/if such an implementation would work well.

@christian-konrad @pinussilvestrus maybe you have some input here? Personally I think option 2 is the right call.

pinussilvestrus: > We serialize the dot in the schema

As far as I can see`45.`  is _valid_ input data for the numbers itself, as it would simply resolve to `45`. You can also specify this via input data; no error is thrown.

Therefore we would be fine persisting it this way in the schema for the increment, as it would be consistent. If we can avoid trimming for such cases, even better - no magic.

vsgoulart (commented):
For the decimal inputs config still accepts non-integer numbers, which I guess it doesn't make sense and it's the point of this PR, no?

Another thing that I think it's unrelated to this PR is that the range validation seems to accept non-integer values but then the validation looks like this
image

In the image above I used 5.3 as a maximum value

Skaiir:
Let's keep the trimming very simple, clearing leading zeroes for decimal digits. 

As far as https://github.com/camunda/camunda-modeler/issues/3382's second part is concerned, let's not tackle this until we have the ability to define custom validation for the properties panel number fields. For now UX will not be ideal but to be honest this has always been the case with number properties.

packages/form-js-editor/src/features/properties-panel/entries/NumberEntries.js
```diff
@@ -91,7 +91,12 @@ function NumberArrowStep(props) {
     return value;
   };
 
-  const setValue = (value) => editField(field, [ 'increment' ], value);
+  const clearLeadingZeroes = (value) => {
+    const trimmed = value.replace(/^0+/g, '');
+    return trimmed.startsWith('.') ? '0' + trimmed : trimmed;
```
vsgoulart, line 96:
```suggestion
    return value.replace(/^0+/g, '0');
```

Skaiir: I think this wouldn't work, because then 00005.25 => 05.25

vsgoulart: Yeah, you're right
We could update the regexp, but I think your solution is easier to read

Skaiir: Actually there's a flaw in this one 😢, gonna push a new version just now. Basically value can be null I haven't accounted or tested for that. Let me just push the commit through.

packages/form-js-editor/src/features/properties-panel/entries/NumberEntries.js
```diff
@@ -91,7 +91,13 @@ function NumberArrowStep(props) {
     return value;
   };
 
-  const setValue = (value) => editField(field, [ 'increment' ], value);
+  const clearLeadingZeroes = (value) => {
+    if (!value) return value;
+    const trimmed = value.replace(/^0+/g, '');
+    return (trimmed.startsWith('.') ? '0' : '') + trimmed;
```
vsgoulart, line 97:
```suggestion
    return `${trimmed.startsWith('.') ? '0' : ''}${trimmed}`;
```

vsgoulart (approved):

packages/form-js-editor/src/features/properties-panel/entries/NumberEntries.js
Skaiir: I think this is personal preference, but personally, I wouldn't use computed strings for concatenation. So I'll merge as-is. Thanks for the review :)

Issue #362, opened 2022-10-17
Discourage the use of disabled fields


pinussilvestrus:
__What should we do?__

The usage of disabled fields should not be encouraged (context, e.g. https://medium.com/@izzmo/anti-pattern-using-disabled-or-read-only-for-form-input-e2c24c669f5b). For displaying static values, rather [use read-only text to display that data](https://github.com/bpmn-io/form-js/issues/247).

Consider linting as one way to discourage the use.

__Why should we do it?__

The original use case of disabled form fields was to display static values in the form that should not be submitted. However, [with emerging use cases that are possible](https://github.com/bpmn-io/form-js/issues/273) with form-js, this workaround is not needed, and should not be needed anymore.

Disabled form fields are not recommended in HTML forms land, nor should they be recommended by our toolkits. We can not ditch them entirely without having a major breaking change in our forms schema.

Also important: [Our state about submitting](https://bpmn.io/blog/posts/2022-form-js-default-values-improved-submit.html) (blog post).

Skaiir:
One thing to note with the linter is that it's reactive to users making a mistake (or upgrading an old version of a form). We should probably also discourage this some other way.

Skaiir:
Closed via https://github.com/camunda/form-linting/pull/68

PR #495, opened 2023-01-12
fix: display datetime time value when disabled

Skaiir:
Related to https://github.com/camunda/camunda-modeler/issues/3383

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/parts/Datepicker.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/parts/Datepicker.js
@@ -50,11 +50,6 @@ export default function Datepicker(props) {
   // setup flatpickr instance
   useEffect(() => {
 
-    if (disabled) {
-      setFlatpickrInstance(null);
-      return;
-    }
-
     let config = {
       allowInput: true,
       dateFormat: 'm/d/Y',
@@ -96,7 +91,7 @@ export default function Datepicker(props) {
       () => instance.calendarContainer.removeEventListener('mousedown', onCalendarMouseDown),
     ];
 
-  }, [ disabled, disallowPassedDates ]);
+  }, [ disallowPassedDates ]);
 
   // onChange is updated dynamically, so not to re-render the flatpicker every time it changes
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/form-js-viewer/src/render/components/form-fields/parts/Datepicker.js
```diff
@@ -51,6 +51,7 @@ export default function Datepicker(props) {
   useEffect(() => {
 
     if (disabled) {
+      flatpickrInstance && flatpickrInstance.setDate(null, true);
```
pinussilvestrus, line 54: This shows the placeholder in any case when disabled, also when valid input data was given.

image

Bildschirmfoto 2023-01-12 um 13 13 47

pinussilvestrus: This is inconsistent with other fields, where we display the initial value even on disabled.

image

Skaiir: Wait this confuses me a bit. Is the goal of 'disabled' to work like a read-only then?

Skaiir: I think a few components need adjustment then

pinussilvestrus: It's not `readonly` as it isn't submitted. However, it should be displayed if data is imported (cf. https://bpmn.io/blog/posts/2022-form-js-default-values-improved-submit.html).

pinussilvestrus: > I think a few components need adjustment then

Indeed it seems like this isn't working for the `taglist` as well.
image
image

I'll create an issue for that.

pinussilvestrus: Not that big of a deal, as we should discourage the use of `disabled` fields I believe (cf. #362)

pinussilvestrus: Here you go: https://github.com/bpmn-io/form-js/issues/496

Skaiir:
We now have the expected behavior: 

image

vsgoulart (approved):

PR #498, opened 2023-01-12
fix: decimal values for min/max number validation

### LINKED ISSUE #497: `number` min/max validation only supports integer numbers
__Describe the Bug__

Currently, it is not possible to define a maximum value for a number field which isn't an integer. The code allows for it but the properties panel field does not. 

__Steps to Reproduce__

1. Create a `number` field
2. Set validation:max to 5.3
3. Check the schema, the value isn't serialized

__Expected Behavior__

Since we now support decimal numbers, we should also support decimal number validation

Skaiir:
Closes #497

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/groups/ValidationGroup.js
+++ b/packages/form-js-editor/src/features/properties-panel/groups/ValidationGroup.js
@@ -235,6 +235,7 @@ function Min(props) {
     element: field,
     id,
     label: 'Minimum',
+    step: 'any',
     getValue: getValue('min'),
     setValue: onChange('min')
   });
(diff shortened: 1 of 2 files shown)


packages/form-js-editor/src/features/properties-panel/groups/ValidationGroup.js
```diff
@@ -235,6 +235,7 @@ function Min(props) {
     element: field,
     id,
     label: 'Minimum',
+    step: 'any',
```
vsgoulart, line 238: is this a standard config of the properties panel?

vsgoulart (approved):

packages/form-js-editor/src/features/properties-panel/groups/ValidationGroup.js
Skaiir: It's a standard config of number fields as per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#step. The properties panel components are pretty thin wrappers around the basic HTML components.

PR #501, opened 2023-01-14
fix: don't collapse dropdownlist on click scroll

### LINKED ISSUE #367: Cannot click the scrollbar on Taglist dropdown (dropdown component in general)
__Describe the Bug__

It is currently impossible to scroll by dragging the bar in a Taglist dropdown, this leads to the dropdown being closed instantly. 

image

__Steps to Reproduce__

1. Add a Taglist to a form
2. Render the form (via the playground for example)
3. Try and drag the above pictured segment of the control (scrollbar)

__Expected Behavior__

This should instead drag like normal. 

__Environment__

- Found on Firefox and Chrome, most likely all browsers

Skaiir:
Closes #367

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/parts/DropdownList.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/parts/DropdownList.js
@@ -87,6 +87,7 @@ export default function DropdownList(props) {
     ref={ dropdownContainer }
     tabIndex={ -1 }
     class="fjs-dropdownlist"
+    onMouseDown={ (e) => e.preventDefault() }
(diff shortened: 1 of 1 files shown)


Skaiir:
I've extended the area that captures clicks so that focus isn't lost on clicking the scrollbar, which was the cause of the bug (bluring the input causes the dropdown to close, which is intentional). 

I haven't provided a test case for this one, mostly because testing scrollbar clicks might be a bit of a challenge without a full on UI integration test. If you think this insufficient / know a way to do this easily, let me know :)

pinussilvestrus (approved):
Nice little improvement 👍

PR #509, opened 2023-01-23
feat: implement searchable select 

### LINKED ISSUE #381: Implement Searchable Select

As part of our initiative to enable more use cases in form-js, we want to enable our users to have a single select component capable of dealing with large-scale data. Unfortunately, the current select is not able to do so as it is not searchable, making it nonviable for any datasets of significant size.

image

There are two options here:

Either replace the current `Select` with this searchable alternative. Or develop it as a new component. Personally I fail to recognize the value in keeping the older version, but this is up for discussion.

Skaiir:
Closes #381


pinussilvestrus:
Let's update the [`schemaVersion`](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-viewer/src/index.js#L7) with this PR to make it explicit.

packages/form-js-viewer/src/render/components/form-fields/parts/SimpleSelect.js
```diff
@@ -0,0 +1,84 @@
...
+  const initialFocusIndex = useMemo(() => value && findIndex(options, (o) => o.value === value) || 0, [ options, value ]);
+
+  return <>
+    <div ref={ selectRef }
+      id={ prefixId(`${id}`, formId) }
+      class={ classNames('fjs-input-group', { 'disabled': disabled }, { 'hasErrors': errors.length }) }
+      onFocus={ () => setIsDropdownExpanded(true) }
+      onBlur={ () => setIsDropdownExpanded(false) }
+      onMouseDown={ (e) => onMouseDown(e) }
+      tabIndex={ disabled ? null : 0 }>
```
vsgoulart, line 70: `null` is not a valid value (probably it works because it's coerced to false)

```suggestion
      tabIndex={ disabled ? undefined : 0 }>
```

packages/form-js-viewer/src/render/components/form-fields/parts/SimpleSelect.js
```diff
@@ -0,0 +1,84 @@
...
+    ds.componentReady = !disabled && loadState === LOAD_STATES.LOADED;
+    ds.displayCross = ds.componentReady && value !== null && value !== undefined;
+    ds.displayDropdown = !disabled && isDropdownExpanded;
+    return ds;
+  }, [ disabled, isDropdownExpanded, loadState, value ]);
+
+  const onMouseDown = useCallback((e) => {
+    const select = selectRef.current;
+    setIsDropdownExpanded(!isDropdownExpanded);
+    isDropdownExpanded ? select.blur() : select.focus();
```
vsgoulart, line 57: I think we should only use ternaries for assignments to avoid bugs in future changes.
```suggestion
    if(isDropdownExpanded) {
      select.blur();
    } else {
      select.focus();
    }
```

packages/form-js-viewer/src/render/components/form-fields/parts/SimpleSelect.js
```diff
@@ -0,0 +1,84 @@
...
+
+  const {
+    state: loadState,
+    values: options
+  } = useValuesAsync(field);
+
+  // We cache a map of option values to their index so that we don't need to search the whole options array every time to correlate the label
+  const valueToOptionMap = useMemo(() => Object.assign({}, ...options.map((o, x) => ({ [o.value]: options[x] }))), [ options ]);
+
+  const valueLabel = useMemo(() => value && valueToOptionMap[ value ]?.label || '', [ value, valueToOptionMap ]);
```
vsgoulart, line 40: I believe we shouldn't use nullish coalescing, right @pinussilvestrus ?

packages/form-js-viewer/src/render/components/form-fields/parts/Timepicker.js
```diff
@@ -164,7 +164,7 @@ export default function Timepicker(props) {
           values={ timeOptions }
           height={ 150 }
           onValueSelected={ onDropdownValueSelected }
-          listenerElement={ timeInputRef.current }
+          keyEventsListener={ timeInputRef.current }
```
vsgoulart, line 167: Why do we need to change this name? I think the previous one was more descriptive

packages/form-js-viewer/src/render/components/form-fields/icons/XMark.svg
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="8px" height="8px" viewBox="0 0 8 8" version="1.1">
+<path style="stroke:black;stroke-width:0.5;fill-rule:evenodd;fill:rgb(0%,0%,0%);fill-opacity:1;"
```
vsgoulart, line 3: Let's not inline CSS
and let's use `fill: currentColor` so it's customizable

packages/form-js-viewer/src/render/components/form-fields/parts/SimpleSelect.js
Skaiir: I'd consider this a code style preference and if it's not in the linting rules, it's should be allowed IMO. Now we could make a case that this should be linted, but personally, I find it much more readable than the alternative.

packages/form-js-viewer/src/render/components/form-fields/parts/Timepicker.js
Skaiir: Yeah fair enough, I just had two names for this in the codebase by accident and just chose the wrong one to homogenize to.

packages/form-js-viewer/src/render/components/form-fields/parts/SimpleSelect.js
vsgoulart: It's not just code style preference because we support only `ES2018` and nullish coallescing is `ES2020` and we don't have any pollyfills in our build pipeline

vsgoulart:
@Skaiir Please when you answer comments feel free to request my review again. Otherwise I probably won't see the notifications :sweat_smile:

packages/form-js-viewer/src/render/components/form-fields/parts/SimpleSelect.js
Skaiir: @vsgoulart Should I then get rid of all optional chaining usages? I feel like it should have popped up as in issue in the past 2 months ([commit](https://github.com/bpmn-io/form-js/commit/67dc09f069bf26632ba6c01f3be566ac412163b8)).

vsgoulart: We could an Eslint rule for that, so we don't have to worry about it on reviews

pinussilvestrus: Last time we checked, nullish coalescing is indeed something we should avoid, as it is ES2020, as @vsgoulart mentioned (cf. https://github.com/bpmn-io/form-js/pull/448). Optional chaining we use already in the code base, e.g. [here](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/src/features/properties-panel/groups/ValidationGroup.js#L38). Still, I wonder why this doesn't blow up, isn't it also ES2020? Most likely, we transpile it via rollup, so we are safe. That's not the case for nullish coalescing.

Personally, I'd like to avoid these "newer" JS features, also for consistency reasons.

Skaiir: @pinussilvestrus We don't transpile it actually, you can find the optional chaining usages down to the desktop modeler code.

Skaiir:
@vsgoulart I think this should be it. I've changed the eslint parser to the ecmascript 2019 one, so there will be no uncertainty in the future. Considering our typescript config was on 2019 it just makes sense to adopt this and if we do want to upgrade we can do it 'officially' in the future.

vsgoulart (approved):

PR #512, opened 2023-01-23
fix: display `taglist` disabled state properly

### LINKED ISSUE #496: Disabled `taglist` fields do not show imported values
__Describe the Bug__

When specifying input data for a disabled `taglist` field, the viewer does not show the values, but instead only the placeholder.

image

image

This is inconsistent with other fields, where we still display the value while disabled. 

image

__Steps to Reproduce__

1. Create a `taglist`
2. Specify input data for that `taglist`
3. Disable the `taglist`
4. See no values are shown

__Expected Behavior__

Like for other fields, the `taglist` values should be shown also in `disabled` mode.

__Environment__

- Host (Browser/Node version), if applicable: Chrome 108
- OS: macOS
- Library version: 0.10.1

Skaiir:
Closes #496

### DIFF
--- a/packages/form-js-viewer/assets/form-js.css
+++ b/packages/form-js-viewer/assets/form-js.css
@@ -270,7 +270,7 @@
 }
 
 .fjs-container .fjs-input-group .fjs-input,
-.fjs-container .fjs-input-group.disabled .fjs-input {
+.fjs-container .fjs-input-group.fjs-disabled .fjs-input {
   border-radius: 3px;
 }
 
@@ -382,9 +382,9 @@
 .fjs-container .fjs-input:disabled,
 .fjs-container .fjs-textarea:disabled,
 .fjs-container .fjs-select:disabled,
-.fjs-container .fjs-number-arrow-container.disabled button,
-.fjs-container .fjs-input-group.disabled,
-.fjs-container .fjs-taglist.disabled {
+.fjs-container .fjs-number-arrow-container.fjs-disabled button,
+.fjs-container .fjs-input-group.fjs-disabled,
+.fjs-container .fjs-taglist.fjs-disabled {
   background-color: var(--color-background-disabled);
   border-color: var(--color-borders-disabled);
 }
@@ -396,8 +396,8 @@
   border-color: var(--color-borders-disabled);
 }
 
-.fjs-container .fjs-input-group.disabled .fjs-input,
-.fjs-container .fjs-input-group.disabled .fjs-input-adornment {
+.fjs-container .fjs-input-group.fjs-disabled .fjs-input,
+.fjs-container .fjs-input-group.fjs-disabled .fjs-input-adornment {
   border-color: var(--color-borders-disabled);
 }
 
@@ -473,7 +473,7 @@
   background-color: var(--color-grey-225-10-90);
 }
 
-.fjs-container .fjs-taglist .fjs-taglist-tag.disabled { 
+.fjs-container .fjs-taglist .fjs-taglist-tag.fjs-disabled { 
   background-color: var(--color-grey-225-10-85);
 }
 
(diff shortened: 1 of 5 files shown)


Skaiir:
firefox_dThAQ7E3ZE

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -128,22 +128,22 @@ export default function Taglist(props) {
...
       id={ prefixId(`${id}-search`, formId) } />
     <div class={ classNames('fjs-taglist', { 'disabled': disabled }) }>
-      {!disabled && loadState === LOAD_STATES.LOADED &&
+      { loadState === LOAD_STATES.LOADED &&
         <div class="fjs-taglist-tags">
           {
             values.map((v) => {
               return (
-                <div class="fjs-taglist-tag" onMouseDown={ (e) => e.preventDefault() }>
+                <div class={ classNames('fjs-taglist-tag', { 'disabled': disabled }) } onMouseDown={ (e) => e.preventDefault() }>
```
vsgoulart, line 136:
```suggestion
                <div class={ classNames('fjs-taglist-tag', { disabled }) } onMouseDown={ (e) => e.preventDefault() }>
```

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -128,22 +128,22 @@ export default function Taglist(props) {
...
                   <span class="fjs-taglist-tag-label">
                     {valueToOptionMap[v] ? valueToOptionMap[v].label : `unexpected value{${v}}`}
                   </span>
-                  <button
+                  { !disabled && <button
                     type="button"
                     title="Remove tag"
-                    class="fjs-taglist-tag-remove"
+                    class={ 'fjs-taglist-tag-remove' }
```
vsgoulart, line 143:
```suggestion
                    class="fjs-taglist-tag-remove"
```

packages/form-js-viewer/assets/form-js.css
```diff
@@ -483,6 +487,10 @@
   padding: 2px 6px 2px 8px;
 }
 
+.fjs-container .fjs-taglist .fjs-taglist-tag.disabled .fjs-taglist-tag-label {
+  padding: 2px 8px 2px 8px;
```
vsgoulart, line 491: Can be shortened a bit

```suggestion
  padding: 2px 8px;
```

vsgoulart (commented):
When I change the taglist selected values manually and then disable it, the taglist value resets to the initial value.
Is this expected?

Skaiir:
@vsgoulart Yeah it's expected, the form re-imports the initial state whenever settings changes are made.

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
Skaiir: This one I won't simplify as it would break if we ever renamed disabled. 

One disabled describes the css class while the other describes the condition for which it should be enabled. The fact that they are the same might change in the future, so for semantic and future-proofing reasons let's keep it this way.

Skaiir: Well okay unlikely to change, but still I think it's good practice to decouple them.

vsgoulart: I think it's not really necessary to separate but ok :sweat_smile: 

but one other thing related this. Shouldn't prefix the class name with `fjs-`

vsgoulart (approved):

Skaiir:
@vsgoulart You're right the fjs- prefix will make this safer. I'll include it and merge.

PR #513, opened 2023-01-23
feat(editor): emit `drag` events

### LINKED ISSUE #462: Add possibility to hook into dragula events

Sometimes, it makes sense to listen to specific `dragula` events.

One use case for `@camunda/form-playground` is to add custom styling during drag operations, add certain behaviors once an element is dropped, etc. 

Currently, [the `drake` instance is for internal use only](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/src/render/components/FormEditor.js#L177).

Make it possible to listen for specific `drake` events, e.g.

```js
formEditor.on('dragula.drag', ...);
formEditor.on('dragula.dragend');
formEditor.on('dragula.over');
```

Expose the whole `drake` instance.

```js
const drake = formEditor.get('drake');

drake.on('drag', ...);
```

However, this would lock our used dragging library pretty much, as replacing it would mean a breaking change. Note that we already fire such events as

```js
eventBus.on('dragula.created');
eventBus.on('dragula.destroyed');
```

__Additional context__

Related to https://github.com/bpmn-io/form-js/issues/461#issuecomment-1348157790

pinussilvestrus:
Closes #462

* `drag.start`
* `drag.end`
* `drag.drop`
* `drag.cancel`
* `drag.over`
* `drag.out`

Example

```js
formEditor.on('drag.start', function(event) {
  const { source } = event;
  source.classList.add('dragged');
});

formEditor.on('drag.drop', function(event) {
  const { source } = event;
  source.classList.remove('dragged');
});
```

Cf. drake events: https://github.com/bevacqua/dragula#drakeon-events

### DIFF
--- a/packages/form-js-editor/src/render/components/FormEditor.js
+++ b/packages/form-js-editor/src/render/components/FormEditor.js
@@ -181,6 +181,11 @@ export default function FormEditor(props) {
   };
 
   useEffect(() => {
+
+    const handleDragEvent = (type, context) => {
+      return eventBus.fire(type, context);
+    };
+
     const createDragulaInstance = () => {
       const dragulaInstance = dragula({
         isContainer(el) {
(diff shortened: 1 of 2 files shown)


nikku:
I'm not a fan of making this public API, because that leaks internals. 

I'd rather expose explicit dragging events for the design editor, with an interface agnostic to dragula (i.e. no `dragula*` in the name).

vsgoulart:
I think @nikku 's suggestion is good, this way we wouldn't introduce breaking changes into the API if we ever need to change the drag and drop implementation

pinussilvestrus:
I like these thoughts as well 👍 I covered this [in the alternatives](https://github.com/bpmn-io/form-js/issues/462), but recognized that we still have `dragula.x` events flying around that we also used to consume in the Desktop Modeler. [That might not be the case anymore, though](https://github.com/camunda/camunda-modeler/search?q=dragula). To keep it consistent, I kept the structure.

However, I also favor a more agnostic approach independent of dragula. You could also argue that the existing `dragula.created|destroyed` events are unrelated to the general dragging interaction. Still, a technical detail of the library that will be obsolete once we'd decide to move to another library.

Let me sketch that out further 👍

pinussilvestrus:
I adjusted the events to be more agnostic. I used the diagram-js [`Dragging` module ](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/dragging/Dragging.js)events as an inspiration. Check out the PR description for an update 👍

packages/form-js-editor/src/render/components/FormEditor.js
```diff
@@ -196,6 +201,31 @@ export default function FormEditor(props) {
         slideFactorY: 5
       });
 
+      // bind life cycle events
+      dragulaInstance.on('drag', (el, source) => {
+        handleDragEvent('drag.start', { element: el, source });
+      });
```
vsgoulart, line 207:
```suggestion
      dragulaInstance.on('drag', (element, source) => {
        handleDragEvent('drag.start', { element, source });
      });
```

vsgoulart (approved):

PR #521, opened 2023-01-26
feat: editor form fields extension point + migrate `Text` to it

Skaiir:
Properly implemented. 

Related to https://github.com/camunda/tasklist/issues/2440

### DIFF
--- a/packages/form-js-editor/src/render/components/FormEditor.js
+++ b/packages/form-js-editor/src/render/components/FormEditor.js
@@ -300,9 +300,7 @@ export default function FormEditor(props) {
     getService(type, strict = true) {
 
       // TODO(philippfromme): clean up
-      if (type === 'formFieldRegistry') {
-        return new Map();
-      } else if (type === 'form') {
+      if (type === 'form') {
         return {
           _getState() {
             return {
--- a/packages/form-js-editor/src/render/components/editor-form-fields/EditorText.js
+++ b/packages/form-js-editor/src/render/components/editor-form-fields/EditorText.js
@@ -1,0 +1,29 @@
+import { Text, isExpression } from '@bpmn-io/form-js-viewer';
+import { editorFormFieldClasses } from '../Util';
+
+import { iconsByType } from '../icons';
+
+export default function EditorText(props) {
+
+  const { type, text = '' } = props.field;
+
+  const Icon = iconsByType['text'];
+
+  if (!text) {
+    return <div class={ editorFormFieldClasses(type) }>
+      <div class="fjs-form-field-placeholder"><Icon viewBox="0 0 54 54" />Text view is empty</div>
+    </div>;
+  }
+
+  if (isExpression(text)) {
+    return <div class={ editorFormFieldClasses(type) }>
+      <div class="fjs-form-field-placeholder"><Icon viewBox="0 0 54 54" />Text view is populated by an expression</div>
+    </div>;
+  }
+
+  return <Text { ...{ ...props, disableLinks: true } } />;
+}
+
+EditorText.create = Text.create;
+EditorText.type = Text.type;
+EditorText.keyed = Text.keyed;
--- a/packages/form-js-editor/src/render/index.js
+++ b/packages/form-js-editor/src/render/index.js
@@ -1,9 +1,8 @@
-import { FormFields } from '@bpmn-io/form-js-viewer';
-
+import EditorFormFields from './EditorFormFields';
 import Renderer from './Renderer';
 
 export default {
   __init__: [ 'formFields', 'renderer' ],
-  formFields: [ 'type', FormFields ],
+  formFields: [ 'type', EditorFormFields ],
   renderer: [ 'type', Renderer ]
 };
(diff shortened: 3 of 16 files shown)


packages/form-js-viewer/src/util/index.js
```diff
@@ -133,10 +132,3 @@ export function getSchemaVariables(schema) {
   // remove duplicates
   return Array.from(new Set(variables));
 }
-
-
```
Skaiir, line 137: Moved this out and referred to the one in feel, that way we only have to change this in one place and won't get weird bugs.

packages/form-js-editor/src/render/components/editor-form-fields/EditorText.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  if (isExpression(text)) {
+    return <div class="fjs-form-field-placeholder"><Icon viewBox="0 0 54 54" />Text view is populated by an expression</div>;
+  }
+
+  return <Text { ...props } />;
+}
+
+// TODO: How could we get rid of the redundance here this ?
```
pinussilvestrus, line 22: Using `classes` maybe? But that would increase the complexity slightly, so I'd ok with having the redundancy here.

pinussilvestrus (commented):
I like this idea 👍 It uses our well-known dependency injection approach to have separate editor and viewer implementation of specific form fields.

packages/form-js-editor/src/render/components/FormEditor.js
```diff
@@ -270,9 +270,7 @@ export default function FormEditor(props) {
     getService(type, strict = true) {
 
       // TODO(philippfromme): clean up
-      if (type === 'formFieldRegistry') {
-        return new Map();
```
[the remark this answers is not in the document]
Skaiir: I searched the entire codebase and found that it simply is never called. 

There is however a registered version of `formFieldRegistry` in the editor's DI. My best guess is that it was once used for mocking, but since all usages of FormFieldRegistry in the editor package directly DI instead of using 'useService', this has never popped up. So I found a bug in advance basically.

packages/form-js-viewer/src/render/FormFields.js
```diff
@@ -5,9 +5,7 @@ export default class FormFields {
     this._formFields = {};
 
     formFields.forEach((formField) => {
-      const { type } = formField;
-
-      this.register(type, formField);
+      this.register(formField.type, formField);
```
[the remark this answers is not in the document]
Skaiir: Alright fair enough that was unnecessary, just in the flow of things I guess :D

packages/form-js-editor/src/render/index.js
```diff
@@ -1,9 +1,8 @@
-import { FormFields } from '@bpmn-io/form-js-viewer';
-
+import EditorFormFields from './EditorFormFields';
 import Renderer from './Renderer';
 
 export default {
   __init__: [ 'formFields', 'renderer' ],
-  formFields: [ 'type', FormFields ],
+  formFields: [ 'type', EditorFormFields ],
```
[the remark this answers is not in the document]
Skaiir: In fact there are already mechanisms in place to do this, additionalModules is implemented everywhere already.

packages/form-js-editor/src/render/components/editor-form-fields/EditorText.js
Skaiir: One thing I've considered also is to group all these meta-properties into a single object. But maybe a refactoring for the future :)

packages/form-js-viewer/src/render/FormFields.js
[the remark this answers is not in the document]
Skaiir: Ah wait no it's just the fact that this probably will never need to be changed, so destructuring for a single property seemed weird.

packages/form-js-editor/test/helper/index.js
```diff
@@ -176,4 +188,20 @@ export async function setEditorValue(editor, value) {
 
   // Requires 2 ticks to propagate the change to bpmn-js
   await act(() => {});
-}
\ No newline at end of file
+}
+
+export async function expectNoViolations(node, options = {}) {
```
pinussilvestrus, line 193:
```suggestion
export { expectNoViolations } from '../../form-js-viewer/test/helper';
```

Should we simply re-export the same helpers here?

packages/form-js-editor/test/TestHelper.js
```diff
@@ -33,4 +33,29 @@ export function insertStyles() {
...
+/**
+ * Create form container:
+ *
+ * <div class="fjs-container">
+ *   <div class="fjs-form"></div>
+ * </div>
+ *
+ * @returns {Element}
+ */
+export function createFormContainer() {
```
pinussilvestrus, line 45: Can we also re-export this one from the viewer test helpers?

pinussilvestrus (commented):
This looks awesome 👍 nice job, and thanks for the initiative to remove technical debt 🎉 

I'm leaving my ✅ as I know you wanted to add something on top. Feel free to merge 👍

PR #523, opened 2023-01-26
Datetime: correct `autocomplete` attribute

### LINKED ISSUE #522: Date time fields use an invalid `autocomplete` attribute
### Describe the Bug

Using the [IBM Equal access checker](https://chrome.google.com/webstore/detail/ibm-equal-access-accessib/lkcagbfjnkomcinoddgooolagloogehp) it throws an error for a form with date time fields.

> The 'autocomplete' attribute has an incorrect value

```html
<input
    data-input="true"
    autocomplete="false"
    placeholder="mm/dd/yyyy"
    class="fjs-input flatpickr-input"
    id="fjs-form-0nl2jhz-datetime1--date"
    type="text">
```

### Steps to Reproduce

1. Create a form with a date, time or date time field
2. Evaluate a11y via [IBM Equal access checker](https://chrome.google.com/webstore/detail/ibm-equal-access-accessib/lkcagbfjnkomcinoddgooolagloogehp) 
3. See error reported

### Expected Behavior

The date time fields should have a valid `autocomplete` attribute value. 

> autocomplete="off"

Cf. https://www.w3schools.com/tags/att_input_autocomplete.asp

### Environment

- Host (Browser/Node version), if applicable: Chrome 109
- OS: macOS 12.6
- Library version: `master`

pinussilvestrus:
Closes #522 

Apparently, we didn't test against the [`autocomplete-valid`](https://dequeuniversity.com/rules/axe/4.6/autocomplete-valid?application=RuleDescription) rule beforehand, as it was not part of the used test definitions. 

I added `cat.forms` to cover this; it makes anyway sense to use these best practices.

### DIFF
--- a/packages/form-js-viewer/test/TestHelper.js
+++ b/packages/form-js-viewer/test/TestHelper.js
@@ -6,6 +6,9 @@ import { insertCSS } from './helper';
 import formCSS from '../dist/assets/form-js.css';
 
 // @ts-ignore-next-line
+import flatpickrCSS from '@bpmn-io/form-js-viewer/dist/assets/flatpickr/light.css';
+
+// @ts-ignore-next-line
 import testCSS from './test.css';
 
 export function isSingleStart(topic) {
@@ -16,6 +19,7 @@ export function isSingleStart(topic) {
 
 function insertStyles() {
   insertCSS('form-js.css', formCSS);
+  insertCSS('flatpickr.css', flatpickrCSS);
   insertCSS('test.css', testCSS);
 }
 
--- a/packages/form-js-viewer/src/render/components/form-fields/parts/Datepicker.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/parts/Datepicker.js
@@ -176,7 +176,7 @@ export default function Datepicker(props) {
           class={ 'fjs-input' }
           disabled={ disabled }
           placeholder="mm/dd/yyyy"
-          autoComplete="false"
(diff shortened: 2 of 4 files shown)


packages/form-js-viewer/test/TestHelper.js
```diff
@@ -5,6 +5,9 @@ import { insertCSS } from './helper';
 // @ts-ignore-next-line
 import formCSS from '../dist/assets/form-js.css';
 
+// @ts-ignore-next-line
+import flatpickrCSS from '@bpmn-io/form-js-viewer/dist/assets/flatpickr/light.css';
```
vsgoulart, line 9: Is this only for tests? Was it already included in the actual implementation?

vsgoulart (approved):

packages/form-js-viewer/test/TestHelper.js
pinussilvestrus: This is only for tests, yes. We forgot this one for our test suite, as it does not make a big difference for the `npm run start:viewer` test, only visually. I recognized it while inspecting the viewer test in the browser.

Skaiir (approved):

packages/form-js-viewer/test/TestHelper.js
vsgoulart: Yeah, if this changed anything for the actual implementation we would have to check if it doesn't break the customization on Tasklist

PR #525, opened 2023-01-27
feat: normalized font and misc styling form-wide

### LINKED ISSUE #506: Establish rules for fonts within input fields and apply consistently.
### What should we do?

Currently, there is a discrepancy between fonts within inputs and fonts in the general UI. Fonts in inputs are sized to 13.3333px while fonts outside of it are sized to 14px. There are also some further inconsistencies in terms of disabled font color, and styling of fonts in these states. Styling of placeholders, etc. 

Let's carry out a discussion on what we want these things to look like, establish some rules and clean up across all components. The implementation itself should be rather trivial, but there's certainly some back and forth that can occur in terms of decision-making.  

For reference here is what an input field looks like currently at 13+1/3px
image

And how it would look with a 14px font size
image

### Why should we do it?

To offer a consistent UX across components, and if relevant in this scenario to keep the readability of our components high (a11y)

Skaiir:
Closes #506

This ended up becoming a bit larger than just the fonts.

### DIFF
--- a/packages/form-js-editor/assets/form-js-editor.css
+++ b/packages/form-js-editor/assets/form-js-editor.css
@@ -76,11 +76,8 @@
 .fjs-editor-container .fjs-element {
   display: flex;
   flex-direction: column;
-  margin: 0 5px;
-}
-
-.fjs-editor-container .fjs-element {
-  margin-top: 4px;
+  margin: 0 4px 5px 4px;
+  padding: 2px 0px;
 }
 
 .fjs-editor-container .fjs-element:first-of-type {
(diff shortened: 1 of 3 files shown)


Skaiir:
Will provide a preview in a bit @RomanKostka

Skaiir:
Preview available here: https://demo-506-font-styling-normalization--camunda-form-playground.netlify.app/

RomanKostka (approved):
nice, thanks!
looks great from my point of view.

vsgoulart (approved):
Looks great, thanks for the effort

vsgoulart:
@Skaiir it seems have failing tests

PR #531, opened 2023-02-01
feat: provide validations for decimal digits prop

Skaiir:
Related to https://github.com/camunda/camunda-modeler/issues/3382


vsgoulart (commented):
Implementation is ok, but there's a merge conflict in the locakfile

Issue #533, opened 2023-02-02
Support historical dates in `Datetime`


Skaiir:
### Describe the Bug

We have disabled dates prior to 1900 to get around a bug related to historical time zones in the date picker. It would be good to instead fully support this component for old historical dates. 

The original bug can be found [here](https://github.com/camunda/camunda-modeler/issues/3380)

### Steps to Reproduce

1. Create a Date/Time picker
2. Try to set the date to any day in 1850

### Expected Behavior

The year should ultimately be selectable, even if it is a very niche use case :)

PR #534, opened 2023-02-02
chore(datepicker): disable dates prior to 1900

Skaiir:
Just going to straight up disable the dates which were causing issues for now, so that we don't have the bug in the next release. We've agreed anyways that the historical use case is not a requirement for this component, fixing this is more effort than I'd like so we can take it as a future item:

Have not provided a test case as it is not really the behavior we want long term. 

Proper fix is tracked [here](https://github.com/bpmn-io/form-js/issues/533).


pinussilvestrus:
As discussed, let's add a test case still. This way we ensure it's not blowing up and we can adjust it later to the behavior we aim for.

pinussilvestrus (approved):

PR #537, opened 2023-02-02
fix(editor): use constant value for custom validation type

pinussilvestrus:
Related to https://github.com/camunda/camunda-modeler/issues/3379#issuecomment-1413802106

Seems like some browsers do not work properly with `undefined` as option type. I think we should avoid this anyway.

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/groups/ValidationGroup.js
+++ b/packages/form-js-editor/src/features/properties-panel/groups/ValidationGroup.js
@@ -19,7 +19,7 @@ import { INPUTS } from '../Util';
 
 const VALIDATION_TYPE_OPTIONS = {
   custom: {
-    value: undefined,
+    value: '',
     label: 'Custom',
   },
   email: {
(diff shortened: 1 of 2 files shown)


packages/form-js-editor/src/features/properties-panel/groups/ValidationGroup.js
```diff
@@ -274,7 +274,12 @@ function ValidationType(props) {
   const debounce = useService('debounce');
 
   const setValue = (validationType) => {
-    onChange('validationType')(validationType || undefined);
+
+    if (!validationType) {
```
Skaiir, line 278: From a quick look, this probably doesn't do anything? validationType is only nullish if the value is `VALIDATION_TYPE_OPTIONS.custom.value OR ''` which is the same thing as `validationType` in this context, right? In this case we would still want to use `|| undefined` I think, which will instead of setting the schema to an empty string, entirely clear the property. Though that's optional.

pinussilvestrus: You are right, this was a leftover from my root causing. I remove the change to keep it untouched, makes sense at ist.

Skaiir (approved):

PR #538, opened 2023-02-03
fix: text adornment ellipsis

Skaiir:
Related to https://github.com/camunda/camunda-modeler/issues/3381


Skaiir:
Ellipsis doesn't work directly on flex elements, so I nested text within a span, also I disabled whitespace wrapping.

pinussilvestrus (approved):
LGTM 👍 

I added an a11y to the text field with adorners to ensure we have it covered. I also added a class name to the new span element, as it makes it easier to customize these elements.

PR #543, opened 2023-02-13
542 select components can consume a simple array of strings for options

### LINKED ISSUE #542: Select components can consume a simple array of strings for options

The select components can consume process variables from the input context to display dynamic options. However, the options need to be provided in the following format

```
{
    "value": String,
    "label": String
}
```
thus the differentiation between value and label is mandatory.

If not provided in this format, the whole component breaks and produces null values.

Even worse, there is no inline documentation (e.g. tooltip) that tells the user what format is expected. Hence, the majority of users fail to use the component.

Not all input select options are available in this format, and not all select options need a separation between values and labels.

Require the form user to do transformations in the embedding bpmn or other surroundings. But this is obsolete in many ways.

christian-konrad:
Closes #542 

You can now provide any mixture of input options, and without breaking the select components and output.

image

This is a draft PR, lacking any tests.

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/InputKeyValuesSourceEntry.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/InputKeyValuesSourceEntry.js
@@ -15,7 +15,7 @@ export default function InputKeyValuesSourceEntry(props) {
 
   const description = <div>
     Define which input property to populate the values from.
-    <br /><br />The input property must follow this schema:
+    <br /><br />The input property may be an array of simple values or alternatively follow this schema:
     <pre><code>{schema}</code></pre>
   </div>;
 
--- a/packages/form-js-viewer/src/render/components/util/valuesUtil.js
+++ b/packages/form-js-viewer/src/render/components/util/valuesUtil.js
@@ -1,0 +1,49 @@
+import { get } from 'min-dash';
+
+// parses the options data from the provided form field and form data
+export function getValuesData(formField, formData) {
+  const { valuesKey, values } = formField;
+  return valuesKey ? get(formData, [ valuesKey ]) : values;
+}
+
+// transforms the provided options into a normalized format, trimming invalid options
+export function normalizeValuesData(valuesData) {
+  return valuesData.filter(_isValueSomething).map(v => _normalizeValueData(v)).filter(v => v);
+}
+
+function _normalizeValueData(valueData) {
+
+  if (_isAllowedValue(valueData)) {
+
+    // if a primitive is provided, use it as label and value
+    return { value: valueData, label: valueData };
+  }
+
+  if (typeof (valueData) === 'object') {
+    if (!valueData.label && _isAllowedValue(valueData.value)) {
+
+      // if no label is provided, use the value as label
+      return { value: valueData.value, label: valueData.value };
+    }
+
+    if (_isValueSomething(valueData.value) && _isAllowedValue(valueData.label)) {
+
+      // if both value and label are provided, use them as is
+      return valueData;
+    }
+  }
+
+  return null;
+}
+
+function _isAllowedValue(value) {
+  return _isAllowedValueType(value) && _isValueSomething(value);
+}
+
+function _isAllowedValueType(value) {
+  return [ 'number', 'string' ].includes(typeof(value));
+}
+
+function _isValueSomething(value) {
+  return value || value === 0;
+}
--- a/packages/form-js-viewer/src/render/components/util/valuesUtil.js
+++ b/packages/form-js-viewer/src/render/components/util/valuesUtil.js
@@ -16,19 +16,19 @@ function _normalizeValueData(valueData) {
   if (_isAllowedValue(valueData)) {
 
     // if a primitive is provided, use it as label and value
-    return { value: valueData, label: valueData };
+    return { value: valueData, label: `${ valueData }` };
   }
 
   if (typeof (valueData) === 'object') {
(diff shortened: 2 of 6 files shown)


pinussilvestrus:
TIL Netlify seems to have its limit in their URLs for previews: https://demo-542-select-components-can-consume-a-simple-array-of-strings-for-options--camunda-form-playground.netlify.app/

Please use this one for the preview: https://63f4e69b5de7520008c9ef39--camunda-form-playground.netlify.app/

pinussilvestrus:
Should we update the description for the input values key as well? Cf. https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/src/features/properties-panel/entries/InputKeyValuesSourceEntry.js#L14

packages/form-js-viewer/src/render/components/util/valuesUtil.js
```diff
@@ -0,0 +1,49 @@
...
+function _isAllowedValue(value) {
+  return _isAllowedValueType(value) && _isValueSomething(value);
+}
+
+function _isAllowedValueType(value) {
+  return [ 'number', 'string' ].includes(typeof(value));
+}
+
+function _isValueSomething(value) {
+  return value || value === 0;
```
pinussilvestrus, line 48: Damn Javascript 😄

packages/form-js-viewer/src/render/components/util/valuesUtil.js
```diff
@@ -0,0 +1,49 @@
...
+function _normalizeValueData(valueData) {
+
+  if (_isAllowedValue(valueData)) {
+
+    // if a primitive is provided, use it as label and value
+    return { value: valueData, label: valueData };
+  }
+
+  if (typeof (valueData) === 'object') {
+    if (!valueData.label && _isAllowedValue(valueData.value)) {
```
pinussilvestrus, line 23: Quick question: Why don't we also allow for

```json
[
  { "label": "foo" }
]
```

and use the label as value as well?

packages/form-js-viewer/test/spec/render/components/form-fields/util/valuesUtil.spec.js
```diff
@@ -0,0 +1,180 @@
...
+
+      // when
+      const result = normalizeValuesData(valuesData);
+
+      // then
+      expect(result).to.eql([ { value: 'john', label: 'john' }, { value: 'jessica', label: 'jessica' } ]);
+    });
+
+
+    it('should filter out random objects', function() {
```
pinussilvestrus, line 85:
```suggestion
    it('should filter out invalid objects', function() {
```

Just to prevent we do not know what random objects mean once we have a look at this time again 🙂

packages/form-js-viewer/test/spec/render/components/form-fields/util/valuesUtil.spec.js
```diff
@@ -0,0 +1,180 @@
...
+
+      // then
+      expect(result).to.eql([ { value: 0, label: 0 } ]);
+    });
+
+
+    it('should ignore boolean definitions', function() {
+
+      // given
+      const valuesData = [ true, false ];
```
pinussilvestrus, line 156: Mh, as I think about this, couldn't that also be a valid use case to specify values for a simple true/false radio? Maybe doesn't make sense, as you could use a checkbox for this, but just in theory.

pinussilvestrus (commented):
Works like a charm, pretty cool 👍

packages/form-js-viewer/src/render/components/util/valuesUtil.js
Skaiir: I guess the simplest way I think about it:

It's okay to derive metadata from state data
It's wrong to derive state data from metadata

pinussilvestrus: > I guess the simplest way I think about it:
> 
> It's okay to derive metadata from state data It's wrong to derive state data from metadata

💯 nice thinking, I agree 👍

packages/form-js-viewer/test/spec/render/components/form-fields/util/valuesUtil.spec.js
Skaiir: Fair point 🤔, well okay it doesn't hurt much to allow that anyways, booleans have a clear string representation.

Skaiir:
@pinussilvestrus 

- Added description detail
- Allowed bools to be used

Skaiir:
Documentation PR: https://github.com/camunda/camunda-platform-docs/pull/1723

Following approval I'll squash and commit

packages/form-js-editor/src/features/properties-panel/entries/InputKeyValuesSourceEntry.js
```diff
@@ -15,7 +15,7 @@ export default function InputKeyValuesSourceEntry(props) {
 
   const description = <div>
     Define which input property to populate the values from.
-    <br /><br />The input property must follow this schema:
+    <br /><br />The input property may be an array of primitives or alternatively follow this schema:
```
pinussilvestrus, line 18: Can we expect our users to know what "primitives" mean? Should we maybe show this via example as well? This could blow up the description, though.

Skaiir: Exactly, that's why I opted against it. This is explained in more detail in the docs anyways. I'll rename it to `array of simple values`.

pinussilvestrus (approved):

PR #553, opened 2023-02-21
Implement forms row layout

### LINKED ISSUE #560: Implement implicit row-based layout solution
### What should we do?

Implement solution of implicit row-based layout solution.

Alongside [the spike](https://github.com/bpmn-io/form-js/pull/553) we need to adjust
- [x] Dragging
    - [x] Refactor drop handlers
    - [x] implement drop between rows
- [x] FormLayouter
  - [x] support nested components 
- [x] MoveField
  - [x] add revert support 
- [x] LayoutGroup
    - [x] split entries
    - [x] consider dropdown for columns
- [x] <del>Resizer</del>
    - add as follow up
    - Check out: https://github.com/react-grid-layout/react-resizable
- [x] Previews
    - [x] rework row drag preview
    - [x] rework drop preview
    - [x] rework field drag preview
- [x] Styles
    - [x] align fields in row vertically (line height)
    - [x] update selection + context pad styles
- [x] Maintain layout rules 
    - [x] restrict 4 fields per row
    - [x] restrict to max 16 cols per row
    - [x] on drop: show disallow preview (similar to BPMN editor)
    - [x] [properties panel: validate dropdown](https://github.com/bpmn-io/properties-panel/pull/230)
- [x] Debug columns
    - [x] hide behind option flag
    - alternative: add grid to editor
- [ ] Dry run with a nested component  
- [x] add proper test coverage

### Why should we do it?

Child of https://github.com/camunda/team-hto/issues/50

pinussilvestrus:
Closes #560 
Depends on https://github.com/bpmn-io/properties-panel/pull/230

[Check out the demo here](https://demo-spike-forms-row-layout--camunda-form-playground.netlify.app/).

----

This includes
* Add [Carbon grid](https://carbondesignsystem.com/guidelines/2x-grid/overview/) under the hood
* Support `columns` in the viewer
* Refactor `dragging` into module
* Support `columns` in the editor
* Validating rows restrictions
  * prevent overflows: maximum 16 columns per row
  * maximum 4 fields per row

(Potential) follow ups
* https://github.com/bpmn-io/form-js/issues/566
* https://github.com/bpmn-io/form-js/issues/569
* https://github.com/bpmn-io/form-js/issues/571
* https://github.com/bpmn-io/form-js/issues/572

image


pinussilvestrus:
@Skaiir @RomanKostka @vsgoulart @christian-konrad I opened up the PR for review 🎉 Feel free to try out the demo and let me know what you think.

I added a note about the follow-ups in the PR description.

Skaiir:
There seems to be a slightly strange behaviour where dragging a row to the end doesn't always insert at the end (needs to be in the right half of the viewport for some reason).

firefox_s5kwalC2L9

Skaiir:
This is personal opinion ultimately but I also think the drag preview should be more compact: 

- overall reducing the padding would help I think
- there **needs** to be a vertical max, see below

firefox_dQASk4XelJ

RomanKostka (commented):
nice!
i found some small issues which are not critical:
* the padding of the components and the selection border seems a bit off -> we should aim to have a 6px padding like this: 
image
instead of this: 
image

*  would be interested in your opinion here: i am not quite sure if we should stay with the `align-items: middle` or move to a top-alignment like this: 
image
* it would be awesome if we would also have a hover and select state for rows
* we have to decide for a border radius. For now it works fine but as soon as we switch to carbonized styling we should remove the border-radius to 0. From my point of view we can do it now already

Skaiir:
+1 for top-alignment

RomanKostka:
from my point of view it would also be sufficient to only show the drag row icon when i hover over the row.
With that we could reduce the amount of icons on the screen.
image

what do you think?

pinussilvestrus:
> regarding the draggable rows:
> 
> > I would question if it even is necessary to preview the contents here to be honest, it's not like we preview the individual components when we drag and drop them.
> 
> Kinda agree, it gets a bit messy when you drag large components. But on the other side it feels natural because you see what you are dragging.
> 
> It also is not 100% consistent with dragging a single component -> there you don't see a preview of the component.
> 
> What we could do is adding a placeholder for dragging a row like this: image This is probably something we could test with some users and collect feedback
> 
> However i would propose to add a reduced opacity to the drag preview. And soften out the drop shadow for dragged components + rows ( e.g. `drop-shadow(0px 0px 20px rgba(0, 0, 0, 0.15));` )

I like the preview much more, but I agree that this would need more validation. I'll at least make sure the preview does not get too much height

packages/form-js-editor/assets/form-js-editor-base.css
```diff
@@ -1,17 +1,19 @@
 .fjs-editor-container {
   --color-children-border: var(--color-grey-225-10-75);
-  --color-children-selected-border: var(--color-blue-205-100-50);
-  --color-children-selected-background: var(--color-blue-205-100-95);
-  --color-children-hover-border: var(--color-blue-205-100-50);
+  --color-children-selected-border: hsl(219, 99%, 53%);
```
Skaiir, line 3: Should we tidy these up?

pinussilvestrus: I'll look whether I find a good form-js representation of this, will anyway be overridden via #557 rather soon.

pinussilvestrus:
@RomanKostka @Skaiir I made the changes to the visuals and resolves the bug with the bottom row drop. Let me know what you think 👍

christian-konrad (changes requested):
Overall, this looks and feels super smooth! Thank you @pinussilvestrus 

However, I found some smaller issues:

1. a slight mismatch between the preview and editor. 
* In the preview, the margins/paddings between fields are too wide, it looks good in the editor
* In the editor, the text view is not top-aligned, it is better in the preview

Bildschirmfoto 2023-03-24 um 09 21 43

Bildschirmfoto 2023-03-24 um 09 25 32

2. The layout breakpoints might not be ideal, the layouts jump too late and it results in squeezed components (especially the date picker which is essentially two fields)

Bildschirmfoto 2023-03-24 um 09 27 34

https://user-images.githubusercontent.com/56470362/227466304-93fb2602-b1c3-4ff2-bba2-e448c1e9d928.mov

I think we could make it easier and say there are only two layouts / one breakpoint: Row-based (16 cols) and single col

pinussilvestrus:
> I think we could make it easier and say there are only two layouts / one breakpoint: Row-based (16 cols) and single col

Thanks for raising this. That's how it should be implemented, so it's likely a bug somewhere. I will add a test case for this as well.

RomanKostka:
Love it!

something we could consider as a follow up:
[redesign the delete icon ](https://www.figma.com/file/CiTSN1vNETjIAy6W9jUIFp/Form-Editor?node-id=851%3A133246&t=gaz84Ucgk2SC1AX8-1)- currently it seems not clear that it has a hover state

Besides that i found some things:

1. the drag row icon could have a larger area.
it is really hard to hit it with the mouse at the moment. An area like this would be super nice 
image

2. when changing the state of an component from idle to dragged it seems that the component + border is jumping a bit around (~1 or 2 px) - not sure why this happens but it would be so clean if that would not happen.
image

https://user-images.githubusercontent.com/114994124/227592671-ed21087c-68b9-4bc5-b29a-f22b876b0f92.mov



3. border of rows are still missing when hovering over the drag-row icon - not sure if this is intended

4. i realised that the drop-preview of single components has a different height then the standard component. would be nice if we match the drop-preview to the standard height. currently it produces some weird look and feel 
image

5. when trying to place components into 'full' rows the 'not-able-to-drop' indicator is jumping around unexpectedly


https://user-images.githubusercontent.com/114994124/227592409-c956af0e-731f-4703-a5e9-7a366fba31bf.mov

pinussilvestrus:
> 3\. border of rows are still missing when hovering over the drag-row icon - not sure if this is intended

I didn't add borders to the rows since I think the full behavior of selecting rows is a bit unclear at the moment. Let's follow up on this I'd propose, I will create an issue for that!

Cf. the other things, I will have a look, nice findings 👍

pinussilvestrus:
@RomanKostka I made some adjustments, let me know what you think.

> the drag row icon could have a larger area.
> it is really hard to hit it with the mouse at the moment. An area like this would be super nice

I fixed that by giving the area a bit more space (32x32px).

> when changing the state of a component from idle to dragged it seems that the component + border is jumping a bit around (~1 or 2 px) - not sure why this happens but it would be so clean if that would not happen.

This is not super easy to fix as a) the drop preview is a fixed element with fixed styles, and b) the form fields are different in their styling. I would not go for implementing drop previews for each individual form element, but I adjusted the height and margins so that it shouldn't jump around now. Still, it could not be super perfect on some occasions.

> i realised that the drop-preview of single components has a different height then the standard component. would be nice if we match the drop-preview to the standard height. currently it produces some weird look and feel

This is fixed with the preview adjustments mentioned above

> when trying to place components into 'full' rows the 'not-able-to-drop' indicator is jumping around unexpectedly

`dragula` decides where to place the drop previews given on some coordinates calculation. I am not sure whether we can make this even more fine graines so I guess we have to deal with that.

packages/form-js-editor/src/render/components/icons/Draggable.svg
```diff
@@ -0,0 +1,14 @@
...
+</style>
+<title>draggable</title>
+<rect x="10" y="6" width="4" height="4"/>
+<rect x="18" y="6" width="4" height="4"/>
+<rect x="10" y="14" width="4" height="4"/>
+<rect x="18" y="14" width="4" height="4"/>
+<rect x="10" y="22" width="4" height="4"/>
+<rect x="18" y="22" width="4" height="4"/>
+<rect id="_Transparent_Rectangle_" class="st0" width="32" height="32"/>
+</svg>
```
vsgoulart, line 14: You can reduce a bit on the SVG size if you this one:

```svg
<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" width="16" height="16" fill="currentcolor" viewBox="0 0 32 32"><path d="M10 6h4v4h-4zm8 0h4v4h-4zm-8 8h4v4h-4zm8 0h4v4h-4zm-8 8h4v4h-4zm8 0h4v4h-4z"/><path d="M0 0h32v32H0z" style="fill:none"/></svg>
```

[Optimized via this website](https://jakearchibald.github.io/svgomg/)

packages/form-js-editor/src/features/dragging/Dragging.js
```diff
@@ -0,0 +1,350 @@
...
+          this._formFieldRegistry.get(targetRow.components[targetRow.components.length - 1])
+        ) + 1;
+      } else {
+
+        // (3) dropped as last item
+        targetIndex = targetFormField.components.length;
+      }
+    }
+
+    return targetIndex;
```
vsgoulart, line 72: I think we can make it a little bit easier to read by using some early returns

```suggestion
if (siblingFormField) {
  return getFormFieldIndex(targetFormField, siblingFormField);
}

// (2) dropped in row => place at the end of row (after last field in row)
if (targetRow) {
  return (
    getFormFieldIndex(
      targetFormField,
      this._formFieldRegistry.get(
        targetRow.components[targetRow.components.length - 1],
      ),
    ) + 1
  );
}

// (3) dropped as last item
return targetFormField.components.length;
```

packages/form-js-editor/src/features/modeling/FormLayoutUpdater.js
```diff
@@ -0,0 +1,53 @@
...
+
+  constructor(eventBus, formLayouter, modeling, formEditor) {
+    super(eventBus);
+
+    this._eventBus = eventBus;
+    this._formLayouter = formLayouter;
+    this._modeling = modeling;
+    this._formEditor = formEditor;
+
+    // @ts-ignore
```
vsgoulart, line 18: Why is this necessary?

packages/form-js-editor/src/features/properties-panel/entries/ColumnsEntry.js
```diff
@@ -1,68 +1,90 @@
...
+  };
+
+  const getOptions = () => {
+    return [
+      {
+        label: 'Auto',
+        value: AUTO_OPTION_VALUE
+      },
+
+      // todo(pinussilvestrus): make options dependant on field type
```
vsgoulart, line 65: I guess you forgot about this todo?

packages/form-js-editor/src/render/components/FormEditor.js
```diff
@@ -125,12 +142,42 @@ function Element(props) {
...
+        position: absolute;
+        bottom: -2px;
+        z-index: 2;
+        font-size: 10px;
+        right: 3px;"
+      class="fjs-debug-columns">
+      { (field.layout || {}).columns || 'auto' }
+    </div>
+  );
+}
```
vsgoulart, line 173: Do you want to leave this here or you forgot about it?

vsgoulart (commented):
A small problem I found:

The cursor seems to be a bit weird, it only shows the dragging cursor if you grab on the left icon for the entire row. I think it should be any time you drag any element

RomanKostka (commented):
awesome!

i found some small misalignments:
1. the drop preview area is now off to the left side - could be because of the larger drag row indicator?
image
2. the drop preview border is overlapping with the select border - not important but a detail i saw
image

3. (future topic) we should have a look into the sizes of each component. It feels weird having elements with different sizes when they still need the same visual space in a row.
image

packages/form-js-editor/src/features/properties-panel/entries/ColumnsEntry.js
vsgoulart: Maybe let's link to an issue? So it's not completely forgotten as just comment

packages/form-js-editor/src/core/FormLayoutValidator.js
```diff
@@ -0,0 +1,72 @@
...
+
+      // allow maximum 16 cols
+      if (columns > 16) {
+        return 'Maximum 16 columns are allowed';
+      }
+    }
+
+    if (!row) {
+      row = this._formLayouter.getRowForField(field);
+    }
```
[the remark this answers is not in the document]
pinussilvestrus: I actually like the option above a bit more, not a huge fan of such extensive argument declaration.

packages/form-js-viewer/test/spec/render/components/FormField.spec.js
```diff
@@ -281,6 +288,23 @@ describe('FormField', function() {
...
+      const { container } = createFormField();
+
+      // then
+      const layout = container.querySelector('.fjs-layout-column');
+
+      expect(classes(layout).has('cds--col-lg-8')).to.be.true;
+      expect(classes(layout).has('cds--col-sm-16')).to.be.true;
+    });
+
+  });
```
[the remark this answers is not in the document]
pinussilvestrus: I agree that we should invest in such tests. However I think the test makes sense as it verifies that the correct classes are set from the schema.

pinussilvestrus:
> The cursor seems to be a bit weird, it only shows the dragging cursor if you grab on the left icon for the entire row. I think it should be any time you drag any element

I think this makes sense, the only thing I could imagine is that it's quite weird that for most of the canvas, the dragging cursor will be active now. But semantic-wise, it makes much sense. /cc @RomanKostka

packages/form-js-editor/src/features/modeling/FormLayoutUpdater.js
pinussilvestrus: Let me double check, it could be that the missing types of `diagram-js` were striking here.

packages/form-js-editor/src/core/FormLayoutValidator.js
```diff
@@ -0,0 +1,72 @@
...
+
+    // calculate columns with and without updated field
+    let sumColumns = (parseInt(columns) || 0);
+    let sumFields = 1;
+    let sumAutoCols = columns ? 0 : 1;
+
+    row.components.forEach(id => {
+      if ((field || {}).id === id) {
+        return;
+      }
```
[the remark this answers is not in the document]
pinussilvestrus: 👍

packages/form-js-editor/src/features/properties-panel/groups/LayoutGroup.js
```diff
@@ -0,0 +1,23 @@
...
+
+  if (type === 'default') {
+    return null;
+  }
+
+  const entries = [
+    ...ColumnsEntry({ field, editField })
+  ];
+
+  if (!entries.length) {
```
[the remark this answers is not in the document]
pinussilvestrus: 👍

packages/form-js-editor/src/features/modeling/FormLayoutUpdater.js
pinussilvestrus: Yes indeed, it's due to lacking diagram-js types. We have similar occasions in the code base (modeling behaviors, command interceptors).

packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
```diff
@@ -3028,6 +3029,7 @@ function createPropertiesPanel(options = {}) {
   let {
     eventBus,
     formEditor,
+    formLayoutValidator,
```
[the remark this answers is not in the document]
pinussilvestrus: I agree that, for this case, it's easier to read, but for the sake of consistency with the other mocks, I'd like to keep it as it is.

RomanKostka (commented):
looks good to me - lets iterate on the visuals after the release.

two things we should take care of in next iterations:
the 'not allowed' curser disappears for some components.
also the space between the rows seem a bit large and could be reduced.

https://user-images.githubusercontent.com/114994124/228210679-5e981593-5ad4-4708-8f8f-f23d235b6bd7.mov

pinussilvestrus:
> also the space between the rows seem a bit large and could be reduced.

I reduced it a bit, still, we need some space to make it easy to drag between rows.

For the cursor problem I opened a follow up: https://github.com/bpmn-io/form-js/issues/576

RomanKostka (approved):

vsgoulart (approved):

Issue #257, opened 2022-05-13
Improve the way we handle CSS to accomodate for the growing complexity of the project


Skaiir:
__Is your feature request related to a problem? Please describe__

As our component library grows the CSS files used to describe them are starting to become more complicated and harder to read. It would be good for us to start using CSS pre-processors instead.

__Describe the solution you'd like__

Adopt a CSS preprocessor like LESS and integrate it all into our CI.

__Describe alternatives you've considered__

n/a

__Additional context__

n/a

MaxTru:
Why not use LESS (e.g., we also use it in camunda-modeler and plan to use it for propPanel: https://github.com/bpmn-io/properties-panel/issues/48)

pinussilvestrus:
My two cents from a dev perspective: I personally never had big issues from working with a single "vanilla" CSS file. I experienced many headaches from using pre-processors as they add more magic to your bundling pipeline. :sweat_smile: Most of the time, I was just happy to use my well-known CSS.

Things, as we have in the Camunda Modeler to split styles, are cool to keep things focused. Having a central place would be fine, too. I have no strong feelings about this. 

By the way, a cool thing I like [about Svelte](https://svelte.dev/tutorial/styling); having the styles in the same file as the component, and then the compiler simply bundles your CSS together. 

I think the styles in our libraries still have a reasonable size. That might change in the future, though, but maybe then a better question would be whether to split the libraries, not the styles.

marstamm:
Some thoughts on this one:
**As a Developer**
- I like using vanilla CSS, I know what the selectors do and don't have to remember what special operators do in sass or less.
- When I was refactoring Drilldown CSS, I missed some class names in the big CSS files. It is not always clear from the names which style is required by which component.
- It would be great to have Code spliting for CSS, so I can easily see what styles belong to the component and makes refactoring easier. We could use a [rollup plugin](https://www.npmjs.com/package/rollup-plugin-import-css) to include them in JS and bundle them all in a single CSS file for the distro.
- This can also result in smaller CSS files for viewers, if they don't have to include all modeler styles.

**As a library user**
- I don't want to care about CSS of the libraries I import. I expect them to work out of the box.
- Tools like codemirror inline their styles and don't require you to include a CSS file in you app.
[...]

PR #557, opened 2023-03-02
Theming + Carbonization

pinussilvestrus:
Closes https://github.com/bpmn-io/form-js/issues/515
Related to https://github.com/camunda/team-hto/issues/143

-----

Makes it possible to theme form viewer + editor by
* Mirror [Carbon Theme variables](https://carbondesignsystem.com/guidelines/themes/overview/)
* Make icons themeable
* Add simple starts for theming (`SINGLE_START=theme`)
* Add another package `@bpmn-io/form-js-carbon-styles` to make it easier to integrate custom form styles
  * Move carbon styles from Tasklist
  * Remove some styles as they are handled in the core styles now (mostly colors)

This is a step more towards Carbon in terms of theming, not implementing the exact component styles.

**Todo**
* [x] Gather feedback
* [x] Adjust changes from recent work stream (e.g. flexible forms)
* [x] Consider: Use carbonish icons in form-js number and select directly
* [x] Fixup wrong input border margins
* [x] Remove unnecessary styles (because they are handled by form-js already)
* [ ] Dry run to see whether something is broken: `npm run start:carbon`

**Theming**

image

image

**Carbon styles**

image

PR #561, opened 2023-03-07
Bundle style distros

pinussilvestrus:
Note: I separated the style bundling into another PR so it's
a) easier to review
b) can be re-used for other topics (e.g. [flexible layouts](https://github.com/bpmn-io/form-js/pull/553))

---
Closes https://github.com/bpmn-io/form-js/issues/257
Related to https://github.com/bpmn-io/form-js/pull/557

This includes
* Introducing `scss` bundling to a single style export per package
* Export `base` style sheets in case they are needed
  * Use case: Tasklist imports the carbonized flatpickr styles, so it does not need the bundled version

BREAKING CHANGES:
* `@bpmn-io/form-js/dist/assets/dragula.css` got removed
* `@bpmn-io/form-js/dist/assets/flatpickr/light.css` got removed
* `@bpmn-io/form-js/dist/assets/properties-panel.css` got removed


packages/form-js-editor/package.json
```diff
@@ -20,9 +19,12 @@
   "types": "dist/types/index.d.ts",
   "scripts": {
     "all": "run-s lint test build",
-    "build": "run-p bundle generate-types",
+    "build": "run-p bundle bundle:scss generate-types",
     "bundle": "rollup -c --failAfterWarnings --bundleConfigAsCjs",
-    "bundle:watch": "rollup -c -w --bundleConfigAsCjs",
+    "bundle:scss": "sass --no-source-map --load-path=\"../../node_modules\" assets/index.scss dist/assets/form-js-editor.css",
```
pinussilvestrus, line 24: I decided to use this simple bundling step instead of a rollup plugin. Rationale: the rollup plugins I used (https://github.com/thgh/rollup-plugin-scss, https://github.com/egoist/rollup-plugin-postcss) didn't work as expected.

Skaiir (approved):
Looks amazing, this really saves a us a lot of maintenance in the future as well which is great :)

pinussilvestrus:
Thanks for your feedback @Skaiir! 🎉  Waiting for @vsgoulart to give his opinion before merging this.

vsgoulart (approved):
LGTM

The old CSS files don't exist anymore, right? If that's right we have to communicate to them otherwise any project that's using `form-js` would break

pinussilvestrus:
> LGTM
> 
> The old CSS files don't exist anymore, right? If that's right we have to communicate to them otherwise any project that's using `form-js` would break

That's correct, I already added the breaking change note to this PR + commits and we will add it to the Changelog as well. [Let's consider also a blogpost](https://github.com/camunda/team-hto/issues/181) + update the form-js-examples. I'll track this in the epic issue.

PR #563, opened 2023-03-14
fix(editor): use proper empty value for select defaults

### LINKED ISSUE #562: Select Element defaults to string "<none>" when selecting <none> as default
### Describe the Bug

it is possible to select default value <none> which will result in the variable value "<none>". 

### Steps to Reproduce

1. Input:
`{}`

2. Drag and Drop Select Element

image

3. Observe Form Output

```
{
  "field_0mzhnhz": "",
  "field_16exx0r": null
}
```

4. Set Default to <none>

image

5. Resulting Schema:
```
{
  "components": [
    {
      "values": [
        {
          "label": "First",
          "value": "Value"
        }
      ],
      "label": "Select",
      "type": "select",
      "id": "Field_0jqm2fz",
      "key": "select",
      "defaultValue": "<none>"
    },
    {
      "label": "Text field",
      "type": "textfield",
      "id": "Field_1igdsio",
      "key": "text",
      "appearance": {}
    }
  ],
  "schemaVersion": 7,
  "exporter": {
    "name": "Camunda Web Modeler",
    "version": "06b1b8a"
  },
  "type": "default",
  "id": "DefaultValues",
  "executionPlatform": "Camunda Cloud",
  "executionPlatformVersion": "8.1.0"
}
```
6. See Form Output
image

```
{
  "field_0mzhnhz": "",
  "field_16exx0r": "<none>"
}
```

### Expected Behavior

I expect to result in null or an empty string.

### Environment

- Host (Browser/Node version), if applicable: Chrome Version 110.0.5481.180
- OS: [e.g. Windows 10]
- Library version: Via Web Modeler

pinussilvestrus:
Closes #562

Also preparing the changelog so we can follow up with a patch release.

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/DefaultValueEntry.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/DefaultValueEntry.js
@@ -15,6 +15,8 @@ import { useService } from '../hooks';
 
 import { countDecimals, INPUTS, isValidNumber, VALUES_INPUTS } from '../Util';
 
+export const EMPTY_OPTION = null;
+
 export default function DefaultOptionEntry(props) {
   const {
     editField,
@@ -193,7 +195,7 @@ function DefaultValueSingleSelect(props) {
   } = props;
 
   const {
-    defaultValue,
+    defaultValue = EMPTY_OPTION,
     values = []
   } = field;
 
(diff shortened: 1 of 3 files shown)


vsgoulart (commented):
Maybe I misunderstood something, but I'm still getting `null`
Shouldn't I get an empty string?

image

pinussilvestrus:
I was not part of the initial implementation, but it is indeed a wanted behavior that we set the value to `null` when `<none>` was selected. I didn't touch that. 

To make it more clearer, I can simply set the `EMPTY_OPTION` to null as well (shouldn't make a difference in the end).

vsgoulart (approved):

PR #567, opened 2023-03-21
Feelers integration

Skaiir:
Related to https://github.com/camunda/team-hto/issues/141


package.json
```diff
@@ -104,5 +104,11 @@
     "source-map-loader": "^4.0.0",
     "typescript": "^5.0.0",
     "webpack": "^5.40.0"
+  },
+  "dependencies": {
+    "feelers": "^0.1.0-alpha.4",
+    "preact-portal": "^1.1.3",
+    "react-draggable": "^4.4.5",
+    "react-resizable": "^3.0.4"
```
pinussilvestrus, line 112: Is any of this dependency needed, still? `feelers` should be installed from the properties panel and I think the other one is absolute without the popup.

packages/form-js-editor/src/FormEditor.js
```diff
@@ -10,6 +10,9 @@ import ModelingModule from './features/modeling';
 import SelectionModule from './features/selection';
 import PaletteModule from './features/palette';
 import PropertiesPanelModule from './features/properties-panel';
+import ExpressionLanguageModule from './features/expression-language';
+
+import { MarkdownModule } from '@bpmn-io/form-js-viewer';
```
pinussilvestrus, line 15: Awesome stuff 💌

packages/form-js-editor/src/features/properties-panel/entries/TextEntry.js
```diff
@@ -2,32 +2,47 @@ import { get } from 'min-dash';
...
       component: Text,
       editField: editField,
       field: field,
-      isEdited: isFeelEntryEdited
+      isEdited: isFeelTemplatingEntryEdited
     }
   ];
+
+  if (templating.isTemplate(field.text)) {
+    entries.push(simpleBoolEntryFactory({
```
pinussilvestrus, line 37: I think a description would be helpful here, as I think it's hard to gather what "strict templating" means.

package.json
Skaiir: Nicely spotted, forgot to get rid

packages/form-js-viewer/src/features/expression-language/ConditionChecker.js
```diff
@@ -81,33 +81,6 @@ export class ConditionChecker {
...
 
-  /**
-   * Evaluate an expression.
-   *
-   * @param {string} expression
-   * @param {import('../types').Data} [data]
-   *
-   * @returns {any}
-   */
-  evaluate(expression, data = {}) {
```
pinussilvestrus, line 92: Is this technical a breaking change? as `form.get('conditionChecker').evaluate(expression)` would break now.

We could mitigate this by keeping the function, but executing `expressionLanguage.evaluate` under the hood. We could also say it was never public API. Wdyt?

packages/form-js-viewer/src/render/hooks/useFilteredFormData.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Returns the conditionally filtered data of a form reactively.
+ * Memoised to minimize re-renders
+ *
+ */
+export default function useFilteredFormData() {
+  const { initialData, data } = useService('form')._getState();
+  const conditionChecker = useService('conditionChecker', false);
+
+  return useMemo(() => {
```
pinussilvestrus, line 13: 👍

packages/form-js-viewer/test/spec/render/components/form-fields/Text.spec.js
```diff
@@ -82,23 +83,27 @@ Some _em_ **strong** [text](#text) \`code\`.
...
-<ol><li>ol li 1</li><li>ol li 2</li></ol>
-
-<p></p><pre class="code"><code>Some Code</code></pre><p></p>
-
+    expect(formField.innerHTML).to.eql(`<div class="markup"><div xmlns="http://www.w3.org/1999/xhtml"><h1 id="h1">h1</h1>
+<h2 id="h2">h2</h2>
+<h3 id="h3">h3</h3>
+<h4 id="h4">h4</h4>
+<h5 id="h5">h5</h5>
+<h6 id="h6">h6</h6>
```
pinussilvestrus, line 91: I know the changes are really small due to the transition to `showdown`, but I think we should at least mention there are some changes in the resulting HTML, e.g. via note in
* the commit message
* the changelog 

Just some note on that we change the markdown parser to `showdown` and a quick reference to the specification (if there is some). I think that would help a lot already.

pinussilvestrus: Just mentioning it as it seems our users are interested in that (https://camunda-platform.slack.com/archives/C6WGNHV2A/p1678953005263759).

packages/form-js-editor/test/spec/render/components/form-fields/helper/index.js
```diff
@@ -1,9 +1,29 @@
+import { FormEditorContext } from '../../../../../../src/render/context';
 import { FormContext } from '@bpmn-io/form-js-viewer';
+import showdown from 'showdown';
+
+class MarkdownRenderer {
```
pinussilvestrus, line 5: I guess we can also use the Markdown Renderer from the viewer here, I just added one so we are covered (shouldn't make a big difference, it should work anyway).

packages/form-js-editor/src/features/properties-panel/entries/TextEntry.js
Skaiir: As discussed, we'll wait for feedback on this one as it is kind of hard to give descriptions for, most likely we'll cover this in the documentation and maybe tooltips.

packages/form-js-viewer/src/features/expression-language/ConditionChecker.js
Skaiir: Well, I doubt it'll affect many people since we don't have a documented public API, but yeah I guess technically a breaking change. But I also think we can say that we never 'officially' supported it. Then again do we ever officially support anything :trollface: . Might be worth having some kind of a 'contract', maybe API documentation. To know what we can and can't change in the future. Well whenever time permits.

packages/form-js-viewer/test/spec/render/components/form-fields/Text.spec.js
Skaiir: Yeah I'll try and break down the commits to make this explicit.

packages/form-js-editor/test/spec/render/components/form-fields/helper/index.js
Skaiir: Since showdown uses some global configs I don't think it's all that bad to have it here, so we can control the details better :)

pinussilvestrus (approved):
🚀 🎸

PR #579, opened 2023-03-29
ci(POST_RELEASE): use Node@16

nikku:
We'd otherwise [generate old style lock files](https://github.com/bpmn-io/bpmn-io-demo/commit/a4d2a5edec8506498a74608fc947dc00d2f6f1af) when upgrading the demo.


pinussilvestrus:
Thanks for the hint @nikku! Really appreciating this

pinussilvestrus (approved):

PR #583, opened 2023-03-29
fix(viewer): added missing feelers deps

### LINKED ISSUE #581: `@bpmn-io/form-js-viewer` doesn't have all dependecies necessary to work
### Describe the Bug

Since #567 `@bpmn-io/form-js-viewer` doesn't work without installing `feelers`

### Steps to Reproduce

1. Install `@bpmn-io/form-js-viewer`
2. Try to build your app (wit Create React App for example)
3. You get the following error message: `Module not found: Error: Can't resolve 'feelers' in '/home/user/repos/tasklist/client/node_modules/@bpmn-io/form-js-viewer/dist'`

### Expected Behavior

`@bpmn-io/form-js-viewer` should have all it's dependencies automatically installed or `feelers` should be considered a peer dependency

Skaiir:
Closes #581


vsgoulart (approved):

packages/form-js-viewer/package.json
```diff
@@ -46,6 +46,7 @@
     "big.js": "^6.2.1",
     "classnames": "^2.3.1",
     "didi": "^9.0.0",
+    "feelers": "^0.1.0-alpha.8",
```
pinussilvestrus, line 49: If we don't use a peer dependency here, we should add it to the roll-up externals, so it won't be bundled in the distros.

pinussilvestrus: Just added it.

PR #588, opened 2023-03-30
Integrate `@bpmn-io/draggle`

pinussilvestrus:
This uses an updated `dragula` version to make the `direction` dynamic. It detects whether the drop would happen inside a row or not and change the direction accordingly.

This makes drag & drop quite a bit smoother as it uses the correct axis depending where the cursor is located. Beforehand it would always have used the y-axis. 

Depends on https://github.com/bpmn-io/dragula/pull/1

Demo: https://demo-improve-dragging--camunda-form-playground.netlify.app/

Kapture 2023-03-30 at 13 59 19

PR #591, opened 2023-03-31
fix: recentered form view

### LINKED ISSUE #582: Horizontally center the form in the editor view
### What should we do?

Horizontally center the form in the editor view.

### Why should we do it?

Since the row drag handles only appear on hover, the form looks "off" in the edit view (different margins).

image

Cf feedback here:
* https://github.com/camunda/camunda-modeler/pull/3523#issuecomment-1488285775

Skaiir:
Closes #582


Skaiir:
**I've had to skip one of the tests to work.**  Generally, I think this test will give us problems in the future. @pinussilvestrus I think I'd leave the decision to you on how to handle this one. IMO it's just a really annoying behavior to test, because it's never really clear when those UI frameworks update the DOM, which is even harder to handle considering dragging shuffles everything around. Adding 100 pixels to the x of the drag fixes the test locally for me, but not on ci.  

Overall my fix simply takes the drag handle out of the placement hierarchy. I've also removed an override we were doing on the basic carbon grid component which was the cause of this pretty large outer band. Instead, we should configure the padding outside of the carbon grid components if we really want more spacing than carbon provides (which I've done here, added 4px of padding on each side. 

The active area of the drag handle can be modified as you guys see fit, @christian-konrad @RomanKostka please view the deployment and let me know if the feel is right. If not, I can always increase the area where the drag handle is active. I can also increase the padding, as my 'fix' definitely reduced the space on each side. 

I notice that on the deployment, the handle is not quite centered, this is something we'd have to adjust on the playground styles. But assume it was during testing. 

It looks like this on form-js proper:

image

Skaiir:
Here is a visual indicator of the active drag area:

image

Skaiir:
I notice that the css change made the dragging in of new components a little more finicky, I'll investigate.

packages/form-js-editor/test/spec/FormEditor.spec.js
```diff
@@ -992,7 +992,7 @@ describe('FormEditor', function() {
     });
 
 
-    it('should move row', async function() {
+    it.skip('should move row', async function() {
```
pinussilvestrus, line 995: This becomes laggy due to the absolute positioning now, I think. That's always hard to test, so I got your rationale about this particular test.

In general, I think the dragging tests make sense, as the behavior is predictable.

pinussilvestrus: Can we at least add an comment above the test why we skip it and maybe a note that we could solve it with proper integration testing? Or should we remove it entirely?

packages/form-js-editor/assets/form-js-editor-base.css
```diff
@@ -404,11 +409,7 @@
 /* row drop preview */
 .fjs-children > .gu-transit {
   height: 28px !important;
-  margin-left: 28px !important;
-  margin-right: -8px !important;
-  flex: unset !important;
```
pinussilvestrus, line 409: I believe we have to keep `flex: unset`, otherwise it will mess up the height of the previews.

image

packages/form-js-editor/test/spec/FormEditor.spec.js
vsgoulart: We can try to move this test o Playwright when I create the setup

RomanKostka (changes requested):
looks good.
only the issue with the broken preview we should take a look at.
https://github.com/bpmn-io/form-js/pull/591#discussion_r1155576847

pinussilvestrus:
> recentering looks good 🎉 unfortunately i found some other weird behavior: the drop is only possible at certain areas

@RomanKostka @Skaiir this is a known bug due to the current way dragula calculates the drop position. This will be fixed via #588 or once we migrate to `dragula@2`.

PR #593, opened 2023-04-03
Resize form fields

### LINKED ISSUE #566: Implement form field resizing on the grid
### What should we do?

It should be possible to resize a form field along the grid.

[Sketches](https://www.figma.com/file/CiTSN1vNETjIAy6W9jUIFp/Form-Editor?node-id=812-130301&t=ZoEPtRoSYq3sp8tn-0)

image

Follow up of #560 

### Why should we do it?

Child of https://github.com/camunda/team-hto/issues/50

### LINKED ISSUE #572: Redesign of delete field icon
### What should we do?

We should redesign the delete field icon. Sketches: https://www.figma.com/file/CiTSN1vNETjIAy6W9jUIFp/Form-Editor?node-id=851%3A133246&t=gaz84Ucgk2SC1AX8-1

### Why should we do it?

Context: https://github.com/bpmn-io/form-js/pull/553#issuecomment-1483133111

### LINKED ISSUE #605: Add odd options for columns
### What should we do?

Enable odd column sizes in the editor ([properties panel](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/src/features/properties-panel/entries/ColumnsEntry.js#L67), [resizer](https://github.com/bpmn-io/form-js/pull/593)). 

What we should discuss: is a dropdown still the right choice here?

### Why should we do it?

In the properties panel and resizer, we only allow for even columns in a form field. [In recent user research](https://docs.google.com/document/d/10V4hitzV3J_OxVvlu7Tk2notSH9B4ZrgGGf49cq3g6g/edit#heading=h.3s5h2l7t286t), we found out that this does not help, but indeed prevents users from creating pleasing layouts.

pinussilvestrus:
Closes #566 
Closes #572 
Closes #605

Demo: https://demo-566-form-field-resize--camunda-form-playground.netlify.app/

Kapture 2023-04-13 at 13 38 27


christian-konrad (approved):
Found some minor issues (as mentioned in Slack), but they aren't related to the resize handles.
Really like the smooth handling of the resize handles!

packages/form-js-editor/src/render/components/Util.js
```diff
@@ -8,4 +8,105 @@ export function editorFormFieldClasses(type, { disabled = false } = {}) {
...
+ * }
+ *
+ * domElement.addEventListener('dragstart', dragger(dragMove));
+ *
+ * @param {Function} fn
+ * @param {Element} dragPreview
+ *
+ * @return {Function} drag start callback function
+ */
+export function createDragger(fn, dragPreview) {
```
pinussilvestrus, line 30: For reference: I had a look at some existing resizable component solutions (e.g., https://github.com/react-grid-layout/react-resizable, https://www.npmjs.com/package/preact-moveable), but none of them gave enough flexibility to achieve what we wanted; especially due to the fact we resize in columns, not pixels.

I decided to re-use proven utilities from bpmn-io to implement the behaviors.

RomanKostka (approved):
love it!
one small thing i noticed.

the resize handles have different heights for some components:
image
image

these should always have the same height, vertically centered

pinussilvestrus:
@RomanKostka I adjusted the resize handle heights to be always 50px and vertically aligned, except for the Checkbox (28px).

image
image
image

This can lead to situations where radios/checklists can have less height when no options are defined. Personally, I wouldn't start applying custom styles for such empty states for certain multi-select elements. Wdyt?

image

christian-konrad:
@pinussilvestrus for the handle height, can't we add a max-height of like `calc(100% - margin)` of the container size?

pinussilvestrus:
> @pinussilvestrus for the handle height, can't we add a max-height of like `calc(100% - margin)` of the container size?

Oh yeah, damn right 😅 I thought it too complicated. Just pushed it, demo should be up in a few minutes 👍

vsgoulart (commented):
@pinussilvestrus 2 things I found

1. Maybe we should disable the handle if an element is the first column of the row?

https://user-images.githubusercontent.com/1022916/232054396-cdf1a70b-321e-4e64-8c54-5d0a3609f7f7.mp4





2. Should we maybe prevent

https://user-images.githubusercontent.com/1022916/232054666-7a540c31-06b2-459b-95f1-2e4777306013.mp4

 user from breaking into a new row?

pinussilvestrus:
> Maybe we should disable the handle if an element is the first column of the row?

Not sure about this one, it could be confusing for users if that's missing for just this case. wdyt @RomanKostka?

> Should we maybe prevent user from breaking into a new row?

That should not happen, but I also see this behavior outside of resizing, also cf. to the findings Christian made with auto columns. I didn't see it with resizing yet. Can you maybe share your setup? So the columns of each form field and maybe whether you are on a smaller screen?

packages/form-js-editor/src/render/components/FormEditor.js
```diff
@@ -136,6 +140,8 @@ function Element(props) {
         }
       </ContextPad>
       { props.children }
+      <FieldResizer position="left" field={ field }></FieldResizer>
```
Skaiir, line 143: Wonder if we can somehow conditionally display this one based on position, to not display in cases where we don't want it.

Skaiir: firefox_GtTQpbMboA

Lack of auto does make the usage a little strange I would say, though I'm not so sure what the behaviour we want is.

Skaiir (commented):
Looks great man 👍, just a few small behavior things that might be worth looking at, but otherwise what a contribution.

packages/form-js-editor/assets/form-js-editor-base.css
```diff
@@ -145,6 +148,60 @@
...
+  right: -4px;
+  cursor: ew-resize;
+}
+
+.fjs-editor-container .fjs-field-resize-handle-left {
+  left: -4px;
+  cursor: ew-resize;
+}
+
+.fjs-editor-container .fjs-form-field-checkbox + .fjs-field-resize-handle,
```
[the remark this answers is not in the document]
Skaiir: Oh, I missed that, already fixed it seems.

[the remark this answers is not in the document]
pinussilvestrus: 👍

packages/form-js-editor/src/render/components/FieldResizer.js
```diff
@@ -0,0 +1,184 @@
...
+
+
+// helper //////
+
+function asColumns(width, parent) {
+  const totalWidth = parent.getBoundingClientRect().width;
+
+  const oneColumn = (1 / 16) * totalWidth;
+
+  return Math.round(width / oneColumn);
```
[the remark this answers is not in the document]
pinussilvestrus: That's true, but we anyway [permit uneven columns via the `FormLayoutValidator`](https://github.com/bpmn-io/form-js/pull/593/files#diff-2844e6cf69aed3c32393e058a5cf19ff3b2414c1961caa1738fb5eace10ff57aR29), that's why a check for even numbers is not needed here. I wanted to keep the helper less restrictive, so I left it out here. 

In the end, we should have the expected behavior of even grid steps.

packages/form-js-editor/src/render/components/FormEditor.js
pinussilvestrus: That was mentioned via https://github.com/bpmn-io/form-js/pull/593#pullrequestreview-1385447889 already (for one case). I agree that the resizers do not make super sense for each case, but leaving them out because _we_ believe they are not useful is maybe not the best solution.

Probably a good indicator for waiting on feedback /cc @RomanKostka

packages/form-js-editor/assets/form-js-editor-base.css
```diff
@@ -145,6 +148,60 @@
...
+  bottom: 9px;
+  border: 1px solid var(--color-resize-handle-border);
+  border-radius: 50px;
+  background: var(--color-white);
+  z-index: 3;
+}
+
+.fjs-editor-container .fjs-field-resize-handle-right {
+  right: -4px;
+  cursor: ew-resize;
```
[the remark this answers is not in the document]
pinussilvestrus: Seems like Firefox has some problems with the (empty) drag preview image. I'll check whether we can do something against this.

Bildschirm­foto 2023-04-17 um 11 12 20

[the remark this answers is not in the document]
pinussilvestrus: I fixed that via https://github.com/bpmn-io/form-js/pull/593/commits/59d464d0aa79ad46ad2351e3b045513c209f97e9 by using a `div` instead of a `canvas` element. Apparently, firefox has its problems with `canvas` 😄

packages/form-js-editor/src/render/components/FieldResizer.js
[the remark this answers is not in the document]
Skaiir: I know we don't permit uneven columns, but this isn't the problem I'm referring to. 

If you look carefully in the below example, you can see that the rounding causes the dragging to not be 'centered', it starts a little bit to the left (1/4 of 2 rows), and ends quite a bit to the right (3/4 of 2 rows). This is because the maths doesn't work out here because you're essentially still rounding single columns but using that result to calculate positions in a 2 column system. 

firefox_oFwPqYezIc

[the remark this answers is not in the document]
Skaiir: image

pinussilvestrus:
To consider as an improvement on top: https://github.com/bpmn-io/form-js/pull/604. It depends on whether we introduce [single step columns](https://github.com/bpmn-io/form-js/issues/605) or not.

packages/form-js-editor/src/render/components/FieldResizer.js
```diff
@@ -0,0 +1,188 @@
...
+
+  return (
+    <div
+      ref={ ref }
+      class={ classNames(
+        'fjs-field-resize-handle',
+        'fjs-field-resize-handle-' + position,
+        DRAG_NO_MOVE_CLS
+      ) }
+      draggable={ true }
```
vsgoulart, line 131: Nitpick, the value can be omitted

```suggestion
      draggable
```

packages/form-js-editor/src/render/components/FieldResizer.js
```diff
@@ -0,0 +1,188 @@
...
+  const ref = useRef(null);
+
+  const formLayoutValidator = useService('formLayoutValidator');
+  const modeling = useService('modeling');
+
+  // we can't use state as we need to
+  // manipulate this inside dragging events
+  const context = useRef({
+    startColumns: null,
+    newColumns: null
```
vsgoulart, line 43: Maybe we should use 0 instead of `null`? This way we keep the types as just `number` for this

packages/form-js-editor/src/render/components/FieldResizer.js
```diff
@@ -0,0 +1,188 @@
...
+    setResizing(target, position);
+  };
+
+  const onResizeEnd = () => {
+    const { layout = {} } = field;
+
+    if (context.current.newColumns) {
+      modeling.editFormField(field, 'layout', {
+        ...layout,
+        columns: parseInt(context.current.newColumns)
```
vsgoulart, line 105: is `parseInt` really necessary? it seems that we always only deal with numbers with the context.

vsgoulart (approved):
Implementation looks good (just made some minor comments), let's just wait to hear Roman's comment about the points I raised earlier

packages/form-js-editor/src/render/components/FieldResizer.js
pinussilvestrus: In this case I think you are right, likely no strings ever involved.

RomanKostka:
> Not sure about this one, it could be confusing for users if that's missing for just this case. wdyt @RomanKostka?

I would propose to leave it as it is and keep the handles. Feels natural at least for me - this is something we can investigate further when we talk to some users

One other thing to mention to keep track of it:
we have to find a solution that something like this (4 components with col-span 2 in a small window):
image
is not happening too often. One solution might be to increase the threshold of the breakpoint.

pinussilvestrus:
> One other thing to mention to keep track of it:
> we have to find a solution that something like this

It's a thing I spent an eye for quite a while now. The thing is that Carbon calculates the breakpoints on the screen basis, not container basis, as container based media queries (CSS) are not well supported across browsers. So it doesn't matter whether you are in a single form context, or have setups like the playground were a lot is going on. Increasing the breakpoint might solve it for quite some cases, but probably not every.

Definitely a thing we should keep tracking, and act on once we need :+1:

Issue #323, opened 2022-09-02
Eagerly validate form input


nikku:
__Is your feature request related to a problem? Please describe__

A form can be fed with input, and that input may be broken. Unfortunately we don't eagerly validate such broken input, i.e. `aaa` for a field that requires `C-\d+` pattern validation. 

The result is that a user _MUST_ change the field or submit the entire form to learn about this error:

![capture TcbhKl_optimized](https://user-images.githubusercontent.com/58601/188123508-7ec1309f-9d0f-4a87-895b-928018515d67.gif)

__Describe the solution you'd like__

* [ ] We should treat invalid external input same as user input and eagerly validate the form.

__Describe alternatives you've considered__

None.

__Additional context__

Suggested by @christian-konrad [here](https://github.com/bpmn-io/internal-docs/issues/617#issuecomment-1235304770).

griimick:
I have been going through the `Form` implementation. Is this one open for contribution?

Skaiir:
This and https://github.com/bpmn-io/form-js/issues/382 need to be considered together.

PR #595, opened 2023-04-04
feat: implemented `required` where it was missing

### LINKED ISSUE #594: Support required attribute everywhere it makes sense

Mirror issue for: https://github.com/camunda/camunda-modeler/issues/3529

We should support:
- checklist
- taglist
- checkbox

All these components should be made require-able

In the context of a checkbox, that means 'must be checked'
In the context of the lists (multiselect), that means 'must have at least one selected'

-

Skaiir:
Closes #594

### DIFF
--- a/packages/form-js-viewer/src/core/Validator.js
+++ b/packages/form-js-viewer/src/core/Validator.js
@@ -69,11 +69,17 @@ export default class Validator {
       ];
     }
 
-    if (validate.required && (isNil(value) || value === '')) {
-      errors = [
-        ...errors,
-        'Field is required.'
-      ];
+    if (validate.required) {
+      const isUncheckedCheckbox = type === 'checkbox' && value === false;
+      const isUnsetValue = isNil(value) || value === '';
+      const isEmptyMultiselect = Array.isArray(value) && value.length === 0;
+
+      if (isUncheckedCheckbox || isUnsetValue || isEmptyMultiselect) {
+        errors = [
+          ...errors,
+          'Field is required.'
+        ];
+      }
     }
 
     if ('min' in validate && (value || value === 0) && value < validate.min) {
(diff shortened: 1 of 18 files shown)


packages/form-js-viewer/src/core/Validator.js
```diff
@@ -69,11 +69,17 @@ export default class Validator {
       ];
     }
 
-    if (validate.required && (isNil(value) || value === '')) {
-      errors = [
-        ...errors,
-        'Field is required.'
-      ];
+    if (validate.required) {
```
pinussilvestrus, line 72: The required radio does not work properly.

image

1. Add a radio
2. Specify as required
3. No error is shown although no value is given

I think this is similar to other fields, [required is only triggering once the value got erased](https://github.com/bpmn-io/form-js/issues/323). That's not possible for the radio though 😅 So it would only show up on submit, so I think we are fine in that place.

pinussilvestrus: Maybe it makes sense to add tests to the `Validator.spec.js#required` to verify the new cases (checkbox, checklist, radio, taglist, datetime)?

pinussilvestrus (commented):
Works as expected 👍 Just one comment on some test cases in the `Validator.spec.js`.

packages/form-js-viewer/src/core/Validator.js
Skaiir: This is expected behavior, we shouldn't trigger the validation straight away as that would lead to prompting a user with errors before they have done anything.

We should validate on losing focus however, which we currently do not do, but that's an old thing that I'll tackle soon.

The proper way to test here is by supplying a submit button.

pinussilvestrus (approved):
I added two more test cases for the Validator. Feel free to merge 👍

PR #597, opened 2023-04-05
fix: compare labels properly in unique validation

### LINKED ISSUE #552: Setting a label to the same value as a value triggers 'must be unique validation'
### Describe the Bug

When setting a label in a statically defined value, a must be unique validation is triggered. This validation shouldn't check against values, as we should be able to define values which have a label equal to their values. Perhaps we should even allow it to be empty and use the value as default. 

image

### Steps to Reproduce

1. Select any select-like component
2. Set options source to static if not already the case
3. Create a new option, set its value part to (myValue), and then try and set the label to the same string
4. The validation error is shown

### Expected Behavior

No validation error should be thrown

Skaiir:
Closes #552

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/StaticValuesSourceEntry.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/StaticValuesSourceEntry.js
@@ -29,7 +29,7 @@ export default function StaticValuesSourceEntry(props) {
     editField(field, VALUES_SOURCES_PATHS[VALUES_SOURCES.STATIC], without(values, entry));
   };
 
-  const validateFactory = (key) => {
+  const validateFactory = (key, getValue) => {
     return (value) => {
       if (value === key) {
         return;
@@ -39,7 +39,7 @@ export default function StaticValuesSourceEntry(props) {
         return 'Must not be empty.';
       }
 
-      const isValueAssigned = values.find(entry => entry.value === value);
+      const isValueAssigned = values.find(entry => getValue(entry) === value);
 
       if (isValueAssigned) {
         return 'Must be unique.';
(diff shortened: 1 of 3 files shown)


packages/form-js-editor/src/features/properties-panel/entries/StaticValuesSourceEntry.js
```diff
@@ -29,7 +29,7 @@ export default function StaticValuesSourceEntry(props) {
     editField(field, VALUES_SOURCES_PATHS[VALUES_SOURCES.STATIC], without(values, entry));
   };
 
-  const validateFactory = (key) => {
+  const validateFactory = (key, map) => {
```
vsgoulart, line 32: I think it's a bit more clear

```suggestion
  const validateFactory = (key, getValue) => {
```

vsgoulart (changes requested):
@Skaiir I think you need to rebase you branch, there's an extra commit from the 0.14.0 release

packages/form-js-editor/src/features/properties-panel/entries/ValueEntry.js
```diff
@@ -99,6 +99,6 @@ function Value(props) {
     id,
     label: 'Value',
     setValue,
-    validate: validateFactory(getValue())
+    validate: validateFactory(getValue(), (entry) => entry.value)
```
pinussilvestrus, line 102: I know it's existing behavior, but with this validation, our users can't specify an empty value option. I'm just thinking that might be a valid use case, actually 🤔

pinussilvestrus (approved):
Works for me!

packages/form-js-editor/src/features/properties-panel/entries/ValueEntry.js
Skaiir: Maybe if it comes up as a request, we can look into it. But I think for now let's ignore it's too minor to care IMO.

PR #600, opened 2023-04-13
style: properly align rows in viewer

pinussilvestrus:
We do not use the Wide grid mode (https://carbondesignsystem.com/guidelines/2x-grid/implementation#the-wide-grid), so a slight negative margin is added to both sides. We want to prevent that when embedding the viewer inside applications.

We still apply the margin to the playground to save space. 

Related to https://github.com/camunda/tasklist/pull/2746

### DIFF
--- a/packages/form-js-playground/src/components/PlaygroundRoot.css
+++ b/packages/form-js-playground/src/components/PlaygroundRoot.css
@@ -160,6 +160,11 @@
   overflow: hidden;
 }
 
+.fjs-pgl-form-container .fjs-container .cds--grid .cds--row {
+  margin-left: -1rem;
+  margin-right: -1rem;
+}
+
 .fjs-pgl-form-container .fjs-form-editor {
   width: 100%;
 }
--- a/packages/form-js-viewer/assets/form-js-base.css
+++ b/packages/form-js-viewer/assets/form-js-base.css
@@ -113,6 +113,8 @@
 
 .fjs-container .cds--grid .cds--row {
   align-items: start;
+  margin-left: 0;
+  margin-right: 0;
 }
 
 @media (max-width: 66rem) {


pinussilvestrus:
@vsgoulart I didn't found a configuration option except from using the wide mode, which we don't want.

vsgoulart (commented):
> @vsgoulart I didn't find a configuration option except using the wide mode, which we don't want.

What does wide mode break? I think if we we should avoid overriding Carbon if we can

pinussilvestrus:
The wide mode adds 32px gutters between each form field (column).

image

That leads to a lot of space wasted between form fields

Kapture 2023-04-13 at 11 02 01

The condensed mode reduced this to a minimum 1px, which is the one we wanted to use (/cc @RomanKostka). I don't read something in the docs about why Carbon is adding the -1rem margins in the narrow and condensed mode, though.

vsgoulart (approved):

RomanKostka:
> The wide mode adds 32px gutters between each form field (column).
> 
> image
> 
> That leads to a lot of space wasted between form fields
> 
> Kapture 2023-04-13 at 11 02 01 [ Kapture 2023-04-13 at 11 02 01 ](https://user-images.githubusercontent.com/9433996/231710488-1be5cb26-3061-4ad2-9e9a-106b42f01ca7.gif) [ ](https://user-images.githubusercontent.com/9433996/231710488-1be5cb26-3061-4ad2-9e9a-106b42f01ca7.gif)
> 
> The condensed mode reduced this to a minimum 1px, which is the one we wanted to use (/cc @RomanKostka). I don't read something in the docs about why Carbon is adding the -1rem margins in the narrow and condensed mode, though.

not sure why they do it to be honest.
Basically we can use whatever grid we want and best suits our need.

@pinussilvestrus one question regarding the 'narrow' grid: are the 16px gutter divided into 8px per side and added as padding to the form components?

pinussilvestrus:
> @pinussilvestrus one question regarding the 'narrow' grid: are the 16px gutter divided into 8px per side and added as padding to the form components?

From the docs, it looks like 16px (1rem) are added as right padding to each form field, so it always has 16px between two form field. That's also how it works when applying `cds--grid-narrow`

image

PR #603, opened 2023-04-14
fix: getSchemaVariable accounts for feelers expressions

Skaiir:
My condolences to whoever reviews this. Send a message if you need assistance, but this is the kind of scenario where writing readable code is very challenging.

### DIFF
--- a/packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
+++ b/packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
@@ -1,11 +1,49 @@
-import { evaluate as evaluateFeelers } from 'feelers';
+import { parser as feelersParser, buildSimpleTree, evaluate as evaluateFeelers } from 'feelers';
 import { isString } from 'min-dash';
+import { getFlavouredFeelVariableNames } from './variableExtractionHelpers';
 
 export default class FeelersTemplating {
 
   constructor() { }
 
-  isTemplate(value) { return isString(value) && (value.startsWith('=') || /{{/.test(value)); }
+  /**
+   * Determines if the given value is a feelers template.
+   *
+   * @param {any} value
+   * @returns {boolean}
+   *
+   */
+  isTemplate(value) { return isString(value) && (value.startsWith('=') || /{{.*?}}/.test(value)); }
+
+
+  /**
+   * Retrieve variable names from a given feelers template.
+   *
+   * @param {string} template
+   *
+   * @returns {string[]}
+   */
+  getVariableNames(template) {
+
+    if (!this.isTemplate(template)) {
+      return [];
+    }
+
+    const expressions = this._extractExpressionsWithDepth(template);
+
+    // defines special accessors, and the change(s) in depth they could imply (e.g. parent can be used to access the parent context (depth - 1) or a child variable named parent (depth + 1)
+    const specialDepthAccessors = {
+      parent: [ -1, 1 ],
+      _parent_: [ -1 ],
+      this: [ 0, 1 ],
+      _this_: [ 0 ],
+    };
+
+    return expressions.reduce((variables, { expression, depth }) => {
+      return variables.concat(getFlavouredFeelVariableNames(expression, 'expression', { depth, specialDepthAccessors }));
+    }, []);
+  }
+
 
   /**
    * Evaluate a template.
@@ -28,6 +66,53 @@ export default class FeelersTemplating {
     } = options;
 
     return evaluateFeelers(template, context, { debug, strict, buildDebugString });
+  }
+
+
+  /**
+ * @typedef {Object} ExpressionWithDepth
+ * @property {number} depth - The depth of the expression in the syntax tree.
+ * @property {string} expression - The extracted expression
+ */
+
+  /**
+ * Extracts all feel expressions in the template along with their depth in the syntax tree.
+ * The depth is incremented for child expressions of loops to account for context drilling.
+
+ * @name extractExpressionsWithDepth
+ * @param {string} template - A feelers template string.
+ * @returns {Array<ExpressionWithDepth>} An array of objects, each containing the depth and the extracted expression.
+ *
+ * @example
+ * const template = "Hello {{user}}, you have:{{#loop items}}\n- {{amount}} {{name}}{{/loop}}.";
+ * const extractedExpressions = _extractExpressionsWithDepth(template);
+ */
+  _extractExpressionsWithDepth(template) {
+
+    // build simplified feelers syntax tree
+    const parseTree = feelersParser.parse(template);
+    const tree = buildSimpleTree(parseTree, template);
+
+    return (function _traverse(n, depth = 0) {
+
+      if ([ 'Feel', 'FeelBlock' ].includes(n.name)) {
+        return [ { depth, expression: n.content } ];
+      }
+
+      if (n.name === 'LoopSpanner') {
+        const loopExpression = n.children[0].content;
+        const childResults = n.children.slice(1).reduce((acc, child) => {
+          return acc.concat(_traverse(child, depth + 1));
+        }, []);
+        return [ { depth, expression: loopExpression }, ...childResults ];
+      }
+
+      return n.children.reduce((acc, child) => {
+        return acc.concat(_traverse(child, depth));
+      }, []);
+
+    })(tree);
+
   }
 }
 
--- a/packages/form-js-viewer/src/features/expression-language/variableExtractionHelpers.js
+++ b/packages/form-js-viewer/src/features/expression-language/variableExtractionHelpers.js
@@ -1,0 +1,167 @@
+import { parseExpressions, parseUnaryTests } from 'feelin';
+
+export const getFlavouredFeelVariableNames = (feelString, feelFlavour, options = {}) => {
+
+  const {
+    depth = 0,
+    specialDepthAccessors = {}
+  } = options;
+
+  if (![ 'expression', 'unaryTest' ].includes(feelFlavour)) return [];
+
+  const tree = feelFlavour === 'expression' ? parseExpressions(feelString) : parseUnaryTests(feelString);
+
+  const simpleExpressionTree = _buildSimpleFeelStructureTree(tree, feelString);
+
+  return (function _unfoldVariables(node) {
+
+    if (node.name === 'PathExpression') {
+
+      if (Object.keys(specialDepthAccessors).length === 0) {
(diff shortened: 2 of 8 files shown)


Skaiir:
As part of this, I've also fixed a broken feel expression behavior which would query all properties as variables.

Skaiir:
@marstamm Added you as well, you can take a look if you like it's optional but I think this kind of problem is very up your alley, maybe I'm overengineering this.

packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
```diff
@@ -1,11 +1,47 @@
...
 
-  isTemplate(value) { return isString(value) && (value.startsWith('=') || /{{/.test(value)); }
+  /**
+   * Determines if the given value is a feelers template.
+   *
+   * @param {any} value
+   * @returns {boolean}
+   *
+   */
+  isTemplate(value) { return isString(value) && (value.startsWith('=') || /{{.*?}}/.test(value)); }
```
pinussilvestrus, line 16: Is it also a template, when it's an expression only?

packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
```diff
@@ -29,6 +65,37 @@ export default class FeelersTemplating {
 
     return evaluateFeelers(template, context, { debug, strict, buildDebugString });
   }
+
+
+  _extractExpressionsWithDepth(template) {
```
pinussilvestrus, line 70: What could help here is an example return via JSDocs to make it easier to understand what the expressions array looks like. Maybe even a `@type` declarator. 

Now that I read through the implementation it becomes clear, but that was my first impression on the start 😅

packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
```diff
@@ -29,6 +65,37 @@ export default class FeelersTemplating {
...
+
+    // build simplified feelers syntax tree
+    const parseTree = feelersParser.parse(template);
```
pinussilvestrus, line 73: Just for my understanding, the `feelersParser` can also parse `= foo` like expressions?

packages/form-js-viewer/src/features/expression-language/variableExtractionHelpers.js
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * @param {Object} node - The node to extract variable names from.
+ * @param {number} initialDepth - The initial depth to start searching for variables.
+ * @param {Object} specialDepthAccessors - An object containing special depth accessor mappings.
+ * @returns {Set} - A set containing the extracted variable names.
+ */
+const _smartExtractVariableNames = (node, initialDepth, specialDepthAccessors) => {
+
+  // double brackets are needed as set constructor unpacks the first layer of the array
+  let currentDepthInfos = new Set([ [ -999, initialDepth - 1 ] ]);
```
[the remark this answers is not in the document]
pinussilvestrus: Just as an idea: conduct a lightning talk with practical examples of this particular extraction ⚡

pinussilvestrus (commented):
My head is smoking 😅 I tried it out in the playground via the autocompletion and everything seems to work as expected 👍 

I wonder whether the whole variable extraction utilities could live under `feelers`, as we might need it elsewhere and it anyway uses internals of the templating language. It's good for now, though.

Good job 🚀

packages/form-js-viewer/test/spec/util/GetSchemaVariables.spec.js
```diff
@@ -43,6 +45,22 @@ describe('util/getSchemaVariables', () => {
...
+
+    expect(variables).to.eql([ 'myText', 'greeting', 'name', 'showAge', 'age', 'hobbies' ]);
+  });
+
+
+  it('should handle complex template cases', () => {
+
+    const variables = getSchemaVariables(complexTemplateSchema);
+
+    expect(variables).to.eql([ 'value', 'minimum', 'display', 'orgs', 'external1', 'external2', 'parent', 'this' ]);
```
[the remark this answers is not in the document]
Skaiir: True, but it's by design.

For the case where users would provide a context like this:

```
{
	parent: 'Greg',
	this: 'Jeremy',
	....
}
```

There have been cases already where users did this kind of thing, and it broke because of special access characters. We cannot know the context in advance, so we have to kind of assume it's always the case that 'this' is needed in this scenario.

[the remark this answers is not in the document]
Skaiir: If they use `_parent_` or `_this_` however, it will break, and rightfully so because at this point it's deserved :trollface:

packages/form-js-viewer/test/spec/util/GetSchemaVariables.spec.js
```diff
@@ -43,6 +45,22 @@ describe('util/getSchemaVariables', () => {
   });
 
 
+  it('should include variables in root of templates', () => {
+
+    const variables = getSchemaVariables(templateSchema);
+
+    expect(variables).to.eql([ 'myText', 'greeting', 'name', 'showAge', 'age', 'hobbies' ]);
```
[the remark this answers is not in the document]
Skaiir: `_parent_.experience`, but yeah, assuming you're only one layer deep. You could also just use `parent.experience`, assuming that there isn't already something called parent at that level.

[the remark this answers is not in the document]
Skaiir: If you're two levels deep, you'd do, `parent.parent.experience`.

packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
```diff
@@ -1,11 +1,47 @@
...
+    if (!this.isTemplate(template)) {
+      return [];
+    }
+
+    const expressions = this._extractExpressionsWithDepth(template);
+    const specialDepthAccessors = {
+      parent: [ -1, 1 ],
+      _parent_: [ -1 ],
+      this: [ 0, 1 ],
+      _this_: [ 0 ],
```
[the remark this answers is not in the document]
Skaiir: Yeah fair enough, but what this does is basically say:

`parent`: can be a child in the object structure (+1 one layer deepers in the tree), or a parent(-1  up one layer in the tree)
`_parent_`: is reserved for the actual parent (-1 up one layer in the tree)
`this`: could be self referential (0), or a property named this (1)
`_this_`: reserved for self reference

packages/form-js-viewer/src/features/expression-language/variableExtractionHelpers.js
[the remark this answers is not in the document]
Skaiir: That's just me obfuscating my GPT calls to actual do the work. (/s)

packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
Skaiir: Sure, will ask the AI to do this one :D

packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
Skaiir: So renaming wouldn't be the right call here because we're basically implementing an interface of the templating module. 

The reason why I am matching for feel is that my template language also handles the same syntax as the expression language. I could probably remove it, but at the same time the templating supports it, so uhm, technically this is the correct implementation. 

Usually the way the code is run is:

if (isExpression): do expression stuff
else if (isTemplate): do template stuff
else: do static stuff

Our API shouldn't care about the specifics of the module implementation, the way it's implemented right now, we are flexible. If I were to make modifications to remove this feel backwards compatibility from feelers, nothing would break in the app. It's a little confusing but at the same time I don't like purposely nerfing feelers to make it less redundant with the feel expression language implementation.

pinussilvestrus: Thanks for explaining, makes sense as is 👍

pinussilvestrus:
@Skaiir @vsgoulart maybe let's do a dry run on Tasklist locally by linking form-js to be super sure 🤞

vsgoulart:
> @Skaiir @vsgoulart maybe let's do a dry run on Tasklist locally by linking form-js to be super sure crossed_fingers

I can do that tomorrow

packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
```diff
@@ -1,11 +1,49 @@
...
+    };
+
+    return expressions.reduce((variables, { expression, depth }) => {
+      return [ ...variables, ...getFlavouredFeelVariableNames(expression, 'expression', { depth, specialDepthAccessors }) ];
+    }, []);
+  }
```
vsgoulart, line 45: How frequently is called? and how big are the datasets usually?

Might be better to reconsider using spread inside a loop for performance reasons

packages/form-js-viewer/src/features/expression-language/variableExtractionHelpers.js
```diff
@@ -0,0 +1,167 @@
...
+  return (function _unfoldVariables(node) {
+
+    if (node.name === 'PathExpression') {
+
+      if (Object.keys(specialDepthAccessors).length === 0) {
+        return depth === 0 ? [ _getVariableNameAtPathIndex(node, 0) ] : [ ];
+      }
+
+      // if using special depth accessors, use a more complex extraction
+      return [ ..._smartExtractVariableNames(node, depth, specialDepthAccessors) ];
```
vsgoulart, line 25: Can't we just return `_smartExtractVariableNames(node, depth, specialDepthAccessors)`? Or was returning a reference a problem?

you could also do `[].concat(_smartExtractVariableNames(node, depth, specialDepthAccessors))` since it should be a bit faster (performance might not an issue in this case though)

vsgoulart (approved):
Tested it on Tasklist and it seems to be working as expected.

packages/form-js-viewer/src/features/expression-language/FeelersTemplating.js
Skaiir: Anywhere from 1 - 20 I'd say. But I'll do a concat instead here.

packages/form-js-viewer/src/features/expression-language/variableExtractionHelpers.js
Skaiir: Oh actually wait it's because it's a set.

PR #604, opened 2023-04-17
chore: adjust resizing to trigger at midpoint between increments

Skaiir:
Adjusted so that the resizing triggers halfway between 2 columns, improving responsiveness.


Skaiir:
We may not want this after all, depends on if we allow 1 column spans.

pinussilvestrus:
Closing this as we decided to go with single column steps: https://github.com/bpmn-io/form-js/pull/606

PR #610, opened 2023-04-19
feat: implement validate on blur & input

### LINKED ISSUE #323: Eagerly validate form input

### LINKED ISSUE #382: Validate on blur

When users blur a required field, the validation does not start. It only starts after typing in something—and deleting the content again.

[IBM Carbon](https://carbondesignsystem.com/components/form/usage/#form-validation) describes to validate immediately after blur. This also helps to provide good a11y, since it allows screen readers and similar assistance devices to catch up with missing content before submission.

Start field validation directly on blur.

Skaiir:
Closes #382
Closes #323

Demo: https://demo-382-validate-on-blur--camunda-form-playground.netlify.app/

### DIFF
--- a/packages/form-js-viewer/src/features/viewerCommands/ViewerCommands.js
+++ b/packages/form-js-viewer/src/features/viewerCommands/ViewerCommands.js
@@ -1,0 +1,38 @@
+import UpdateFieldValidationHandler from './cmd/UpdateFieldValidationHandler';
+
+export default class ViewerCommands {
+  constructor(commandStack, eventBus) {
+    this._commandStack = commandStack;
+
+    eventBus.on('form.init', () => {
+      this.registerHandlers();
+    });
+  }
+
+  registerHandlers() {
+    Object.entries(this.getHandlers()).forEach(([ id, handler ]) => {
+      this._commandStack.registerHandler(id, handler);
+    });
+  }
+
+  getHandlers() {
+    return {
+      'formField.validation.update': UpdateFieldValidationHandler
+    };
+  }
+
+  updateFieldValidation(field, value) {
+    const context = {
+      field,
+      value
+    };
+
+    this._commandStack.execute('formField.validation.update', context);
+  }
+
+}
+
+ViewerCommands.$inject = [
+  'commandStack',
+  'eventBus'
+];
--- a/packages/form-js-viewer/src/render/components/FormField.js
+++ b/packages/form-js-viewer/src/render/components/FormField.js
@@ -1,4 +1,4 @@
-import { useContext } from 'preact/hooks';
+import { useCallback, useContext, useEffect, useMemo } from 'preact/hooks';
 
 import { get } from 'min-dash';
 
@@ -24,9 +24,11 @@ export default function FormField(props) {
   } = props;
 
   const formFields = useService('formFields'),
+        viewerCommands = useService('viewerCommands', false),
         form = useService('form');
 
   const {
+    initialData,
     data,
     errors,
     properties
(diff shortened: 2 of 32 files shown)


Skaiir:
Rebased to develop, surprisingly went very smoothly despite 200 commits behind :D

pinussilvestrus:
I will have a look at this tomorrow 👍 @christian-konrad @RomanKostka feel free to have a look at this as well, as this quite a change in the validation flow. 

https://demo-382-validate-on-blur--camunda-form-playground.netlify.app/

packages/form-js-viewer/src/features/viewerCommands/behaviors/index.js
```diff
@@ -0,0 +1,9 @@
+// import SomethingBehavior from 'SomethingBehavior';
+
+export default {
+
+  // __init__: [ 'somethingBehavior' ]
+  // somethingBehavior: [ 'type', SomethingBehavior ]
+};
+
+export { /* SomethingBehavior */ };
```
pinussilvestrus, line 9: This can be removed I guess?

packages/form-js-viewer/src/render/components/FormField.js
```diff
@@ -55,6 +59,18 @@ export default function FormField(props) {
...
+    if (viewerCommands) {
+      viewerCommands.updateFieldValidation(field, value);
+    }
+  }, [ viewerCommands, field, value ]);
+
+  useEffect(() => {
+    if (viewerCommands && initialValue) {
+      viewerCommands.updateFieldValidation(field, initialValue);
+    }
+  }, [ viewerCommands, field, initialValue ]);
```
pinussilvestrus, line 72: I'm wondering whether this should be configurable, so you can restore the legacy behavior if you want to. Something like (with a better name):

```js
const form = new Form({
  properties: {
    noValidationBlurChange: true
  }
});
```

Let's also add a note in the breaking changes in the Changelog already 👍

Skaiir: Since this is kind of a standard behavior, I would say it's not necessarily something we should qualify as a breaking change. Personally my take is that we treat it as a UX improvement, and if someone gets back to us saying they preferred the old behavior, we implement a conditional on this.

packages/form-js-viewer/src/features/viewerCommands/behaviors/index.js
Skaiir: Yeah I got rid of it, we can bring in behaviors some time later it's not like we lack the examples for this.

packages/form-js-viewer/src/render/components/FormField.js
pinussilvestrus: agree!

pinussilvestrus (approved):

packages/form-js-viewer/src/features/viewerCommands/ViewerCommands.js
```diff
@@ -0,0 +1,38 @@
...
+  constructor(commandStack, eventBus) {
+    this._commandStack = commandStack;
+
+    eventBus.on('form.init', () => {
+      this.registerHandlers();
+    });
+  }
+
+  registerHandlers() {
+    Object.entries(this.getHandlers()).forEach(([ id, handler ]) => {
```
nikku, line 13: I found the usage of CommandStack in the viewer context interesting here. Just to better understand what is happening, why did we add it? Is undo/redo a needed capability?

nikku: CC @Skaiir

PR #615, opened 2023-04-19
fix: custom properties/static values redundancy

### LINKED ISSUE #614: Adding custom properties doesn't work when certain keys are set
### Describe the Bug

image

In the above scenario, we cannot add any new custom keys, this is due to the fact that we generate the key name based on the number of existing keys, so in the example we keep overriding key2 with a new key2.

A similar problem exists with static values, but in this case, we get duplicate labels and values.

image

The different behavior here is due to the fact that we store these as array, and not maps. 

### Steps to Reproduce

1. add on custom property
2. rename the key to key2
3. try to add more
4. nothing happens

### Expected Behavior

4. a new key should be added, most likely with the number incremented by 1 (or however many times needed to get a new key name)

Skaiir:
Close #614

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/StaticValuesSourceEntry.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/StaticValuesSourceEntry.js
@@ -20,7 +20,7 @@ export default function StaticValuesSourceEntry(props) {
 
     const index = values.length + 1;
 
-    const entry = getIndexedEntry(index);
+    const entry = getIndexedEntry(index, values);
 
     editField(field, VALUES_SOURCES_PATHS[VALUES_SOURCES.STATIC], arrayAdd(values, values.length, entry));
   };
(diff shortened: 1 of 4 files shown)


packages/form-js-editor/src/features/properties-panel/entries/StaticValuesSourceEntry.js
```diff
@@ -75,16 +75,33 @@ export default function StaticValuesSourceEntry(props) {
...
     entry.label += ` ${index}`;
     entry.value += `${index}`;
   }
 
   return entry;
+}
+
+function labelOrValueIsAlreadyAssignedForIndex(index, values) {
+
+  if (index === 0) {
```
pinussilvestrus, line 98: I wonder whether this can ever happen, as we set `const index = values.length + 1;` above.

Skaiir: You're right, this is ironically, a redundant check :trollface:

packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
```diff
@@ -2998,6 +3029,34 @@ describe('properties panel', function() {
     });
 
 
+    it('should add property with different index if already used', function() {
+
+      // given
+
+      const editFieldSpy = spy();
```
pinussilvestrus, line 3036:
```suggestion
      // given
      const editFieldSpy = spy();

```

pinussilvestrus (approved):

PR #618, opened 2023-04-20
fix: prioritize adorner sizing

### LINKED ISSUE #598: Adorners cut off their text depending on window width
### Describe the Bug

You can't reliably use adorners, since they cut off their own text depending on the window width.

Bildschirmfoto 2023-04-12 um 19 43 52

Bildschirmfoto 2023-04-12 um 19 44 28

### Steps to Reproduce

1. Add a adorner (suffix or prefix) to a component
2. Resize the window
3. The adorner resizes, and likely hides it's text

### Expected Behavior

Adorners should always have prio compared with the content, i.e. the content can be cropped, but not the adorner (since the adorner provides important context information and can't be scrolled)

### Environment

- Recent form.js 0.14.0 / Tasklist 8.2.1

Skaiir:
Closes #598

### DIFF
--- a/packages/form-js-viewer/assets/form-js-base.css
+++ b/packages/form-js-viewer/assets/form-js-base.css
@@ -272,7 +272,6 @@
   padding: 8px;
   width: auto !important;
   min-width: 34px;
-  max-width: 30%;
   display: flex;
   overflow: hidden;
 }
@@ -354,6 +353,7 @@
 .fjs-container .fjs-input-group .fjs-taglist,
 .fjs-container .fjs-input-group .fjs-select {
   height: unset;
+  min-width: min(60px, 50%);
   width: 100%;
 }
 


vsgoulart (approved):

pinussilvestrus:
Maybe let's create a demo and ask @RomanKostka and @christian-konrad for feedback, to avoid another back and forth with this 👍

pinussilvestrus:
Demo can be found under: https://demo-598-prioritize-adorner-sp--camunda-form-playground.netlify.app/

RomanKostka (changes requested):
way better!
i found two things:
1. the definition view of the input seems not behaving right. the input grows without being resized.  Not sure if this is by intention
image

2. nothing for this issue but something worth to mention again and again (as far as i know this is already tracked in another issue):
we should give our components a min-width to avoid things like this:
image

Skaiir:
Both problems are present before and after the fix, so we're still merging this for the sake of iterative improvements. I will track the first in a new issue.

Skaiir:
[1]: https://github.com/bpmn-io/form-js/issues/623
[2]: https://github.com/bpmn-io/form-js/issues/624

PR #620, opened 2023-04-20
feat(refactor): extract component config to obj

### LINKED ISSUE #612: Move component options to an object

Currently we define a bunch of parameters for each component. We need these to fine-tune the behavior of components, however, it causes problems in some cases, like overriding component behaviors. 

image

image

We should wrap these up in a `ComponentParameters`, object so that the interface is a bit more clear which will allow us to be a bit more flexible.

-

Skaiir:
Closes #612

Please smoke test the app, this touches a huge part of the codebase and some references might not be in place. I've checked pretty thoroughly but we never know.


packages/form-js-viewer/src/render/components/FormField.js
```diff
@@ -43,9 +43,9 @@ export default function FormField(props) {
     throw new Error(`cannot render field <${field.type}>`);
   }
 
-  const value = get(data, _path);
+  const value = get(data, field._path);
```
pinussilvestrus, line 46: Why is this change necessary?

pinussilvestrus (changes requested):
Looks good to me 👍 

As we export the form components via `viewer`, we should at least mention this change in the changelog. Let's add a note to `Unreleased`. I'm sure this is rather something internally, we to be sure.

```md
### Breaking Changes

We change the structure in static component configuration properties. To use these properties, you need to access the `config` property of each individual component.
```

```js
import { Button } from '@bpmn-io/form-js-viewer';

console.log('Button default label is ' + Button.config.label);
```

packages/form-js-viewer/src/render/components/FormField.js
Skaiir: I wanted to homogenize the access, it was weird to me that we did object de-structuring on this one. But I forgot to delete the above.

pinussilvestrus (approved):

PR #625, opened 2023-04-21
fix: sanitize textfield newlines 

### LINKED ISSUE #380: Newlines don't get cleaned up when provided as input to text fields
__Describe the Bug__

While not rendered in the text field, newlines provided via inputs are still present in the output data, which is confusing

image
image

__Steps to Reproduce__

1. create a text field
2. set its key to `a`
3. set input data for `a` to a string containing `\n`

__Expected Behavior__

To discuss

### LINKED ISSUE #612: Move component options to an object

Skaiir:
Draft due to dependency on #620 
Closes #380

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Textfield.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Textfield.js
@@ -75,6 +75,17 @@ Textfield.config = {
   label: 'Text field',
   group: 'basic-input',
   emptyValue: '',
-  sanitizeValue: ({ value }) => (isArray(value) || isObject(value)) ? '' : String(value),
+  sanitizeValue: ({ value }) => {
+    if (isArray(value) || isObject(value)) {
+      return '';
+    }
+
+    // sanitize newlines to spaces
+    if (typeof value === 'string') {
+      return value.replace(/[\r\n\t]/g, ' ');
+    }
+
+    return String(value);
+  },
   create: (options = {}) => ({ ...options })
 };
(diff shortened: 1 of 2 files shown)


packages/form-js-viewer/src/render/components/form-fields/Textfield.js
```diff
@@ -75,6 +75,17 @@ Textfield.config = {
...
   emptyValue: '',
-  sanitizeValue: ({ value }) => (isArray(value) || isObject(value)) ? '' : String(value),
+  sanitizeValue: ({ value }) => {
+    if (isArray(value) || isObject(value)) {
+      return '';
+    }
+
+    // sanitize newlines to spaces
+    if (typeof value === 'string') {
+      return value.replace('\n', ' ');
```
vsgoulart, line 85: Is `sanitizeValue` called on change? I think we could change to the option below to cover all occurrences, specially if a user pastes the text.

Also, do we have to worry about other special symbols like tab? We could do a more complete replacement like this `value.replace(/\n/g, ' ');`

```suggestion
      return value.replace(/[\r\n\t]/g, ' ');
```

Skaiir: Yeah I agree, let's do that.

Skaiir: And not it's not called onChange, I think we want to handle that at the component level if we need to. Sometimes we make use of it onChange, but it's explicit.

vsgoulart (approved):

PR #626, opened 2023-04-21
feat: Correlate error messages to fields

### LINKED ISSUE #203: Make error messages accessible
__Describe the Bug__

Currently, `@bpmn-io/form-js` shows errors, but there's nothing connecting these errors to the appropriate fields. This can be achieved by some methods described on this [link](https://www.w3.org/WAI/tutorials/forms/notifications/#inline)

__Steps to Reproduce__

1. Model a form that has a field with a validation
2. Render this form and try to submit it with invalid
3. The error message is shown, but it's not accessible (The field doesn't have any `aria-describedby` prop pointing to the error)

__Expected Behavior__

The error messages should be accessible

vsgoulart:
Closes #203

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Textfield.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Textfield.js
@@ -47,6 +47,7 @@ export default function Textfield(props) {
   };
 
   const { formId } = useContext(FormContext);
+  const errorMessageId = errors.length === 0 ? undefined : `${prefixId(id, formId)}-error-message`;
 
   return <div class={ formFieldClasses(type, { errors, disabled }) }>
     <Label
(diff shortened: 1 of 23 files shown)


packages/form-js-viewer/src/render/components/form-fields/Textfield.js
```diff
@@ -60,10 +61,11 @@ export default function Textfield(props) {
         id={ prefixId(id, formId) }
         onInput={ onChange }
         type="text"
-        value={ value } />
+        value={ value }
+        aria-describedby={ errorMessageId } />
     </InputAdorner>
     <Description description={ description } />
-    <Errors errors={ errors } />
+    <Errors errors={ errors } id={ errorMessageId } />
```
pinussilvestrus, line 68: We should add a [a11y test case](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-viewer/test/spec/render/components/form-fields/Textfield.spec.js#L249) where we render an error and check that everything works as expected.

```js
it('should have no violations', async function() {

  // given
  this.timeout(5000);

  const { container } = createTextfield({
    value: 'John Doe Company',
    errors: [ 'Something went wrong' ]
  });

  // then
  await expectNoViolations(container);
});
```

Not necessarily for each form field, but at least a few would make a lot of sense.

vsgoulart: Should be done now

pinussilvestrus (approved):

PR #632, opened 2023-04-24
test: Setup visual regression tests

vsgoulart:
Related to https://github.com/camunda/team-hto/issues/76


vsgoulart:
@pinussilvestrus @Skaiir I wanted to add the Playwright setup inside a package in the monorepo, but it wasn't working properly with Lerna+Nx, so I decided to put it on the root

e2e/README.md
```diff
@@ -0,0 +1,12 @@
+# E2E Tests
```
pinussilvestrus, line 1: Should we quickly mention this one in the main README? https://github.com/bpmn-io/form-js#build-and-run

So it's easier to find on first setup.

package.json
```diff
@@ -28,7 +28,12 @@
     "start:viewer": "npm run start:pkg -- @bpmn-io/form-js-viewer",
     "start:viewer:stress": "cross-env SINGLE_START=stress npm run dev -- @bpmn-io/form-js-viewer",
     "start:editor": "npm run start:pkg -- @bpmn-io/form-js-editor",
-    "start:playground": "npm run start:pkg -- @bpmn-io/form-js-playground"
+    "start:playground": "npm run start:pkg -- @bpmn-io/form-js-playground",
+    "dev:e2e": "vite",
+    "test:visual": "playwright test visual",
+    "build:e2e": "vite build",
```
pinussilvestrus, line 34: This gives me an error on my arm64 macOS machine

```text
> vite build

failed to load config from /work/vite.config.js
error during build:
Error:
You installed esbuild for another platform than the one you're currently using.
This won't work because esbuild is written with native code and needs to
install a platform-specific binary executable.

Specifically the "@esbuild/darwin-arm64" package is present but this platform
needs the "@esbuild/linux-arm64" package instead. People often get into this
situation by installing esbuild on Windows or macOS and copying "node_modules"
into a Docker image that runs Linux, or by copying "node_modules" between
Windows and WSL environments.

If you are installing with npm, you can try not copying the "node_modules"
directory when you copy the files over, and running "npm ci" or "npm install"
on the destination platform after the copy. Or you could consider using yarn
instead of npm which has built-in support for installing a package on multiple
platforms simultaneously.

If you are installing with yarn, you can try listing both this platform and the
other platform in your ".yarnrc.yml" file using the "supportedArchitectures"
feature: https://yarnpkg.com/configuration/yarnrc/#supportedArchitectures
Keep in mind that this means multiple copies of esbuild will be present.

Another alternative is to use the "esbuild-wasm" package instead, which works
the same way on all platforms. But it comes with a heavy performance cost and
can sometimes be 10x slower than the "esbuild" package, so you may also not
want to do that.

    at generateBinPath (/work/node_modules/esbuild/lib/main.js:1874:17)
    at esbuildCommandAndArgs (/work/node_modules/esbuild/lib/main.js:1955:33)
    at ensureServiceIsRunning (/work/node_modules/esbuild/lib/main.js:2119:25)
    at build (/work/node_modules/esbuild/lib/main.js:2011:26)
    at bundleConfigFile (file:///work/node_modules/vite/dist/node/chunks/dep-24daf00c.js:64347:26)
    at loadConfigFromFile (file:///work/node_modules/vite/dist/node/chunks/dep-24daf00c.js:64323:31)
    at resolveConfig (file:///work/node_modules/vite/dist/node/chunks/dep-24daf00c.js:63932:34)
    at build (file:///work/node_modules/vite/dist/node/chunks/dep-24daf00c.js:46246:26)
    at CAC.<anonymous> (file:///work/node_modules/vite/dist/node/cli.js:812:15)
````

pinussilvestrus (commented):
I like this 👏 I believe it's fine to have this in the root folder. If we see problems with that we can always move to a sub package as `form-js-integration`.

/cc @bpmn-io/modeling-dev fyi

nikku:
To better understand what you're doing here a descriptive PR could help :see_no_evil:.

What do we plan to do / accomplish with this PR?

package.json
vsgoulart: @pinussilvestrus Added a new note [here](https://github.com/bpmn-io/form-js/pull/632/files#diff-d9770bfbd9e3f53581cb560341b3ab0afe2fae65ab497e8d05f2877a429513e1R14)

I tested it on my Mac M2 and it worked. Let me know what you think

pinussilvestrus (approved):
Works for me 👍 

Let's add more tests as a follow up. One nice use case would be to incorporate the carbon styles (cf. #557).

vsgoulart:
> To better understand what you're doing here a descriptive PR could help see_no_evil.
> 
> What do we plan to do / accomplish with this PR?

@nikku Yeah, sorry for the lack of context. We discussed during our team meetings so I didn't put more context here.
The intention is basically what Niklas mentioned. We've been using it on Tasklist with some success, so now we decided to add it to `form-js`.

Below you can see a video of how it works.

https://user-images.githubusercontent.com/1022916/234344261-d5763e93-b341-4035-9ee1-3cdf89145784.mp4

nikku:
Thanks for providing these details. I think I understood now.

Issue #396, opened 2022-11-09
Support `readonly` fields via FEEL expression


pinussilvestrus:
__What should we do?__

Implement the `readonly` property.

Sketches
![image](https://user-images.githubusercontent.com/9433996/234811314-5a4fdbfa-4c1d-4df1-ba09-1f61cb8a3ce1.png)

![image](https://user-images.githubusercontent.com/9433996/234811380-b249b4c0-fcc1-4d8c-9aee-91eb32314665.png)

* [x] add `readonly` property for form fields
* [x] allow to configure `readonly` in the properties panel
* [x] style `readonly` fields
* [ ] make it possible to add an expression to set the `readonly` condition
  * [ ] update `getSchemaVariables` 
* [x] add carbonized styles of `readonly`
* [x] update schema version
* [ ] update JSON schema (WIP)
* [ ] update linting rules
* [ ] update documentation

Up for discussion: Shortly, we potentially want to have a choice to provide an expression or a simple checkbox to define such a simple property (cf. https://github.com/bpmn-io/properties-panel/issues/194). 

__Why should we do it?__

Child of https://github.com/bpmn-io/form-js/issues/587
Depends on #374 
Design specs: https://www.figma.com/file/CiTSN1vNETjIAy6W9jUIFp/Form-Editor?node-id=1034-148176&t=T2sW2Ylfu9QonnHr-0

pinussilvestrus:
Demo spike available via #399 

_Missing for `develop`_
- clarify generic FEEL expression handling together with conditional-rendering
- label is only FEEL for demo purposes (to be removed)
- clarify readonly styling for all components (most are simply not interactible for now)

pinussilvestrus:
Design specs: https://www.figma.com/file/CiTSN1vNETjIAy6W9jUIFp/Form-Editor?node-id=1034-148176&t=T2sW2Ylfu9QonnHr-0

pinussilvestrus:
_Originally posted [via Slack](https://camunda.slack.com/archives/C046XS3GGCR/p1682602137794199)_

![image](https://user-images.githubusercontent.com/9433996/234877328-df1f625f-5cfd-493c-92ae-938f87f4d1df.png)

PR #636, opened 2023-05-02
Add `readonly` property

pinussilvestrus:
Related to #396 

This
* adds `readonly` property for all input fields
* configures `readonly` in the properties panel via toggle switch
* adds `properties.disabled` and adjusts the behavior of `properties.readonly` (cf. the breaking change below).
  => **Rationale**: I believe it makes sense to make this change now to clean up the behaviors. We still need `properties.disabled` to have the correct behavior in the editor where we don't want the inputs to be interactive (only the upper form elements).
* adds carbon styles for `readonly`

Demo (without Carbon): https://demo-396-readonly-property--camunda-form-playground.netlify.app/
To show the carbon styles:  `npm run start:carbon`

Bildschirm­foto 2023-05-03 um 11 26 18

-------

BREAKING CHANGES (tbd)

We changed the behavior when providing the `readOnly` property to a Form. From this version, the form fields will be rendered as `readOnly` if the property is set. Previously, the form fields were rendered as `disabled`. To restore the same behavior, please use the `disabled` property instead.

```js
const form = new Form({
  container: document.querySelector('#form'),
  properties: {
    disabled: true
  }
});
```

### DIFF
--- a/packages/form-js-carbon-styles/src/carbon-styles.js
+++ b/packages/form-js-carbon-styles/src/carbon-styles.js
@@ -57,110 +57,108 @@ const getSelectArrowStyles = ({
 `;
 
 const MARKDOWN_STYLES = css`
-  ${({ theme }) => css`
-    .fjs-container .fjs-form-field.fjs-form-field-text .markup {
-      & {
-        font-size: var(--cds-body-long-01-font-size);
-        font-weight: var(--cds-body-long-01-font-weight);
-        line-height: var(--cds-body-long-01-line-height);
-        letter-spacing: var(--cds-body-long-01-letter-spacing);
-      }
+  .fjs-container .fjs-form-field.fjs-form-field-text .markup {
+    & {
+      font-size: var(--cds-body-long-01-font-size);
+      font-weight: var(--cds-body-long-01-font-weight);
+      line-height: var(--cds-body-long-01-line-height);
+      letter-spacing: var(--cds-body-long-01-letter-spacing);
+    }
 
-      & h1 {
-        font-size: var(--cds-productive-heading-06-font-size);
-        font-weight: var(--cds-productive-heading-06-font-weight);
-        line-height: var(--cds-productive-heading-06-line-height);
-        letter-spacing: var(--cds-productive-heading-06-letter-spacing);
-      }
+    & h1 {
+      font-size: var(--cds-productive-heading-06-font-size);
+      font-weight: var(--cds-productive-heading-06-font-weight);
+      line-height: var(--cds-productive-heading-06-line-height);
+      letter-spacing: var(--cds-productive-heading-06-letter-spacing);
+    }
 
-      & h2 {
-        font-size: var(--cds-productive-heading-05-font-size);
-        font-weight: var(--cds-productive-heading-05-font-weight);
-        line-height: var(--cds-productive-heading-05-line-height);
-        letter-spacing: var(--cds-productive-heading-05-letter-spacing);
-      }
+    & h2 {
+      font-size: var(--cds-productive-heading-05-font-size);
+      font-weight: var(--cds-productive-heading-05-font-weight);
+      line-height: var(--cds-productive-heading-05-line-height);
+      letter-spacing: var(--cds-productive-heading-05-letter-spacing);
+    }
 
-      & h3 {
-        font-size: var(--cds-productive-heading-04-font-size);
-        font-weight: var(--cds-productive-heading-04-font-weight);
-        line-height: var(--cds-productive-heading-04-line-height);
-        letter-spacing: var(--cds-productive-heading-04-letter-spacing);
-      }
-      & h4 {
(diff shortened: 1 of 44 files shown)


vsgoulart (changes requested):
I found some issues on the Tasklist integration
image

packages/form-js-viewer/src/render/components/form-fields/parts/Timepicker.js
```diff
@@ -141,18 +142,20 @@ export default function Timepicker(props) {
...
           class="fjs-input"
           value={ rawValue }
           disabled={ disabled }
+          readOnly={ readonly }
           placeholder={ use24h ? 'hh:mm' : 'hh:mm ?m' }
           autoComplete="off"
-          onFocus={ () => useDropdown && setDropdownIsOpen(true) }
-          onClick={ () => useDropdown && setDropdownIsOpen(true) }
+          onFocus={ () => !readonly && useDropdown && setDropdownIsOpen(true) }
+          onClick={ () => !readonly && useDropdown && setDropdownIsOpen(true) }
```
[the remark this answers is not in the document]
pinussilvestrus: Exactly, only when the input is active or readonly.

pinussilvestrus:
> I found some issues on the Tasklist integration image

I'm inclined of introducing a layer into the local form-js test as well. We'd anyway quite often run into that problem:
* form-js has no layer
* Tasklist form has layer
=> the `cds-background` color will not match

pinussilvestrus:
@vsgoulart feel free to re-check, it should be fine now.

vsgoulart:
> @vsgoulart feel free to re-check, it should be fine now.

@pinussilvestrus The integration seems to be working well. I'll do a more thorough check tomorrow on the styles

packages/form-js-carbon-styles/src/carbon-styles.js
```diff
@@ -199,6 +197,108 @@ const ANCHOR_STYLES = css`
...
+            background: transparent;
+          }
+        }
+
+        &.fjs-form-field-select {
+          .fjs-input-group {
+            cursor: unset;
+            background-image: ${getSelectArrowSvg(theme.iconDisabled)};
+            background-color: transparent;
+            border-bottom: 1px solid var(--cds-border-strong, var(--cds-border-strong-01));
```
vsgoulart, line 236: I think you don't need this fallback, it's a Carbon internal thing and it should be covered already if you set up the Carbon properly.

Did you find any issues with it?

```suggestion
            border-bottom: 1px solid var(--cds-border-strong);
```

packages/form-js-carbon-styles/src/carbon-styles.js
```diff
@@ -271,7 +371,7 @@ const DISABLED_STYLES = css`
     }
 
     .fjs-taglist.fjs-disabled .fjs-taglist-tag {
-      background-color: var(--cds-field-02);
+      background-color: var(--cds-field-01);
```
vsgoulart, line 374: I think the proper way of doing this is to just use `--cds-field`, no? This way it will automatically respect the layers. This was a probably a mistake of mine when implementing it

vsgoulart (commented):
The styles look good :+1: 

I just have those 2 comments

packages/form-js-carbon-styles/src/carbon-styles.js
pinussilvestrus: I added this according to the [design sheet](https://www.figma.com/file/CiTSN1vNETjIAy6W9jUIFp/Form-Editor?node-id=1034-148176&t=T2sW2Ylfu9QonnHr-0). It was supposed to be a bit less contrast in comparison to the current state. 

image

`--cds-field` would lead to the same background as for the taglist input.

pinussilvestrus: Oh actually it should be the current state, not sure why I changed it 😅 I will revert the change at all.

vsgoulart (approved):

RomanKostka (changes requested):
nice! states look great.
only one thing:
would be great to have the toggle moved to the right of the label.
image
With that we gain consistency across all toggles that we will introduce in future.
( all checkboxes will become toggles + fx-boolean-toggle )

pinussilvestrus:
@RomanKostka did you agree with @YanaSegal on something regarding [this comment](https://github.com/bpmn-io/form-js/pull/641#issuecomment-1534196065) in the other PR?

pinussilvestrus:
> @RomanKostka did you agree with @YanaSegal on something regarding [this comment](https://github.com/bpmn-io/form-js/pull/641#issuecomment-1534196065) in the other PR?

Just saw your comment 👍 https://github.com/bpmn-io/form-js/pull/641#issuecomment-1547384190

I will handle this as part of https://github.com/bpmn-io/properties-panel/pull/245. I think we are good to go in this PR then 👍

PR #641, opened 2023-05-03
chore(editor): move `disabled` entry to a toggle switch

### LINKED ISSUE #639: Move `disabled` to a toggle switch
### What should we do?

With #636 we aim to add the `readonly` property as a toggle switch. We should do the same for `disabled`, to be consistent.

### Why should we do it?

We think a toggle switch adds more value for these two properties than a simple checkbox. A toggle usually has an immediate effect on the UI, which is the case.

pinussilvestrus:
Closes #639

image

image

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/DisabledEntry.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/DisabledEntry.js
@@ -2,7 +2,7 @@ import { get } from 'min-dash';
 
 import { INPUTS } from '../Util';
 
-import { CheckboxEntry, isCheckboxEntryEdited } from '@bpmn-io/properties-panel';
+import { ToggleSwitchEntry, isToggleSwitchEntryEdited } from '@bpmn-io/properties-panel';
 
 
 export default function DisabledEntry(props) {
@@ -23,7 +23,7 @@ export default function DisabledEntry(props) {
       component: Disabled,
       editField: editField,
       field: field,
-      isEdited: isCheckboxEntryEdited
+      isEdited: isToggleSwitchEntryEdited
     });
   }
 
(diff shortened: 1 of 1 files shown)


RomanKostka:
@pinussilvestrus 
should we directly move the toggle to the right?
image

pinussilvestrus:
The toggle switch component of the properties panel renders it top-bottom as you can configure a `switcherLabel` to describe the state of the toggle (e.g. on/off).

image

We could
* change this behavior in form-js as we anyway don't display such a switcher label
* change it in the properties panel library to display it inline once no switcher label is declared => this would need extra effort outside of this PR, but would solve it for all toggle switches

Wdyt?

vsgoulart (approved):
Code wise it looks ok and couldn't find any problems while testing.
It just occurred to me that it might be better to use `cursor: pointer` for the toggle, I just don't know how easy it is to achieve that on from `form-js` side

pinussilvestrus:
If we anyway would need to override the styles to display the toggle inline (cf. https://github.com/bpmn-io/form-js/pull/641#issuecomment-1534196065), then we could also add the pointer.

Depends on whether we see these things a form-js exclusive or worth adding to the properties panel core /cc @christian-konrad @RomanKostka

RomanKostka:
> If we anyway would need to override the styles to display the toggle inline (cf. https://github.com/bpmn-io/form-js/pull/641#issuecomment-1534196065), then we could also add the pointer.

> Depends on whether we see these things a form-js exclusive or worth adding to the properties panel core

@pinussilvestrus @christian-konrad @YanaSegal from my point of view this is worth it adding to the property panel.

pinussilvestrus:
Handling the styling improvements as part of https://github.com/bpmn-io/properties-panel/pull/245.

Issue #653, opened 2023-05-11
Enable FEEL or template for following properties


pinussilvestrus:
### What should we do?

Enable FEEL or template for following properties

- [x]  [read-only ](https://github.com/bpmn-io/form-js/pull/656) (optional feel toggle switch)
- [x]  [labels](https://github.com/bpmn-io/form-js/pull/658) (feelers one line/optional feel)
- [x]  [description](https://github.com/bpmn-io/form-js/pull/658) (feelers one line/optional feel)
- [x] [source and alt](https://github.com/bpmn-io/form-js/pull/663) (feelers one line/optional feel)
- [ ]  [min-max](https://github.com/bpmn-io/form-js/pull/668) (number/optional feel)
- [x]  [adorners](https://github.com/bpmn-io/form-js/pull/663) (feelers one line/optional feel)

### Why should we do it?

Child of https://github.com/bpmn-io/form-js/issues/587

PR #656, opened 2023-05-16
Readonly via expressions

### LINKED ISSUE #396: Support `readonly` fields via FEEL expression

pinussilvestrus:
Closes #396 
Related to #653 

-----

Adds expression support for `readonly`.

Demo: https://demo-653-readonly-expressions--camunda-form-playground.netlify.app/

image

### DIFF
--- a/packages/form-js-viewer/src/render/hooks/useReadonly.js
+++ b/packages/form-js-viewer/src/render/hooks/useReadonly.js
@@ -1,0 +1,31 @@
+import useService from './useService.js';
+import useFilteredFormData from './useFilteredFormData.js';
+
+/**
+ * Retrieve readonly value of a form field, given it can be an
+ * expression optionally or configured globally.
+ *
+ * @typedef { import('../../types').FormProperties } FormProperties
+ *
+ * @param {any} formField
+ * @param {FormProperties} properties
+ *
+ * @returns {boolean}
+ */
+export default function useReadonly(formField, properties = {}) {
+  const expressionLanguage = useService('expressionLanguage');
+  const conditionChecker = useService('conditionChecker', false);
+  const filteredData = useFilteredFormData();
+
+  const { readonly } = formField;
+
+  if (properties.readOnly) {
+    return true;
+  }
+
+  if (expressionLanguage && expressionLanguage.isExpression(readonly)) {
+    return conditionChecker ? conditionChecker.check(readonly, filteredData) : false;
+  }
+
+  return readonly || false;
+}
--- a/.github/workflows/TASKLIST_CARBONISATION.yml
+++ b/.github/workflows/TASKLIST_CARBONISATION.yml
@@ -7,7 +7,7 @@ jobs:
     name: Run visual regression
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.32.3-focal
+      image: mcr.microsoft.com/playwright:v1.34.1
       options: --user 1001:1000
     steps:
       - name: Checkout form-js
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "@babel/core": "^7.18.10",
         "@babel/plugin-transform-react-jsx": "^7.14.5",
         "@babel/plugin-transform-react-jsx-source": "^7.14.5",
-        "@bpmn-io/properties-panel": "^2.0.0",
+        "@bpmn-io/properties-panel": "^2.1.0",
         "@carbon/react": "^1.24.0",
         "@carbon/styles": "^1.23.1",
         "@playwright/test": "1.34.1",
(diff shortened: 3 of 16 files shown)


pinussilvestrus:
We can't merge this until https://github.com/bpmn-io/properties-panel/pull/245 is done, but feel free to already have a look 👍

packages/form-js-viewer/src/render/hooks/useReadonly.js
```diff
@@ -0,0 +1,32 @@
...
+  const conditionChecker = useService('conditionChecker', false);
+  const filteredData = useFilteredFormData();
+
+  const { readonly } = formField;
+
+  if (properties.readOnly) {
+    return true;
+  }
+
+  // todo(pinussilvestrus): memoize?
```
vsgoulart, line 26: Can we remove or create an issue for this?

vsgoulart (changes requested):
Outine looks wrong for me on Firefox and Chroms

image

packages/form-js-viewer/src/render/hooks/useReadonly.js
pinussilvestrus: I removed it.

pinussilvestrus:
Seems like there were wrong styles applied. Could you try again?

vsgoulart:
> Seems like there were wrong styles applied. Could you try again?

@pinussilvestrus Did you update the demo? The outline is still wrong for me

vsgoulart:
@pinussilvestrus We also have some visual regression tests failing

pinussilvestrus:
> @pinussilvestrus Did you update the demo? The outline is still wrong for me

It works for me on Firefox and Chrome (macOS).

image

> @pinussilvestrus We also have some visual regression tests failing

Yes, this is because we use another properties panel dependency which is not yet rebased on `v2.0.0`. Cf. https://github.com/bpmn-io/form-js/pull/657#issuecomment-1549709631. It will be fixed once we have the properties panel PR merged and integrated.

vsgoulart (approved):
@pinussilvestrus The focus is still wrong for me on Linux, maybe there's an issue because of the way we're consuming the properties panel dep? I'll approve if for now and I'll check it again when you use the real package
Can you please remind me of taking a look into this before you merge it?

pinussilvestrus:
@vsgoulart the dependency is updated now. Can you maybe double check? For reference, the arrow action for opening a properties group should have the same outline color. If it's black on your machine as well, maybe it's a general issue we'd need to address in the properties panel.

image

vsgoulart:
> @vsgoulart the dependency is updated now. Can you maybe double check? For reference, the arrow action for opening a properties group should have the same outline color. If it's black on your machine as well, maybe it's a general issue we'd need to address in the properties panel.
> 
> image

@pinussilvestrus Still wrong to me, but it seems unrelated to your PR so I just created a followup issue #664

vsgoulart:
There's also some issues with the visual regression tests, I'll fix it tomorrow

vsgoulart (approved):

pinussilvestrus:
@vsgoulart I updated the playwright version in the carbonization action as well (it's not on master, yet). I kept it in this PR.

vsgoulart:
> Ok @vsgoulart seems like due to the updated playwright image the emojis are rendered differently sweat_smile
> 
> image
> 
> I guess we accept that? I'd update the screenshot then.

@pinussilvestrus yeah, let's just update the screenshot :+1:

vsgoulart:
I think we should also generate a new `form-js` version and update it on Tasklist

christian-konrad (approved):
Works like a charm!

RomanKostka (approved):
Nice! Works as expected 🎉

i found a minor thing and am not sure if this was intended:

When i switch the toggle to `true`, then enable FEEL expressions, change the expression to `false` and then disable FEEL expressions the toggle is still `true`
image
image
I would expects that the toggle should become 'false'

The other way around it works as expected.
Switch the toggle to `true`, then again to `false`, enable FEEL expressions = `false`.

Also from the very beginning (before interacting with the toggle or the FEEL expression) the toggle is 'false' but the FEEL expression is empty.
image
I would expect the FEEL expression is `false` as the toggle is `false` too.

christian-konrad:
@RomanKostka empty evaluates to `false` as well, so leaving it empty reduces the need to delete `false` first from the input, and users can immediately start typing. Hence I would propose to keep it as it is!

For detecting `false` for the toggle: I got your point. We could build this special logic, however we can never provide a smart detection of falsy values (e.g. an expression or variable evaluates to `false`) so I would say let's not spend time on this one case now.

PR #667, opened 2023-05-25
chore(CHANGELOG): update to v1.0.0-alpha.0

pinussilvestrus:
@Skaiir @vsgoulart As I'm writing the third breaking change into the `v15.0.0-alpha.x` versions I'm inclined to release this as the first major alpha now. With the stuff we are going to release around additional modules next (@Skaiir), I think it deserves to be a major release 👍 

Let me know what you think!


vsgoulart (approved):
@pinussilvestrus Yeah, I think it's a good idea to bump the major version

PR #685, opened 2023-06-08
fix: refresh properties panel schema on re-import

Skaiir:


packages/form-js-editor/src/features/properties-panel/PropertiesPanel.js
```diff
@@ -53,57 +53,42 @@ export default function FormPropertiesPanel(props) {
 
   const formEditor = injector.get('formEditor');
   const modeling = injector.get('modeling');
-  const selection = injector.get('selection');
+  const selectionModule = injector.get('selection');
 
-  const { schema } = formEditor._getState();
+  const [ state , setState ] = useState({ selectedFormField: selectionModule.get() || formEditor._getState().schema });
```
vsgoulart, line 58: There's not reason to save this in a object like, this will just encourage us add another property to the object if we need to extend this, which will break reactivity with further unnecessary updates

```suggestion
  const [ selectedFormField , setSelectedFormField ] = useState(selectionModule.get() || formEditor._getState().schema);
```

vsgoulart (approved):

packages/form-js-editor/src/features/properties-panel/PropertiesPanel.js
Skaiir: As discussed, this is an old problem, we don't have proper reactivity in the properties panel, the following issue is my attempt at getting this sort of stuff fixed.

https://github.com/bpmn-io/form-js/issues/686

Issue #674, opened 2023-06-01
`TASKLIST_CARBONIZATION` action is flaky


pinussilvestrus:
### Describe the Bug

Our [GitHub action for checking the Takslist Carbonization](https://github.com/bpmn-io/form-js/blob/develop/.github/workflows/TASKLIST_CARBONISATION.yml) is flaky due to limited resources (we believe?). Let's find ways to mitigate that.

Example: https://github.com/bpmn-io/form-js/actions/runs/5141278596/jobs/9253592506

### Steps to Reproduce

1. Check out [action run](https://github.com/bpmn-io/form-js/actions/runs/5141278596/jobs/9253592506)
2. See we have time outs here

### Expected Behavior

Timeouts should not happen.

pinussilvestrus:
I found a fix for that: https://github.com/camunda/tasklist/pull/3115

PR #693, opened 2023-06-20
Add `fjs-no-theme` selector

### LINKED ISSUE #680: Offer styles without theming
### What should we do?

In some cases library users might want to use Carbon variables but not apply the theme to the form components. One example is the form-playground [which is not yet ready for using a proper theme](https://github.com/camunda/form-playground/pull/70) (cf. https://github.com/camunda/web-modeler/pull/4940#issuecomment-1578024353). If you integrate the Carbon variables into the public scope, the theme will immediately be applied.

image

We should offer a way to consume our styles without the theming, e.g. by providing a `no-theme` selector which removes all [theme variable bindings](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/assets/form-js-editor-base.css#L3).

### Why should we do it?

Allow our users to consume our styles without themes but Carbon styles imported.

pinussilvestrus:
Closes #680

* Adds `fjs-no-theme` selector to remove all themed variables from the components
* Add visual regression for no theme
* Add axe steps to some visual regression tests

Note: I considered adding this as a property to `FormEditor` and `Form` as well, but decided against this as I found it easier to use it via CSS selector. It's anyway just an internal feature we can use in Web Modeler. Let me know what you think.


.github/workflows/TASKLIST_CARBONISATION.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     name: Run visual regression
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.34.1
+      image: mcr.microsoft.com/playwright:v1.35.1
```
pinussilvestrus, line 10: Unfortunately, this still does not fix the timeout issue (cf. https://github.com/bpmn-io/form-js/issues/674)

package.json
```diff
@@ -78,7 +79,7 @@
     "@types/node": "^14.17.4",
     "@types/sinon": "^10.0.0",
     "@types/sinon-chai": "^3.2.5",
-    "axe-core": "^4.4.3",
+    "axe-core": "4.4.3",
```
pinussilvestrus, line 82: I had to pin this as it seems the axe version coming from playwright is not working with our karma setup.

vsgoulart: We probably should pin all our deps to be honest

e2e/visual/no-theme.spec.js
```diff
@@ -0,0 +1,84 @@
+const { expect } = require('@playwright/test');
+
+const { test: axeTest } = require('../axe-test');
```
vsgoulart, line 3: [I think we could just make a generic fixture like I did here on Tasklist](https://github.com/camunda/tasklist/blob/master/client/e2e/test-fixtures.ts)

package.json
pinussilvestrus: In bpmn-io land we always opted against this as it's not recommended for library providers (rationale: https://docs.renovatebot.com/dependency-pinning/#ranges-for-libraries, avoid duplicated libraries in the node modules). I'm not 100% sure what's best practice in the current ecosystem, but I'd still avoid pins until we really need it - as in this example.

For security reasons, I believe we should do housekeeping (`npm audit fix`) more often + utilize Renovate.

e2e/visual/no-theme.spec.js
pinussilvestrus: Thanks for the hint, I adjusted it via https://github.com/bpmn-io/form-js/pull/693/commits/0f3152ae77c91207f7ee2c658bb45b72a409effc 👍

vsgoulart (approved):

PR #697, opened 2023-06-29
Fix Carbonization

vsgoulart:
Closes camunda/tasklist#2966

### DIFF
--- a/packages/form-js-carbon-styles/src/carbon-styles.js
+++ b/packages/form-js-carbon-styles/src/carbon-styles.js
@@ -1,7 +1,6 @@
 // todo(pinussilvestrus): consider moving away from styled-components,
 // cf. https://github.com/bpmn-io/form-js/issues/633
 import { css, createGlobalStyle } from 'styled-components';
-
 import { rem } from '@carbon/elements';
 
 function getSelectArrowSvg(color) {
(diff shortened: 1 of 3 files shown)


packages/form-js-viewer/assets/form-js-base.css
```diff
@@ -75,7 +75,7 @@
     --cds-border-subtle, 
     var(--cds-border-subtle-01, var(--color-grey-225-10-85))
   );
-  --color-borders-readonly: var(--cds-border-disabled, var(--color-grey-225-10-75));
+  --color-borders-readonly: var(--cds-border-subtle, var(--color-grey-225-10-75));
```
vsgoulart, line 78: [The border color was wrong ](https://react.carbondesignsystem.com/?path=/story/components-select--playground&args=readOnly:true)

packages/form-js-carbon-styles/src/carbon-styles.js
```diff
@@ -1084,7 +1085,7 @@ const DROPDOWN_STYLES = css`
 
       & .fjs-dropdownlist-item:hover,
       & .fjs-dropdownlist-item.focused {
-        background-color: var(--cds-layer-hover, var(--cds-layer-hover-01));
+        background-color: var(--cds-layer-hover);
```
vsgoulart, line 1088: I also took the opportunity to remove the unnecessary fallbacks

pinussilvestrus:
I will have a look today 👍

packages/form-js-carbon-styles/src/carbon-styles.js
pinussilvestrus: I think while experimenting with the styles, I was able to create setups where the `cds-layer-hover` without a specific number wasn't specified, so that's why these fallbacks landed in initially. However, I think I pretty much misconfigured something, so it's very unlikely that this will happen.

pinussilvestrus (approved):
LGTM 👍 

I added one commit to use the layer-one in our test setup.

PR #711, opened 2023-07-12
fix(FEEL): vars in filters are properly parsed

Skaiir:
Related to https://github.com/camunda/team-hto/issues/279

### DIFF
--- a/packages/form-js-viewer/src/util/index.js
+++ b/packages/form-js-viewer/src/util/index.js
@@ -143,9 +143,7 @@ export function getSchemaVariables(schema, expressionLanguage = new FeelExpressi
       const property = get(component, prop.split('.'));
 
       if (property && !expressionLanguage.isExpression(property) && templating.isTemplate(property)) {
-
         const templateVariables = templating.getVariableNames(property);
-
         variables = [ ...variables, ...templateVariables ];
       }
     });
--- a/packages/form-js-viewer/src/util/index.js
+++ b/packages/form-js-viewer/src/util/index.js
@@ -148,7 +148,7 @@ export function getSchemaVariables(schema, expressionLanguage = new FeelExpressi
       }
     });
 
-    return variables;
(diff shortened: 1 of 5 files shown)


packages/form-js-viewer/test/spec/util/GetSchemaVariables.spec.js
```diff
@@ -148,4 +149,16 @@ describe('util/getSchemaVariables', () => {
     ]);
   });
 
+
+  it('should include variables in ships example', () => {
+
+    const variables = getSchemaVariables(shipsExampleSchema);
+
+    // currently produces: ['selectedShip', undefined] :zap:
```
pinussilvestrus, line 157:
```suggestion
```

pinussilvestrus (commented):
Really nice 👍 

👍 for all the additional test cases !

Skaiir:
@pinussilvestrus Fixed the one comment and also decided to filter null and undefined variables. Mostly so that we don't ever get that issue again.

packages/form-js-viewer/src/util/index.js
```diff
@@ -143,14 +143,12 @@ export function getSchemaVariables(schema, expressionLanguage = new FeelExpressi
...
       if (property && !expressionLanguage.isExpression(property) && templating.isTemplate(property)) {
-
         const templateVariables = templating.getVariableNames(property);
-
         variables = [ ...variables, ...templateVariables ];
       }
     });
 
-    return variables;
+    return variables.filter(variable => variable !== undefined || variable !== null);
```
pinussilvestrus, line 151: Not 100% sure about this one. The good thing about the current behavior (keeping null values) was that we catched the error in Tasklist really easily and it didn't slip our attention. However we should not crash the whole workflow, so I believe it's fine now.

pinussilvestrus (approved):

PR #714, opened 2023-07-13
test: Fix visual regression integration linking

vsgoulart:
@pinussilvestrus I think one of the reasons the tests weren't very reliable some weeks ago is because I forgot to link changes in the Carbon custom styles

### DIFF
--- a/.github/workflows/TASKLIST_CARBONISATION.yml
+++ b/.github/workflows/TASKLIST_CARBONISATION.yml
@@ -39,11 +39,17 @@ jobs:
       - name: Build form-js
         run: npm run build
         working-directory: ./form-js
-      - name: Link form-js-viewer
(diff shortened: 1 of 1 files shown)


pinussilvestrus (approved):
Makes sense. I guess we were a bit lucky that it didn't crash more often without 😄

pinussilvestrus:
I think we need to update the playwright image as well 😓 https://github.com/bpmn-io/form-js/actions/runs/5543067140/jobs/10118582951

vsgoulart:
> I think we need to update the playwright image as well sweat https://github.com/bpmn-io/form-js/actions/runs/5543067140/jobs/10118582951

@pinussilvestrus Yeah, my bad
I saw the wrong CI result I updated it here https://github.com/bpmn-io/form-js/pull/715

PR #720, opened 2023-07-17
fix: hide empty rows via css

### LINKED ISSUE #684: Hidden components take up some space
### Describe the Bug

Hidden components take up some space, since their wrappers still have padding. Both for native and Carbon style.

### Steps to Reproduce

1. Create a form with conditioned components
2. Run a form
3. Hide one component based on decision
4. It still takes up some space

Bildschirmfoto 2023-06-07 um 23 20 52

### Expected Behavior

Hidden components do not take up any space and visually behave like they do not exist.

### Environment

- OS: [e.g. Windows 7]

Skaiir:
Closes  #684

Initially tried to approach this one programmatically, but it's not as clean as it seems, so I thought the simplest and most effective approach is to just not render the row if it has no children.

### DIFF
--- a/packages/form-js-viewer/assets/form-js-base.css
+++ b/packages/form-js-viewer/assets/form-js-base.css
@@ -175,6 +175,10 @@
   position: relative;
 }
 
+.fjs-container .fjs-layout-row:empty {
+  display: none;
+}
+
 .fjs-container .fjs-column {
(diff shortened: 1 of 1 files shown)


pinussilvestrus (approved):
LGTM, works 👍

PR #724, opened 2023-07-18
Fixing carbonization issues

pinussilvestrus:
Related to https://github.com/camunda/tasklist/issues/3189
Related to https://github.com/camunda/tasklist/issues/2879

----

* Making select arrow clickable
  * The click handler is on the original arrow, so making it transparent so the click wents through
* Fixing wrong link styles in text views
  * Docs: https://carbondesignsystem.com/components/link/usage/ (inline)

image


Skaiir (approved):
LGTM 👍

PR #725, opened 2023-07-18
feat: empty form editor state

Skaiir:
Closes #336

PR #732, opened 2023-07-20
feat: implement `spacer` component

### LINKED ISSUE #731: Implement spacer component

In some of the usages we've observed, people are using the text view to add spacing to the form. This isn't ideal, so let's simplify this and have a component that simply serves this purpose. 

Vertical spacing should be handled via the grid, so no need to do anything there.
Horizontal sizing will be set via the properties panel for now, might make sense to allow drag resizing there as well in the future.

Reqs:
- [x] Component
- [x] Properties panel configuration
- [x] New Icon

Keep using the text components for now, but eh it's a low hanging fruit let's just set this one up.

Skaiir:
Closes #731


pinussilvestrus:
Same here as for https://github.com/bpmn-io/form-js/pull/725#issuecomment-1647345978

pinussilvestrus:
When the spacer is a min lower than 20, it looks a bit odd in the editor. 
image

Maybe shrinking the icon is an option here?

packages/form-js-editor/test/spec/FormEditor.spec.js
```diff
@@ -73,7 +73,7 @@ describe('FormEditor', function() {
     });
 
     // then
-    expect(formEditor.get('formFieldRegistry').getAll()).to.have.length(16);
+    expect(formEditor.get('formFieldRegistry').getAll()).to.have.length(17);
```
pinussilvestrus, line 76: I just updated the schema version via https://github.com/bpmn-io/form-js/pull/732/commits/73d004cbcd05ed645702bf8bf2145a0a2c359a5f ✅

pinussilvestrus:
@RomanKostka @christian-konrad feel free to have a look at this handy new component: https://demo-spacer-component--camunda-form-playground.netlify.app/

Skaiir:
@pinussilvestrus Actually I think another fix would be to limit the min-height to 20. In any case, if it gets smaller than that it becomes too small to easily drag around. Or we could limit the minHeight in the editor only, and allow any kind of value in the preview, but then we lose the visual sync. WDYT?

pinussilvestrus:
I think a min-height of 20 makes much sense here 👍 cc @RomanKostka

packages/form-js-editor/test/spec/FormEditor.spec.js
pinussilvestrus: 🚀 thanks!

Skaiir:
@pinussilvestrus I went for the approach of limiting the height via styles only, but allow configurations to be set lower. It's okay to miss out on the perfect sync, I think the scenario of a very short spacer is rare anyways so let's go flexibility over aesthetics in this case. Whenever I finish scrollsync ( 2024? :trollface: ) it'll not matter anyways.

christian-konrad (changes requested):
Very useful!
One thing on how it appears in the editor:

image

Most trial users do not open the preview (yet). Maybe they are confused about the symbol you show in the editor. What about this instead:

image

```
.fjs-editor-container .fjs-form-field-spacer {
  min-height: 30px;
  background: repeating-linear-gradient(  45deg,  #ffffff,  #ffffff 10px,  #f5f5f5 10px,  #f5f5f5 20px);
}
```

Skaiir:
@christian-konrad What a good idea!

PR #734, opened 2023-07-20
fix: render text placeholder for whitespace values

### LINKED ISSUE #721: Text views with non-rendering characters are rendered really short in the editor.
### Describe the Bug

image

The text view renders really tiny when no rendered characters are input as its text.

### Steps to Reproduce

1. Add a text view
2. Set the text value to one or several spaces

### Expected Behavior

Either this should be treated as an empty text view, or we should match the vertical spacing like regular text.

Skaiir:
Closes #721


Skaiir:
image

pinussilvestrus (approved):

PR #737, opened 2023-07-22
fix: textarea resizing behavior improvements

### LINKED ISSUE #546: Rendered height of multiline text components with default values is broken in Web Modeler
### Describe the Bug

The rendered height of multiline text components with default values is broken on Camunda Web Modeler.

image

image

### Steps to Reproduce

1. Add a multiline text
2. Add a default value
3. Save + leave the editor
4. Open the form again
--> The component height is too small to contain the text

### Expected Behavior

The component height should contain the whole text

### Environment

- Form Editor on recent Web Modeler as of February 15th. Does not happen on Desktop Modeler

Skaiir:
Closes #546

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Textarea.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Textarea.js
@@ -1,5 +1,5 @@
 import { isArray, isObject } from 'min-dash';
-import { useCallback, useContext, useEffect, useRef } from 'preact/hooks';
+import { useContext, useEffect, useLayoutEffect, useRef } from 'preact/hooks';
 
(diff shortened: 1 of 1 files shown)


Skaiir:
The fix isn't exactly perfect, but this is the 3rd time I try and fix this bug. So we'll keep with that, behavior was a lot nicer though, with proper resizing on load and no frame-flicker when resizing anymore thanks to useLayoutEffect.

I opted to not test, because it's already really finicky to replicate the bug as it stands in the browser, so the test containers I don't even want to think about it.

pinussilvestrus (approved):
Looks smooth to me in the demo 👍

PR #744, opened 2023-07-25
chore(GHA): overwrite screenshots on master/dev

### LINKED ISSUE #743: Implement a way to update playwright screenshots via GHA
### What should we do?

We should be able to trigger a playwright screenshot overwrite via GHA. 

### Why should we do it?

We only upload visual regression tests for mac, but this also means non-mac users can't overwrite the playwright tests. Which obviously is a bottleneck in dev experience. Allowing us to trigger an overwrite would circumvent this need to have a mac user on hand every time we change something visual :D

Skaiir:
Closes #743


pinussilvestrus (approved):
Let's try this out 🔥

PR #758, opened 2023-08-02
feat: localize datepicker dates to browser

### LINKED ISSUE #733: Support alternate date format

As a first iteration, we had decided to simply provide an American date format for the date picker, but now we've had some reasons to support multiple, obviously, not everyone is used to the American date format, so this makes the component a little incomplete.

We have two ways we can go with this, either configure the format, or auto-convert based on the local environment, which would make the most sense really, as this would require no configuration overheads and appear correct everywhere the component is used (much like the time is treated currently). 

-

### Additional context

Conversations leading up to this (internal access only): https://camunda.slack.com/archives/C046XS3GGCR/p1689630534887359

Skaiir:
Closes #733

### DIFF
--- a/packages/form-js-viewer/src/render/components/util/localisationUtil.js
+++ b/packages/form-js-viewer/src/render/components/util/localisationUtil.js
@@ -1,0 +1,68 @@
+/**
+ * Returns date format for the provided locale.
+ * If the locale is not provided, uses the browser's locale.
+ *
+ * @param {string} [locale] - The locale to get date format for.
+ * @returns {string} The date format for the locale.
+ */
+export function getLocaleDateFormat(locale = 'default') {
+  const parts = new Intl.DateTimeFormat(locale).formatToParts(new Date(Date.UTC(2020, 5, 5)));
+  return parts.map(part => {
+    const len = part.value.length;
+    switch (part.type) {
+    case 'day': return 'd'.repeat(len);
+    case 'month': return 'M'.repeat(len);
+    case 'year': return 'y'.repeat(len);
+    default: return part.value;
+    }
+  }).join('');
+}
+
+/**
+ * Returns readable date format for the provided locale.
+ * If the locale is not provided, uses the browser's locale.
+ *
+ * @param {string} [locale] - The locale to get readable date format for.
+ * @returns {string} The readable date format for the locale.
+ */
+export function getLocaleReadableDateFormat(locale) {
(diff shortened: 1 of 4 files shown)


Skaiir:
Visual regression most likely failing due to the locale it's running on not being en-us, which is what we were enforcing initially.

packages/form-js-viewer/src/render/components/util/localisationUtil.js
```diff
@@ -0,0 +1,404 @@
...
+  };
+
+  dateFormat = useLeadingZero.day ? dateFormat.replace('dd', 'd') : dateFormat.replace('d', 'j');
+  dateFormat = useLeadingZero.month ? dateFormat.replace('MM', 'm') : dateFormat.replace('M', 'n');
+  dateFormat = useLeadingZero.year ? dateFormat.replace('yyyy', 'Y') : dateFormat.replace('yy', 'y');
+
+  return dateFormat;
+}
+
+const countryFormats = {
```
vsgoulart, line 68: Why do we have these mappings if we're not using them?

Skaiir: We do use them at line 11

Skaiir: Basically, not everything is supported all the time, so we have specific mappings that I took from a source of international mappings (as a big dictionary instead of having yet another dependency), I also created a fallback dictionary for when only the country is specified, and then finally there's a global fallback to dd/mm/yyyy.

vsgoulart: Ah, sorry. I missed the `navigator.language` and only focused on the `locale`.
But anyway, I think we shouldn't keep this dictionary on `form-js`, very likely it's incomplete and it could easily get out of date form some locales.
I think we could use the browser `Intl` API to get the correct format (since we're already using it to get the locale/language)
We could try the functions below (slightly modified from StackOverflow):
```js
function getDateFormatString(lang = 'default') {
  const parts = new Intl.DateTimeFormat(lang).formatToParts(new Date());

  return parts
    .map(part => {
      switch (part.type) {
        case "day":
          return "DD";
        case "month":
          return "MM";
        case "year":
          return "YYYY";
        default:
          return part.value;
      }
    })
    .join("");
}
```

PR #766, opened 2023-08-10
style: do not show empty state while dragging

pinussilvestrus:
I think it doesn't make sense to show during dragging. Let me know what you think.

We can also introduce another flag for dragging mode, but as we already set this one, I thought that would be enough.


vsgoulart (approved):
It makes sense to me :+1:

PR #772, opened 2023-08-14
deps: set renovate reviewers

pinussilvestrus:
Auto-assign reviews from the team @bpmn-io/hto-dev


pinussilvestrus:
In bpmn-io it's only us three in this group currently. For the camunda orga I want to create a new one for the Frontend devs, if this settles.

PR #799, opened 2023-09-14
chore: Update Node

vsgoulart:
[Node v16 reached end of life](https://github.com/nodejs/release#release-schedule)


pinussilvestrus:
@Skaiir @vsgoulart I was able to fix the three test cases by using `act`. It could be that bigger forms take longer in some environments to import so it makes sense to wait for it.

pinussilvestrus:
pinging @bpmn-io/modeling-dev that we're targetting Node@18 now for our CI, with the reasons mentioned above (End of Life).

nikku:
Any reason you don't switch to Node@20 right away? I guess that is what we'd attempt to do, to safe some effort in 1 year.

nikku:
Created a [tracking issue on our end](https://github.com/bpmn-io/internal-docs/issues/818), thanks for the heads-up @pinussilvestrus :medal_sports:

pinussilvestrus:
> Any reason you don't switch to Node@20 right away? I guess that is what we'd attempt to do, to safe some effort in 1 year.

No real reason, we will go to 20 directly in case we see no blockers 👍 thanks!

PR #811, opened 2023-09-21
fix(use-values): update values when evaluated values changed

### LINKED ISSUE #809: Options defined through feel expressions aren't dynamic
### Describe the Bug

firefox_H4t38wCPEc

As seen above, changing the isGood variable does not affect the select's values, which is unexpected

### Steps to Reproduce

1. add a checkbox with key `isGood`
2. add a select with options expression `if isGood then [ "good"] else ["bad"]`
3. toggle the checkbox and check the select values
4. nothing changes

### Expected Behavior

The select's values should be dynamically evaluated. 

### Environment

- Host Web modeler saas

pinussilvestrus:
Closes #809

This makes sure we re-calculate the values everytime the evaluated values from a `valuesExpression` changed.

### DIFF
--- a/packages/form-js-viewer/src/render/hooks/useValuesAsync.js
+++ b/packages/form-js-viewer/src/render/hooks/useValuesAsync.js
@@ -1,7 +1,8 @@
-import { useEffect, useMemo, useState } from 'preact/hooks';
+import { useEffect, useState } from 'preact/hooks';
 import { normalizeValuesData } from '../components/util/valuesUtil';
 import useExpressionEvaluation from './useExpressionEvaluation';
 import useService from './useService';
+import usePrevious from './usePrevious';
 
 /**
  * @enum { String }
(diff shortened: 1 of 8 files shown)


Skaiir (changes requested):
Unfortunately there are some issues here, cannot change option source to expression without crashing the form editor.

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
```diff
@@ -35,11 +33,8 @@ export default function(field) {
   const [ valuesGetter, setValuesGetter ] = useState({ values: [], error: undefined, state: LOAD_STATES.LOADING });
   const initialData = useService('form')._getState().initialData;
 
-  const evaluatedValues = valuesExpression ? (useExpressionEvaluation(valuesExpression) || []) : [];
-
-  // re-calculate values if evaluated expression changes
-  const previousEvaluatedValues = usePrevious(evaluatedValues, [], [ evaluatedValues ]);
-  const evaluatedValuesChanged = !compare(previousEvaluatedValues, evaluatedValues);
+  const expressionEvaluation = useExpressionEvaluation(valuesExpression);
```
Skaiir, line 36: I fixed our hook usage here, getting rid of the condition (hooks should not be conditionally evaluated, this confuses react/preact).

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
```diff
@@ -35,11 +33,8 @@ export default function(field) {
...
+  const evaluatedValues = useDeepCompareState(expressionEvaluation || [], []);
```
Skaiir, line 37: I abstracted away all that funky compare logic into its own hook, makes everything cleaner. Plus we may want to re-use this deep comparison stuff other places, for example here:

grafik

Skaiir (approved):

Skaiir:
@pinussilvestrus Approved by me, but I put in some additional changes. Feel free to merge if you like em.

packages/form-js-viewer/src/render/hooks/useValuesAsync.js
pinussilvestrus: Thanks a lot 🏅

PR #827, opened 2023-09-28
fix(viewer): clean up errors for hidden fields

### LINKED ISSUE #825: If a field is hidden, it should NOT be validated
### Describe the Bug

Hidden fields are validated even though there is no way to change their values.

### Steps to Reproduce

1. Open this form:

```json
{
  "components": [
    {
      "label": "Checkbox",
      "type": "checkbox",
      "layout": {
        "row": "Row_0ifu9f7",
        "columns": null
      },
      "id": "Field_0r8ir96",
      "key": "checkbox_4u82gk"
    },
    {
      "label": "Number",
      "type": "number",
      "layout": {
        "row": "Row_0ifu9f7",
        "columns": null
      },
      "id": "Field_17uk1c9",
      "key": "number_o4f027",
      "conditional": {
        "hide": "=checkbox_4u82gk"
      },
      "validate": {
        "required": true
      }
    },
    {
      "action": "reset",
      "label": "Reset",
      "type": "button",
      "id": "Field_1jcm8tv",
      "layout": {
        "row": "Row_0ifu9f7"
      }
    },
    {
      "action": "submit",
      "label": "Button",
      "type": "button",
      "layout": {
        "row": "Row_15idciq",
        "columns": null
      },
      "id": "Field_0bqqn6m"
    }
  ],
  "type": "default",
  "id": "Form_0k71hbi",
  "exporter": {},
  "schemaVersion": 11
}
```

2. Tick the checkbox
3. Submit and read the errors
4. `Field_17uk1c9` appears in the errors, even though it's hidden.

### Expected Behavior

Field should not be validated if it's hidden.

### Environment

- OS: [e.g. Windows 7]

---

Reported via forums: https://forum.camunda.io/t/required-validation-on-hidden-inputs/47782

pinussilvestrus:
Closes #825

I actually wonder whether this worked beforehand, as conditions never refer to form field IDs, but their respective key.

Seems like we had a lack of test coverage over here. Feel free to check @Skaiir!

### DIFF
--- a/packages/form-js-viewer/src/Form.js
+++ b/packages/form-js-viewer/src/Form.js
@@ -177,11 +177,10 @@ export default class Form {
     const data = this._getSubmitData();
 
     const errors = this.validate();
-    const filteredErrors = this._applyConditions(errors, data);
 
     const result = {
       data,
-      errors: filteredErrors
+      errors
     };
 
     this._emit('submit', result);
@@ -208,6 +207,8 @@ export default class Form {
 
     const { data } = this._getState();
 
+    const getErrorPath = (field) => [ field.id ];
+
     const errors = formFieldRegistry.getAll().reduce((errors, field) => {
       const {
         disabled
@@ -221,12 +222,14 @@ export default class Form {
 
       const fieldErrors = validator.validateField(field, value);
 
-      return set(errors, [ field.id ], fieldErrors.length ? fieldErrors : undefined);
+      return set(errors, getErrorPath(field), fieldErrors.length ? fieldErrors : undefined);
     }, /** @type {Errors} */ ({}));
 
-    this._setState({ errors });
+    const filteredErrors = this._applyConditions(errors, data, { getFilterPath: getErrorPath });
 
-    return errors;
+    this._setState({ errors: filteredErrors });
+
+    return filteredErrors;
   }
 
   /**
@@ -433,9 +436,9 @@ export default class Form {
   /**
    * @internal
    */
-  _applyConditions(toFilter, data) {
+  _applyConditions(toFilter, data, options = {}) {
     const conditionChecker = this.get('conditionChecker');
-    return conditionChecker.applyConditions(toFilter, data);
+    return conditionChecker.applyConditions(toFilter, data, options);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


Skaiir:
My changes include:

Filtering of errors is done in the validation method directly. The error state is changed within it, it doesn't make sense to me that the error state should be any different (old bug, I've extended the test case to handle this too).
I've also made conditionChecker default to using a path, but also allowing it to have a specific path function. That way when we want to filter data on a basis other than the object's value path (such as id), we can specify the path function externally.

packages/form-js-viewer/src/Form.js
```diff
@@ -221,12 +222,14 @@ export default class Form {
...
       const fieldErrors = validator.validateField(field, value);
 
-      return set(errors, [ field.id ], fieldErrors.length ? fieldErrors : undefined);
+      return set(errors, getErrorPath(field), fieldErrors.length ? fieldErrors : undefined);
     }, /** @type {Errors} */ ({}));
 
-    this._setState({ errors });
+    const filteredErrors = this._applyConditions(errors, data, { getFilterPath: getErrorPath });
+
+    this._setState({ errors: filteredErrors });
```
pinussilvestrus, line 230: Just wondering whether this is a breaking change for `validate`, but I believe we're indeed fixing something here, the behavior beforehand was broken.

packages/form-js-viewer/src/features/expression-language/ConditionChecker.js
```diff
@@ -19,11 +19,17 @@ export default class ConditionChecker {
...
+   * @param {Object} [options]
+   * @param {Function} [options.getFilterPath]
    */
-  applyConditions(properties, data = {}) {
+  applyConditions(properties, data = {}, options = {}) {
 
     const newProperties = clone(properties);
 
+    const {
+      getFilterPath = (field) => this._pathRegistry.getValuePath(field)
```
pinussilvestrus, line 30: 👍

packages/form-js-viewer/src/Form.js
pinussilvestrus: Maybe we can add a test case for `#validate` that verifies this behavior as well?

pinussilvestrus (commented):
Thanks for the changes! Only one potential test case to add, feel free to merge as you like 👍

packages/form-js-viewer/src/Form.js
pinussilvestrus: added it ✅

PR #828, opened 2023-10-02
fix: outline styles for properties panel

### LINKED ISSUE #664: Wrong outline on properties panel on Linux
### Describe the Bug

In some properties panel fields the focus outline is wrong on Linux machines (when using Firefox and Chrome)

### Steps to Reproduce

1. Tab through the properties panel in the playground
2. The outline color is wrong

### Expected Behavior

The outline should match our styles

### Environment

https://github.com/bpmn-io/form-js/assets/1022916/4e6df489-1c09-48c7-8490-166e7cd3493a

- Host (Browser/Node version), if applicable: Firefox, Chrome
- OS: Archlinux

kerlon5:
Closes #664

### DIFF
--- a/packages/form-js-editor/assets/form-js-editor-base.css
+++ b/packages/form-js-editor/assets/form-js-editor-base.css
@@ -194,6 +194,8 @@
   --feel-required-color: var(--cds-icon-disabled, var(--color-grey-225-10-55));
 
   --feel-indicator-background-color: var(--cds-background-hover, var(--color-grey-225-10-90));
+  
+  --outline-definition: 1px solid var(--cds-focus, var(--input-focus-border-color));
 }
 
 .fjs-properties-container .bio-properties-panel-input {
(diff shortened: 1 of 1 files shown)


pinussilvestrus:
@kerlon5 Thanks a lot for your contribution ❤️ 

@bpmn-io/hto-dev I'm wondering what the best solution here would be. Is that something we need to fix in form-js or rather in the [properties panel core library](https://github.com/bpmn-io/properties-panel) to harmonize the focus styles?

packages/form-js-editor/assets/form-js-editor-base.css
```diff
@@ -675,6 +677,10 @@
   box-sizing: border-box;
 }
 
+.fjs-properties-container .fjs-properties-panel *:focus-visible {
+  outline: var(--outline-definition);
```
pinussilvestrus, line 681: In the core styles, we do not show any outline on purpose, but indeed [show a border on focus](https://github.com/bpmn-io/properties-panel/blob/main/src/assets/properties-panel.css#L512). Adding this outline would lead to a too big border.

image

We should rather only set the border color here.

pinussilvestrus (changes requested):
As mentioned via https://github.com/bpmn-io/form-js/pull/828#issuecomment-1748372057 I think this should rather be fixed in the core properties panel styles - if we want that at all.

pinussilvestrus:
Let's close this one, as we do not intend to fix this here. We accept pull requests in the [properties panel core](https://github.com/bpmn-io/properties-panel) as well.

PR #829, opened 2023-10-02
fix: readonly number fields should not show controls

kerlon5:
Closes #

### DIFF
--- a/packages/form-js-viewer/assets/form-js-base.css
+++ b/packages/form-js-viewer/assets/form-js-base.css
@@ -599,6 +599,10 @@
   color: var(--cds-text-inverse, var(--color-text));
 }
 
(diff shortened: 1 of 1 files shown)


pinussilvestrus (commented):
@kerlon5 thanks a lot for your contribution!

What I'm wondering is that the number field is not distinguishable now with text fields in the read-only state. Also in in read-only, it might be interesting _what data type_ this field has, although I can't edit it.

image


@bpmn-io/hto-dev @volodymyr-melnykc what do you think?

volodymyr-melnykc:
In the read-only mode, we don't need to show controls as the user can't interact with it, so I agree with the fix.

If an input field has a "prefix" or "suffix" defined, we should still display it when in read-only. I believe that this remains the case (based on code updates).

@kerlon5 @pinussilvestrus

PR #835, opened 2023-10-11
fix: ensure values not in options clear

### LINKED ISSUE #817: Properly cleanup select-like components values when options change at form runtime
### Describe the Bug

Currently, when the options of a field change at form runtime, but a value has already been set for them, the value is kept in the form state despite now not being valid anymore, see below. 

grafik

This is particularly obvious with taglists which have a failsafe in place for when that happens, but occurs with every option consuming field.

### Steps to Reproduce

1. Create a checklist/taglist/select or radio component
2. Create a checkbox with key isChecked
3. Set the expression options for the first component to something like `if isChecked then someArray else someOtherArray`, both arrays should have no overlap in values
4. Set a value on the first component
5. Check the checkbox
6. See the value is still present in the output data. In the case of taglist components this is very obvious.

### Expected Behavior

6. The value should be cleaned up, as we shouldn't allow a user to get in such an unsupported data state

If the value is not present in options, then the value is reset to null. Or on the case of multiselects, it should be filtered out and the state should be refreshed.

We should probably get rid of that failsafe as well, ideally by making it impossible for that failsafe state to be reached in the first place.

Skaiir:
Closes #817


packages/form-js-viewer/src/render/hooks/useOptionsAsync.js
```diff
@@ -0,0 +1,78 @@
+import { useEffect, useState } from 'preact/hooks';
```
pinussilvestrus, line 1: We didn't export this hook, right?

Skaiir: Oh I didn't think about it but no we didn't. Luuuckyyy.

pinussilvestrus: People might ask about it once custom components are released. Once we do it, we have to be more careful about renaming 👍

packages/form-js-viewer/src/render/hooks/useRestrictedMultiSelectValues.js
```diff
@@ -0,0 +1,32 @@
...
+
+  const {
+    field,
+    options,
+    loadState,
+    onChange,
+    values
+  } = props;
+
+  // Ensures that the values are always a subset of the options
```
pinussilvestrus, line 14:
```suggestion
  // Ensures that the values are always a subset of the possible options
```

packages/form-js-viewer/src/render/components/form-fields/parts/SimpleSelect.js
```diff
@@ -32,9 +33,17 @@ export default function SimpleSelect(props) {
...
+    options
+  } = useOptionsAsync(field);
+
+  useRestrictedSingleSelectValue({
+    field,
+    loadState,
+    options,
+    value,
+    onChange: props.onChange
+  });
```
pinussilvestrus, line 46: This looks a bit like black magic. Looking inside the hook reveals it's simply another effect, but maybe we can improve the readability a bit to indicate we might updating the values?

Maybe something like `useCleanupSingleSelectValue`, something that is signalizing an action.

packages/form-js-viewer/src/render/hooks/useRestrictedMultiSelectValues.js
```diff
@@ -0,0 +1,32 @@
...
+    if (loadState !== LOAD_STATES.LOADED) {
+      return;
+    }
+
+    const hasValuesNotInOptions = values.some(v => !options.map(o => o.value).includes(v));
+
+    if (hasValuesNotInOptions) {
+      onChange({
+        field,
+        value: values.filter(v => options.map(o => o.value).includes(v))
```
pinussilvestrus, line 26: As in the comment above having updates here might be unexpected. What would be the disadvantage of only sanitizing `value` here and not updating it directly? I guess it would lead to unexpected side effects as we never really clean things up?

packages/form-js-viewer/src/render/hooks/useRestrictedMultiSelectValues.js
```diff
@@ -0,0 +1,32 @@
+import { useEffect } from 'preact/hooks';
```
pinussilvestrus, line 1: I see no test added for this new behavior. Did I miss something?

pinussilvestrus (commented):
Some things to clarify, but in general, it looks like a nice direction to keep things clean 👍 

Demo also works ✅

packages/form-js-viewer/src/render/hooks/useRestrictedMultiSelectValues.js
Skaiir: Yeah because then you have a mismatch between the form and the actual state. So if you have lets say, a feel expression dependent on this value, it'll still evaluate despite the form not showing that value being selected.

packages/form-js-viewer/src/render/hooks/useRestrictedMultiSelectValues.js
Skaiir: There be tests now

pinussilvestrus (approved):

pinussilvestrus:
Thanks for the follow up 👍

PR #836, opened 2023-10-11
fix: move dynamic options hint to tooltip

### LINKED ISSUE #305: Explanatory tooltip for new dynamic input options

Skaiir:
Closes #305

There's a mild styling issue which are going to be fixed via https://github.com/bpmn-io/properties-panel/pull/300/commits, but since we're eventually going to bump the properties panel and the visuals aren't that big a deal, I feel like it's not worth monkey patching or syncing. If you think we should still wait lmk.

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/InputKeyValuesSourceEntry.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/InputKeyValuesSourceEntry.js
@@ -11,20 +11,10 @@ export default function InputKeyValuesSourceEntry(props) {
     id
   } = props;
 
-  const schema = '[\n  {\n    "label": "dollar",\n    "value": "$"\n  }\n]';
-
-  const description = <div>
-    Define which input property to populate the values from.
-    <br /><br />The input property may be an array of simple values or alternatively follow this schema:
-    <pre><code>{schema}</code></pre>
-  </div>;
-
   return [
     {
       id: id + '-key',
       component: InputValuesKey,
-      label: 'Input values key',
-      description,
       isEdited: isTextFieldEntryEdited,
       editField,
       field,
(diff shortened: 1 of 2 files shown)


packages/form-js-editor/src/features/properties-panel/entries/InputKeyValuesSourceEntry.js
```diff
@@ -11,20 +11,10 @@ export default function InputKeyValuesSourceEntry(props) {
...
-    Define which input property to populate the values from.
-    <br /><br />The input property may be an array of simple values or alternatively follow this schema:
-    <pre><code>{schema}</code></pre>
-  </div>;
-
   return [
     {
       id: id + '-key',
       component: InputValuesKey,
-      label: 'Input values key',
```
pinussilvestrus, line 26: Removing the label from the `entries` configuration would make it impossible to override via extension (custom properties provider) I think?

Not sure whether this is a widely used pattern in our entries, though.

Skaiir: I looked around and it's one or the other. I mostly changed this one to be in line with the Expressions equivalent which did define the label on the component itself.

Skaiir: I think when we know that this matters, we can fix it. Because there's a lot of inconsistency at the moment.

pinussilvestrus: If it's not widely used, we can consider this as not-supported at the moment.

Happy to clean this up at all places in one of the next spring cleanings 🧹

pinussilvestrus (approved):

Issue #839, opened 2023-10-12
Document all events


Skaiir:
### What should we do?

We haven't been very good with keeping the readme's updated with all events we support, let's update this.

### Why should we do it?

To make it easier for anyone implementing our library.

Skaiir:
https://github.com/bpmn-io/form-js/issues/526 included

PR #838, opened 2023-10-12
feat(taglist): add skip to search

pinussilvestrus:
This is to provide a skip link to quickly navigate to the search via keyboard. This becomes useful once you have a lot of tags selected and want to quickly navigate to the search bar.

Kapture 2023-10-12 at 15 05 56

In general, skip links are a widely used pattern to skip parts of the content, only accessible for keyboard users and screenreaders.

Cf. https://webaim.org/techniques/skipnav/#hidden

Closes #435

PR #841, opened 2023-10-13
Add focus and blur events + viewer event documentation

### LINKED ISSUE #60: Make it possible to listen to on focus events

On Zeebe a user might have a variable of up to 4MB and because of that there might be cases on Zeebe Tasklist that a task has multiple big  variables as initial values on the form which might cause the request to take too long to load or timeout.
To solve that we want to load only a truncated value of each requested variable and only load the full value of the variable when the user focuses on the field

Create a new event that's dispatched when the user focuses on the field:
```js
/**
  * @type { (event: 'focus', callback: (state: { field: {name: string, type: 'textfield' | 'number' | ... } } ) => void) => void }
  */
form.on('focus', ({ field }) => {
  // we can filter by field type so we don't fetch unnecessary variables, like boolean variables for example
  if(field.type === 'textfield') {
    request(field.name)
  }
})
```

__Context__

The idea is to use this feature along with #39 and #67 to implement lazy loading of data.

Skaiir:
Closes #60
Related to #839


packages/form-js-viewer/README.md
```diff
@@ -154,9 +154,33 @@ Remove form from the document.
 
 ## Events
 
-The form emits the `changed` and `submit` events you may hook into.
+### `changed :: { data, errors }`
+Fired off every time there is a form state change.
 
-Both events receive `{ data, errors }` as a payload.
+### `submit :: { data, errors }`
```
pinussilvestrus, line 160: Interesting, short format to visualize this! Do we have any other place where we use this syntax or how did you come up with it?

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -49,7 +50,8 @@ export default function Taglist(props) {
   const [ isDropdownExpanded, setIsDropdownExpanded ] = useState(false);
   const [ hasOptionsLeft, setHasOptionsLeft ] = useState(true);
   const [ isEscapeClosed, setIsEscapeClose ] = useState(false);
-  const searchbarRef = useRef();
```
pinussilvestrus, line 52: You probably have to rebase to `develop` as we used this one for the new `skipLink` to the search bar (cf. #838).

packages/form-js-viewer/test/spec/Form.spec.js
```diff
@@ -474,6 +476,100 @@ describe('Form', function() {
   });
 
 
+  describe('focus and blur events should trigger for', async function() {
+
+    const runFocusBlurTest = async function(id, index, selector) {
```
pinussilvestrus, line 481: I like this! To even save more duplication, we can do something[ like in other places](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/test/spec/FormEditor.spec.js#L1342)

```js
const runFocusBlurTest = function(id, index, selector) {
  it('focus and blur events should trigger for ' + id, async function() {
    
    // given
    // ...
  });
}

runFocusBlurTest('number', 0, 'input');

runFocusBlurTest('date', 1, 'input');
```

packages/form-js-viewer/test/spec/render/components/FormField.spec.js
```diff
@@ -316,6 +316,7 @@ describe('FormField', function() {
           ...defaultField,
         },
         setState: setStateSpy,
+        onBlur: () => {},
```
pinussilvestrus, line 319: Shouldn't this be optional here as well?

packages/form-js-viewer/README.md
Skaiir: To be fair there isn't a standard in place yet. Although, I didn't want to fill the documentation with a lot of events and descriptions, especially if we do the right thing and do more event driven stuff in the future. That's also why I only add descriptions when the event name doesn't make it entirely obvious.

We can always ask for GPT's opinion on this, while keeping the constraints of the compact format.

Skaiir: GPT didn't seem to mind :D

packages/form-js-viewer/README.md
```diff
@@ -154,9 +154,33 @@ Remove form from the document.
...
+- `attach`
+- `form.init`
+- `form.clear`
+- `form.destroy`
+- `diagram.clear`
+- `diagram.destroy`
+
+### `<formFieldEvent> :: { formField }`
+- `formField.add` 
+- `formField.remove`
```
pinussilvestrus, line 181: These are editor events, correct? Maybe we can add the full event documentation to the general `form-js` README?

Skaiir: Actually no these events are bound to the form field registry. So they technically are viewer events, they would fire during import.

Skaiir: I think it's still good to separate them, as you access them differently. form.on and formEditor.on.

Skaiir: You can programmatically add fields to the viewer in theory, there may be other use cases we haven't thought of that do this. For example, a streamed version of the form generator using only a viewer.

pinussilvestrus: Good catch!

pinussilvestrus (approved):

PR #845, opened 2023-10-16
chore: document editor events

### LINKED ISSUE #60: Make it possible to listen to on focus events

### LINKED ISSUE #839: Document all events

Skaiir:
Closes #839

Added documentation of the editor events. Still quite simple for now, we can judge if we need to get into more detail later.
Also adjusted the viewer events to put the format individually, as I noticed the next event we need for form-fields doesn't match the usual format.

### DIFF
--- a/packages/form-js-editor/README.md
+++ b/packages/form-js-editor/README.md
@@ -110,6 +110,42 @@ Subscribe to an [event](#events).
 
 Remove form from editor the document.
 
+## Events
+
+### `selection.changed :: { selection }`
+
+### Properties panel events
+- `propertiesPanel.focusin`
+- `propertiesPanel.focusout`
+- `propertiesPanel.showEntry :: { id }`
+- `propertiesPanel.updated :: { formField }`
+
+### Form lifecycle events
+- `detach`
+- `attach`
+- `rendered`
+- `form.init`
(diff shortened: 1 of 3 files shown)


vsgoulart (approved):
LGTM

pinussilvestrus (approved):

PR #884, opened 2023-11-06
fix(playground): add `tabIndex` to json editors

pinussilvestrus:
Related to https://github.com/camunda/team-hto/issues/385

### DIFF
--- a/packages/form-js-playground/src/components/PlaygroundRoot.js
+++ b/packages/form-js-playground/src/components/PlaygroundRoot.js
@@ -91,14 +91,14 @@ export function PlaygroundRoot(props) {
   useEffect(() => {
     const dataEditor = dataEditorRef.current = new JSONEditor({
       value: toString(data),
-      contentAttributes: { 'aria-label': 'Form Input' },
+      contentAttributes: { 'aria-label': 'Form Input', tabIndex: 0 },
       placeholder: createDataEditorPlaceholder()
     });
 
     const resultView = resultViewRef.current = new JSONEditor({
       readonly: true,
       value: toString(resultData),
-      contentAttributes: { 'aria-label': 'Form Output' }
+      contentAttributes: { 'aria-label': 'Form Output', tabIndex: 0 }
     });
 
     const form = formRef.current = new Form({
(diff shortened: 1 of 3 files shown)


pinussilvestrus:
Added the suggested fix from https://discuss.codemirror.net/t/is-there-a-way-to-configure-the-tabindex-of-the-textbox-without-hacking-the-source/5111/2

packages/form-js-playground/src/components/PlaygroundRoot.js
```diff
@@ -91,14 +91,14 @@ export function PlaygroundRoot(props) {
   useEffect(() => {
     const dataEditor = dataEditorRef.current = new JSONEditor({
       value: toString(data),
-      contentAttributes: { 'aria-label': 'Form Input' },
+      contentAttributes: { 'aria-label': 'Form Input', tabIndex: 0 },
```
vsgoulart, line 94: isn't it better to use `undefined` as default?

packages/form-js-playground/test/spec/Playground.spec.js
```diff
@@ -790,6 +791,60 @@ describe('playground', function() {
 
   });
 
+
+  describe('a11y', function() {
+
+    it('should have no violations', async function() {
+
+      // given
+      this.timeout(5000);
```
vsgoulart, line 800: Why is the timeout necessary?

vsgoulart (approved):

packages/form-js-playground/test/spec/Playground.spec.js
pinussilvestrus: `axe` takes a while to run its validations, leading `karma` to run out of time sometimes (flaky).

packages/form-js-playground/src/components/PlaygroundRoot.js
pinussilvestrus: I think the creators of Code Mirror suggested a non-undefined value as the parent `.cm-scroller` container is `-1`. This is not configurable so we need to set the `cm-content` container explicitly.

image

This is where the a11y error comes from in the first place

image

PR #896, opened 2023-11-13
chore: Bump `schemaVersion` to 13

### LINKED ISSUE #889: Extend schema to support new table component

vsgoulart:
Closes #889

### DIFF
--- a/packages/form-json-schema/README.md
+++ b/packages/form-json-schema/README.md
@@ -46,11 +46,11 @@ npm run test
 
 ## Schema compatibility notice
 
-This schena is currently only compatible with following [form-js](https://github.com/bpmn-io/form-js) schema versions.
+This schema is currently only compatible with following [form-js](https://github.com/bpmn-io/form-js) schema versions.
 
 | JSON schema version | form-js schema version |
 |---|---|
-| >= 0.2  | <= 12 |
+| >= 0.2  | <= 13 |
 
 ## License
 
--- a/packages/form-json-schema/README.md
+++ b/packages/form-json-schema/README.md
@@ -50,7 +50,7 @@ This schema is currently only compatible with following [form-js](https://github
 
 | JSON schema version | form-js schema version |
 |---|---|
-| >= 0.2  | <= 13 |
+| >= 1.5.0  | <= 13 |
 
 ## License
 
--- a/packages/form-json-schema/src/defs/component.json
+++ b/packages/form-json-schema/src/defs/component.json
@@ -227,6 +227,21 @@
     "properties": {
       "$id": "#/component/properties",
       "$ref": "properties.json"
+    },
+    "columns": {
+      "$id": "#/component/columns",
+      "$ref": "columns.json"
+    },
+    "columnsExpression": {
+      "$id": "#/component/columnsExpression",
+      "description": "Table columns can be configured with a FEEL expression defining the individual columns the table will render.",
+      "type": "string"
+    },
+    "rowCount": {
+      "$id": "#/component/rowCount",
+      "description": "Define the amount of rows a table will render.",
+      "type": "number",
+      "default": 10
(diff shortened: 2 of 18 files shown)


packages/form-json-schema/README.md
```diff
@@ -46,11 +46,11 @@ npm run test
 
 ## Schema compatibility notice
 
-This schena is currently only compatible with following [form-js](https://github.com/bpmn-io/form-js) schema versions.
+This schema is currently only compatible with following [form-js](https://github.com/bpmn-io/form-js) schema versions.
 
 | JSON schema version | form-js schema version |
 |---|---|
-| >= 0.2  | <= 12 |
+| >= 0.2  | <= 13 |
```
pinussilvestrus, line 53: We will need to change this matrix anyway as we follow the form-js release versions now.

packages/form-json-schema/src/defs/examples/components.json
```diff
@@ -143,5 +143,13 @@
         "components": []
       }
     ]
+    ,
+    [
```
pinussilvestrus, line 147: Is this valid JSON? 😄

packages/form-json-schema/test/spec/validation.spec.js
```diff
@@ -257,6 +257,15 @@ describe('validation', function() {
...
     testForm('defaultValue-no-array');
 
+
+    testForm('columns-not-allowed');
+
+    testForm('columnsExpression-not-allowed');
+
+    testForm('rowCount-not-allowed');
+
+    testForm('rowCount-no-number');
```
pinussilvestrus, line 267:
```suggestion
    testForm('columns-not-allowed');
    

    testForm('columnsExpression-not-allowed');
    

    testForm('rowCount-not-allowed');
    

    testForm('rowCount-no-number');
```

packages/form-json-schema/src/defs/component.json
```diff
@@ -227,6 +227,21 @@
     "properties": {
       "$id": "#/component/properties",
       "$ref": "properties.json"
+    },
+    "columns": {
+      "$id": "#/component/columns",
+      "$ref": "columns.json"
+    },
+    "columnsExpression": {
```
pinussilvestrus, line 235: We could consider making these two exclusive, so only one can exist in the schema.

However for `values`, `valuesKey` and `valuesExpression` we don't have such a rule as we anyway pick one and ignore the others.

vsgoulart: The message is not very clear, I don't know if there's a better way of doing than the way I did

image

packages/form-json-schema/src/defs/examples/components.json
vsgoulart: it is, but I improved it anyway

packages/form-json-schema/src/defs/component.json
pinussilvestrus: In the element-templates-json-schema we have a mechanism to provide custom error messages: https://github.com/camunda/element-templates-json-schema/blob/f06ab6b42a32904eff97688294bd9ad4a994ffe3/packages/zeebe-element-templates-json-schema/src/error-messages.json#L29

We'd have the infrastructure already in place in form-js; the task for generating the errors is there. 

However, these messages are rather for our Desktop Modeler tooling. Monaco / VSCode supports the `errorMessage` keyword, but not with string literals, that's why we separate them. What you could try out is using the `errorMessage` keyword directly somewhere in the rule with a static string.

vsgoulart: Done

image

pinussilvestrus: Nice one!

pinussilvestrus (approved):

PR #898, opened 2023-11-16
iFrame component

### LINKED ISSUE #887: [Implement] Support iFrame component
> This is the implementation epic for https://github.com/camunda/product-hub/issues/1897

### What should we do?

We want to create an `iFrame` component that allows users to specify an external website to be rendered in a form (first iteration). We consider expanding that by providing raw HTML to render into an iFrame, but this is part of a second iteration - and considered as part of another component (cf. https://github.com/camunda/product-hub/issues/1897).

This includes
- [x] Support `iFrame` component that renders a given URL
- [x] Add placeholders for editor
- [x] Allow to specify `iFrame` attributes
  - [x] Url: FEEL and Template support
  - [x] Label: FEEL and Template support
  - [x] Height
  - [x] <del>Sandbox attributes</del>
- [x] Ensure schema version is updated
- [x] Ensure carbon styles are updated
- [x] Ensure JSON schema is updated
- [x] https://github.com/camunda/form-linting/issues/94
- [x] https://github.com/camunda/team-hto/issues/414
- [ ] Integrate updates into products
  - [ ] Tasklist
  - [ ] Desktop Modeler
  - [ ] Web Modeler
  - [ ] Zeebe Play

### Why should we do it?

Reduce the feature gap for special use cases. Examples: Render a Google Doc, render a PDF, etc.

An iFrame component also opens up a lot of potential form fields that we can't deliver on our own.

Cf. to the [product hub](https://github.com/camunda/product-hub/issues/1897) to gather the full rationale.

### Additional context

Related to https://github.com/camunda/product-hub/issues/1897
Related to https://github.com/camunda/team-hto/issues/393
Related to [custom components roadmap](https://miro.com/app/board/uXjVNWfBoyE=/)
Design: https://www.figma.com/file/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?type=design&node-id=465-18970&mode=design&t=uDfUcnzmaWQKvUwz-0
Miro: https://miro.com/app/board/uXjVNQg-WL4=/
Context: https://www.w3schools.com/html/html_iframe.asp
Context: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

pinussilvestrus:
Related to #887

Adds `iFrame` component.
* Specify `url` to embed web content into a form
* Specify title via `label`
* Specify height

Demo: https://demo-887-iframe-component--camunda-form-playground.netlify.app/


volodymyr-melnykc:
@pinussilvestrus The iframe (svg) icon looks too big. I see that it doesn't have inner paddings. Can you please update it to this one?

iframe icon

packages/form-js-viewer/src/render/components/Sanitizer.js
```diff
@@ -45,6 +45,7 @@ const ALLOWED_ATTRIBUTES = [
 
 const ALLOWED_URI_PATTERN = /^(?:(?:(?:f|ht)tps?|mailto|tel|callto|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i; // eslint-disable-line no-useless-escape
 const ALLOWED_IMAGE_SRC_PATTERN = /^(https?|data):.*/i; // eslint-disable-line no-useless-escape
+const ALLOWED_IFRAME_SRC_PATTERN = /^(https?):\/\/*/i; // eslint-disable-line no-useless-escape
```
pinussilvestrus, line 48: @volodymyr-melnykc @christian-konrad @vsgoulart @Skaiir We could also update this pattern to only allow for `https`, which is recommended anyways.

Wdyt?

pinussilvestrus:
@volodymyr-melnykc I updated the icon 👍

packages/form-js-viewer/src/render/components/Sanitizer.js
vsgoulart: Yeah, I think thatś a good idea

pinussilvestrus:
TIL: We need to set the `sandbox` attribute explicitly, "empty" is not empty enough.

image

Thanks to the InfoSec team ❤️

packages/form-js-viewer/src/render/components/form-fields/IFrame.js
```diff
@@ -0,0 +1,78 @@
...
+    <Label id={ prefixId(id, formId) } label={ evaluatedLabel } />
+    {
+      safeUrl &&
+        <iframe
+          src={ safeUrl }
+          title={ evaluatedLabel }
+          height={ height }
+          class="fjs-iframe"
+          id={ prefixId(id, formId) }
+          sandbox=""
```
vsgoulart, line 53: @pinussilvestrus Maybe we should add test for this?

packages/form-js-viewer/src/render/components/Sanitizer.js
volodymyr-melnykc: Let's also be explicit to the user that we require a `https` URL. 
@pinussilvestrus I updated the copy for the tooltip. Can you please update it in the codebase?

image

pinussilvestrus: Updated it accordingly (preview still building) 👍

packages/form-js-viewer/src/render/components/form-fields/IFrame.js
pinussilvestrus: Good catch, added one test case.

pinussilvestrus:
> @pinussilvestrus Based on the [latest discussion with InfoSec](https://camunda.slack.com/archives/CKWSYH1S4/p1700129353262419) I updated the tooltip to be more descriptive: https://www.figma.com/file/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?type=design&node-id=465%3A20473&mode=design&t=4Sc34AF6MsKAnHS4-1

Updated it 👍

vsgoulart (approved):
Only issue I found was the PDF one we discussed via Slack which probably is not fixable from our side

Another small improvement we could make is to display the default height on the properties panel instead of hiding it, what do you think @pinussilvestrus ?

pinussilvestrus:
> Another small improvement we could make is to display the default height on the properties panel instead of hiding it, what do you think @pinussilvestrus ?

Yeah that makes sense 👍 I added the default value accordingly.

pinussilvestrus:
@esraagamal6 I think you can start testing the demo: https://demo-887-iframe-component--camunda-form-playground.netlify.app/. I don't expect any bigger change for now.

@volodymyr-melnykc @christian-konrad @Skaiir feel free to provide your feedback of course :)

Skaiir:
Might be worth having a pattern match or implicitly prefix https:// here: 
grafik

pinussilvestrus:
@volodymyr-melnykc Do you think the tooltip is enough or should we add a validation message below the static input? Personally I think auto-prefixing is something we should avoid.

Skaiir:
This is all take it or leave it opinion btw. The unloaded state feels a little bit technical on the viewer side. I don't know but a user will see this and see iFrame, cool but that might not mean anything to them and it's not explicit visually that something has gone wrong here. 

I'd expect something like, external site couldn't be fetched or something like that. Maybe good as a follow-up.

Wdyt?

grafik

Skaiir:
> Personally I think auto-prefixing is something we should avoid.

Yeah I agree actually, even if I suggested it :trollface:

pinussilvestrus:
> This is all take it or leave it opinion btw. The unloaded state feels a little bit technical on the viewer side. I don't know but a user will see this and see iFrame, cool but that might not mean anything to them and it's not explicit visually that something has gone wrong here.
> 
> I'd expect something like, external site couldn't be fetched or something like that. Maybe good as a follow-up.
> 
> Wdyt?
> 
> grafik

I think we could be a bit more descriptive if something went wrong. A cheap way would be to place the existing placeholder in the editor and when the source is empty + showing a different text when the URL was not valid after we went through our sanitization.

image

volodymyr-melnykc:
@pinussilvestrus When the URL is not provided, we can show a placeholder in the form preview:
https://www.figma.com/file/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?type=design&node-id=575%3A12427&mode=design&t=0TnKCDbS6b4xH7ex-1

Form builder - No iFrame URL

When the provided URL is not "https", it makes sense to show the validation message below the static input (as you suggested).
And if the content can't be loaded, we can show the message in the form preview.
https://www.figma.com/file/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?type=design&node-id=575%3A12815&mode=design&t=0TnKCDbS6b4xH7ex-1

Form builder - screen

pinussilvestrus:
@volodymyr-melnykc I updated it as you suggested 👍

packages/form-js-editor/src/features/properties-panel/entries/LabelEntry.js
```diff
@@ -51,7 +51,12 @@ export default function LabelEntry(props) {
       editField,
       field,
       isEdited: isFeelEntryEdited,
-      isDefaultVisible: (field) => INPUTS.includes(field.type) || field.type === 'button' || field.type === 'group'
+      isDefaultVisible: (field) => (
+        INPUTS.includes(field.type) ||
```
Skaiir, line 55: Might make sense to have a parameter for "labeled" at this point, but I don't know. Just a thought for the future, since this conditional is starting to get more and more arbitrary.

packages/form-js-editor/src/features/properties-panel/entries/LabelEntry.js
```diff
@@ -157,4 +162,20 @@ function TimeLabel(props) {
...
+// helpers //////////
+
+function getLabelText(field) {
+  const { type } = field;
+
+  if (type === 'group') {
+    return 'Group label';
+  }
+
+  if (type === 'iframe') {
```
Skaiir, line 176: Not sure this is super necessary to change the wording here, I presume there were discussions around this?

packages/form-js-editor/src/features/properties-panel/entries/factories/simpleBoolEntryFactory.js
```diff
@@ -8,6 +8,8 @@ export default function simpleBoolEntryFactory(options) {
     description,
     path,
     props,
+    getValue,
```
Skaiir, line 11: Might have to start renaming it to just boolEntryFactory, but I'm not against the change, less boilerplate is always good.

packages/form-js-editor/src/features/properties-panel/entries/factories/simpleBoolEntryFactory.js
```diff
@@ -1,5 +1,5 @@
 import { get } from 'min-dash';
-import { CheckboxEntry, isCheckboxEntryEdited } from '@bpmn-io/properties-panel';
+import { ToggleSwitchEntry, isToggleSwitchEntryEdited } from '@bpmn-io/properties-panel';
```
Skaiir, line 2: I've already made this change somewhere else, maybe dynamic list, anyways +1

packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
```diff
@@ -31,6 +31,7 @@ import {
 import schema from '../../form.json';
 import defaultValuesSchema from '../../defaultValues.json';
 import redundantValuesSchema from '../../redundantValues.json';
+import iframeSchema from '../../../../../form-js-viewer/test/spec/iframes.json';
```
Skaiir, line 34: I'm wondering if generally we should start centralizing our JSON files on a single project, instead of copying across the same JSON for the editor, viewer and playground.

packages/form-js-editor/src/render/components/editor-form-fields/EditorIFrame.js
```diff
@@ -1,13 +1,19 @@
...
 
 export default function EditorIFrame(props) {
   const { field } = props;
 
-  // remove url to display placeholder
-  return <IFrame { ...{ ...props, field: { ...field, url: null } } } />;
+  const Icon = iconsByType(field.type);
+
+  return (
+    <div class="fjs-iframe-placeholder">
```
Skaiir, line 13: Yeah I actually like this more, less inter-dependency.

packages/form-js-carbon-styles/src/carbon-styles.scss
```diff
@@ -1202,3 +1202,17 @@
     color: var(--cds-text-disabled);
   }
 }
+
+// iFrame styles ////////////
+
+.fjs-container {
```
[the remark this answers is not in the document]
pinussilvestrus: Yeah the thing is, theming in general form-js looks slightly different in comparison to Carbon styles, especially with the layers we have in Tasklist. I tried to use these colors in the base form-js styles as well, but it looks a bit odd with the existing background colors and a theme applied.

packages/form-js-editor/src/features/properties-panel/entries/LabelEntry.js
pinussilvestrus: We kinda misused this one, because technically it's not only a label but also the accessible iFrame title. So we decided to use the existing label for it as it works for both.

packages/form-js-editor/src/features/properties-panel/entries/factories/simpleBoolEntryFactory.js
pinussilvestrus: Let's say I got inspired 😅

packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
pinussilvestrus: Yeah I thought about a `form-js-testing` package already, could also give good guidance around testing forms in general.

vsgoulart: I had plans on doing that on Tasklist, I would just call it `mocks` instead of testing (unless we want to include some utility functions too)

pinussilvestrus:
QA Testing is done ✅ Merging it now 👍

PR #902, opened 2023-11-17
ci: Prevent auto-update of visual regression screenshots

vsgoulart:
I don't believe the auto-update of screenshots is working out, it just makes us ignore the visual regression CI run on PRs and if we there's problems when we merge master to develop or vice-versa it updates the baseline screenshots wrongly like in [here](https://github.com/bpmn-io/form-js/commit/648c5d1e3d49952111a13958330c5095905b70b5) (and this is not the first time I noticed an update that shouldn't happen)

We have the instructions on how to update the screenshots [here](https://github.com/bpmn-io/form-js/blob/develop/e2e/README.md), it takes less than 5 min to update so I believe it's not a blocker

### DIFF
--- a/.github/workflows/UPDATE_VISUAL_SNAPSHOTS.yml
+++ b/.github/workflows/UPDATE_VISUAL_SNAPSHOTS.yml
@@ -9,7 +9,7 @@ jobs:
     if: github.event.pull_request.state != 'closed' && (github.event.label.name == 'update-snapshots' || contains( github.event.pull_request.labels.*.name, 'update-snapshots'))
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.41.0
+      image: mcr.microsoft.com/playwright:v1.40.1
       options: --user 1001:1000
 
     steps:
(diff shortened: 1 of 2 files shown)


pinussilvestrus:
My two cents: in case we rely on automation - which should be the goal as much as possible! - we need to make sure it works reliably. If that's not the case, I agree doing the work manually saves time in the end.

What I also see it that our E2E setup works nicely on Mac, but it seems like there are some mitigations on Windows machines. Maybe @Skaiir can share his opinion here.

vsgoulart:
>My two cents: in case we rely on automation - which should be the goal as much as possible!

IMO the task is manual by nature, when it fails you'll have to download the assets from the CI and check it manually if the failure makes sense and then wait until it gets updated on master/develop and check if the update didn't mess up anything on the CI

If there's issue on Windows let's fix that, and not just ignore the tests

Skaiir:
So the issue with windows is this:

grafik

vsgoulart:
> It's creating a bunch of new snapshots, might be worth figuring out.

@Skaiir did you follow the instructions I linked? If you follow them and use Docker this won't happen

vsgoulart:
@Skaiir This PR should be ready for review again, sorry for taking so long

I added a workflow that updates the screenshots automatically when you add the label `update-snapshots`, this way we can also review the UI updates

douglasbouttell-camunda (approved):
This is great. My team at Amazon had so many issues with visual regressions and never had anything like this.

Skaiir (approved):
Elegant solution that makes us all happy, nice one :)

PR #905, opened 2023-11-21
feat: Add table component to editor

### LINKED ISSUE #891: Add table component to the editor
### TODO

- [ ] Add new table component to the palette under the `Presentation` section
- [ ] The table component should render only 1 row on the editor panel
- [ ] When we add a new table component a new field should be created in the sample context with mock data to populate the table in the viewer

### Additional information
- [Figma link](https://www.figma.com/file/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?type=design&node-id=290-68794&mode=design&t=0dXyhVaRNNXSJBTH-0)

vsgoulart:
Closes #891


packages/form-js-playground/src/components/PlaygroundRoot.js
```diff
@@ -137,6 +137,27 @@ export function PlaygroundRoot(props) {
...
+          [id]: initialDemoData,
+        };
+
+        dataEditorRef.current.setValue(
+          toString(newData)
+        );
+
+        return newData;
+      });
+    });
```
vsgoulart, line 159: Couldn't really find a good way to test this
I'm open to ideas

packages/form-js-playground/src/components/PlaygroundRoot.js
```diff
@@ -137,6 +137,27 @@ export function PlaygroundRoot(props) {
     paletteRef.current = formEditor.get('palette');
     propertiesPanelRef.current = formEditor.get('propertiesPanel');
 
+    formEditor.on('formField.add', ({ formField }) => {
+      const { initialDemoData, id } = formField;
```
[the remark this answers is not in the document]
vsgoulart: Yeah, your way is better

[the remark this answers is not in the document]
pinussilvestrus: What could also be an alternative, to keep these things out of `form-js-viewer`, to have PlaygroundFormFields that override the config in `form-js-playground` similar to the [EditorFormFields](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/src/render/EditorFormFields.js). This way we could scope these demo data to the Playground only.

Just an idea, I see also good arguments to define these things in the viewer package.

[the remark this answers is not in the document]
vsgoulart: Even though I guess this is an editor feature I think it's ok to "break" the boundaries a bit for the sake of having all the config in a single place. So I think I'll keep as this if you don't see it as a blocker

vsgoulart:
@pinussilvestrus I also updated the schema for the table component a bit, I was going to use the `key` property as the data source, but then this clashes with the `keyed` logic. So I introduced the `dataSource` property to the schema

packages/form-js-playground/src/components/PlaygroundRoot.js
```diff
@@ -137,6 +137,30 @@ export function PlaygroundRoot(props) {
     paletteRef.current = formEditor.get('palette');
     propertiesPanelRef.current = formEditor.get('propertiesPanel');
 
+    formEditor.on('formField.add', ({ formField }) => {
```
pinussilvestrus, line 140: What I'm now wondering after having a second look, does this work with `undo` / `redo`? I think undo, redo might anyway not work super sophisticated in the Form Playground, so it might be ok for now.

packages/form-json-schema/test/spec/validation.spec.js
```diff
@@ -237,6 +237,8 @@ describe('validation', function() {
 
     testForm('rowCount-not-allowed');
 
+    testForm('dataSource-not-allowed');
```
pinussilvestrus, line 240:
```suggestion

    testForm('dataSource-not-allowed');
```

pinussilvestrus (approved):
Looks good from my side 👍 Only one minor thing and a comment.

packages/form-js-playground/src/components/PlaygroundRoot.js
vsgoulart: This basically resets the edit history, I couldn't find a better API to properly append the history

pinussilvestrus: We have the concept of [modeling behaviors](https://github.com/bpmn-io/form-js/tree/develop/packages/form-js-editor/src/features/modeling/behavior) that intercept `commands` as removing, adding form fields, updating IDs etc. That ensures proper handling of the Form Editor's command stack and ensures we properly clean up things on undo/redo.

In this case, we use a simple event listener that runs _after_ the `formField.add` command, so it would not run on undo/redo, e.g. the JSON editor would not properly remove the demo data. I think that's not a problem currently, but something you could think about in the future to implement this as a proper modeling behavior.

pinussilvestrus: cc @Skaiir

vsgoulart: ok, merging this for now then

Skaiir: @vsgoulart 

What's the use case for the demo data? To show something in the editor?

vsgoulart: > @vsgoulart
> 
> What's the use case for the demo data? To show something in the editor?

@Skaiir yes, the goal is to make it easier to get "onboarded" on this new component so he/she doesn't just see an empty table in the preview

PR #906, opened 2023-11-21
chore: Add noEmit

vsgoulart:
Fixes this:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "lerna-build-watch": "lerna run bundle:watch --stream --parallel --include-dependencies --scope",
     "lerna-publish": "lerna publish -m \"chore(project): publish %s\"",
     "lint": "eslint packages",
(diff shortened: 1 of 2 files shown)


pinussilvestrus:
In the `generate-types` npm tasks we do indeed do some transpilation, but only for the `.d.ts` files. We specify `--emitDeclarationOnly` for this specific task, I'd assume this does not conflict with `noEmit` then?

vsgoulart:
> In the `generate-types` npm tasks we do indeed do some transpilation, but only for the `.d.ts` files. We specify `--emitDeclarationOnly` for this specific task, I'd assume this does not conflict with `noEmit` then?

@pinussilvestrus The sub-packages are not extending each other, so there's no issue
I tried it on the `form-js` subpackage and it worked normally
image

pinussilvestrus (approved):

PR #912, opened 2023-11-23
feat: Add table entries to properties panel

### LINKED ISSUE #890: Create properties panels for the table component
### TODO

- [ ] Add section `General`, with the following fields:
   - [ ] `Table header`: optional, type text, FEEL toggable
   - [ ] `Data source`: required, type text, FEEL toggable
   - [ ] `Default number of rows`: required, type number, greater than 0, default value 10
   - [ ] `Pagination`: required, toggle, enabled by default
- [ ] Add section `Columns`, FEEL toggable, list of objects `header` and `context key`

### Additional information
- [Figma link](https://www.figma.com/file/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?type=design&node-id=290-68794&mode=design&t=0dXyhVaRNNXSJBTH-0)

vsgoulart:
Closes #890


packages/form-js-editor/src/features/properties-panel/entries/TextEntry.js
```diff
@@ -25,17 +24,6 @@ export default function TextEntry(props) {
     }
   ];
 
-  // todo: skipped to make the release without too much risk
-  // if (templating.isTemplate(field.text)) {
-  //   entries.push(simpleBoolEntryFactory({
-  //     id: 'strict',
```
vsgoulart, line 31: @Skaiir I deleted these comments, should I create an issue to replace them?

packages/form-js-editor/src/features/properties-panel/entries/ColumnEntry.js
vsgoulart, line 1: The logic of the the static column headers is sort of similar with the static options in the select, but I think they're not that related to be worth it to create a common reusable component

packages/form-js-editor/src/features/properties-panel/entries/TableDataSourceEntry.js
```diff
@@ -0,0 +1,94 @@
...
+
+export function TableDataSourceEntry(props) {
+  const {
+    editField,
+    field
+  } = props;
+
+  const entries = [];
+  entries.push({
+    id: 'source',
```
pinussilvestrus, line 17: Maybe `dataSource`?

packages/form-js-editor/src/features/properties-panel/entries/LabelEntry.js
```diff
@@ -51,13 +51,29 @@ export default function LabelEntry(props) {
...
 
   return entries;
 }
 
+
+/**
+ * @param {string} type
+ * @returns {string}
+ */
+function getEntryLabel(type) {
```
pinussilvestrus, line 66: I added exactly that for the `iFrame`, so you probably need to rebase the branch.

packages/form-js-editor/src/features/properties-panel/entries/HeadersSourceSelectEntry.js
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  /**
+   * @param {string|void} value
+   */
+  const setValue = (value) => {
+    switch (value) {
+    case OPTIONS.static.value:
+      editField(field, {
+        columnsExpression: undefined,
```
pinussilvestrus, line 74: For `values` we have a [modeling behavior](https://github.com/bpmn-io/form-js/blob/develop/packages/form-js-editor/src/features/modeling/behavior/ValuesSourceBehavior.js#L29) that removes the other properties. The advantage of that approach is that it also does the clean ups when using the modeling API directly, and not only in the properties panel entry. It's also pluggable as an extension can simply remove that behavior (for whatever reason). 

Wdyt, should we do the same for `columnsExpression` and `columns` or only once it becomes more complex with more possibilities do define the source?

packages/form-js-editor/src/features/properties-panel/groups/TableHeaderGroups.js
```diff
@@ -0,0 +1,58 @@
...
+      ...StaticColumnsSourceEntry({ field, editField, id })
+    });
+  }
+
+  return groups;
+}
+
+// helpers //////////
+
+const TOOLTIP_TEXT = `"Static" defines a constant, predefined set of form options.
```
pinussilvestrus, line 55:
```suggestion
const TOOLTIP_TEXT = `"List of items" defines a constant, predefined set of form options.
```

packages/form-js-editor/src/features/properties-panel/entries/HeadersSourceSelectEntry.js
vsgoulart: Yeah, I was waiting for you to raise this topic. I'll take a look into this tomorrow

packages/form-js-editor/src/features/properties-panel/entries/TextEntry.js
Skaiir: Nah we've been good without it, if it's needed it'll come up again.

Skaiir:
Had a look over, agreeing with Niklas's points but that's about all I saw at a glance. Will have a serious long review on the complete PR.

packages/form-js-editor/src/features/properties-panel/entries/HeadersSourceSelectEntry.js
vsgoulart: @pinussilvestrus Should be done now

pinussilvestrus: Thanks for the follow up, looks good to me 👍

packages/form-js-editor/test/spec/features/modeling/behavior/ColumnsSourceBehavior.spec.js
```diff
@@ -0,0 +1,157 @@
...
+describe('features/modeling - ColumnsSourceBehavior', function() {
+
+  beforeEach(bootstrapFormEditor(schema, {
+    additionalModules: [
+      modelingModule
+    ]
+  }));
+
+
+  it('should NOT remove values source properties', inject(function(formFieldRegistry, modeling) {
```
pinussilvestrus, line 19:
```suggestion
  it('should NOT remove columns source properties', inject(function(formFieldRegistry, modeling) {
```

pinussilvestrus (approved):

Issue #888, opened 2023-11-08
[Implement] Support table component


vsgoulart:
> This is the implementation epic for https://github.com/camunda/product-hub/issues/1894

### What should we do?
We want to create a table component to render tabular data in a user friendly way

### Issues

- [x] #889
- [x] #890
- [x] #891
- [x] #892
- [x] #936
- [x] #939
- [x] #940

### Test cases:
- [x]  [Table component](https://camunda.testrail.com/index.php?/cases/view/227637)
- [x] [Table component in Modeler and Tasklist](https://camunda.testrail.com/index.php?/cases/view/228503)

esraagamal6:
@vsgoulart Tested Ok on form JS

PR #920, opened 2023-11-29
feat: Add table component to viewer

### LINKED ISSUE #892: Add table component to the viewer

vsgoulart:
I was having some issues with the tests, I'll add them tomorrow morning

Closes #892


Skaiir:
Before I start looking into the code, just a few first impressions / remarks on my side. I checked the [parent issue](https://github.com/bpmn-io/form-js/issues/888), after this PR there's carbonization but I guess you're still planning to change non-carbon styles as well a bit? Anyways there's a few style issues but I won't discuss them for now assuming the above.

**Adding tables pollutes the input data:**

grafik

This section shouldn't be programmatically updated as a side effect of adding a component. I guess this is probably related to the way we add the sample data. I also have some reservations about that but we'll discuss on the full PR.

**Pagination by default is quite a lot right out of the gates:**

grafik

Very visually complex, do we really need to introduce pagination immediately to our users? Especially with a height of 1 which makes it:
- Not a reasonable default, meaning it's a property everyone will have to modify to get a "standard" behavior
- Not the default you usually get from toggling pagination, leading to these kinds of interactions where toggling the checkbox actually changes things

firefox_Gvn7dEDJZx

The that second point, I think we could also just keep the page size in the schema, up for debate.

**Keys and data-sources should probably not be overlappable**

grafik

Binding a key to the component destroys the data and hence leads to the table showing nothing. This is normal sanitation behavior though. So at the very least they should be exclusive.

**Data-sources bypass groups paths**

grafik

This is kind of an interesting one. Because the way we've gone about this is, text path definitions stack with eachother, but feel expressions require using of 'this' keyword to access the local scope (they default to the root). Should we copy the model here? I'm not sure.

**Drag and drop is compressed on small screens**

The minimum width makes things difficult when it comes to having two items side by side:

grafik

pinussilvestrus:
> This section shouldn't be programmatically updated as a side effect of adding a component. I guess this is probably related to the way we add the sample data. I also have some reservations about that but we'll discuss on the full PR.

I think the demo data makes sense with the rationale provided via #905. What makes it a bit clumsy is the fact you will really likely update the `dataSource` after first create, which needs another update in the data editor everytime.

vsgoulart:
>I also have some reservations about that but we'll discuss on the full PR.

@Skaiir I split this task into smaller PRs, everything is merged into `888-table-component`, so this is the last big PR we'll have for epic. After the canonization we can just merge the feature branch into develop since everything is already reviewed

>but I guess you're still planning to change non-carbon styles as well a bit? Anyways there's a few style issues but I won't discuss them for now assuming the above.

Yeah, I'll create a separate last PR for that

> This section shouldn't be programmatically updated as a side effect of adding a component. I guess this is probably related to the way we add the sample data.

This was agreed on the define phase. As @pinussilvestrus mentioned the goal is to provide some data to render so the user isn't completely lost when adding a new table into the schema, this will probably be immediately edited out when modelling it and I don't believe users will add too many tables too. Anyway, it's too late to not go with it now, we have iterate over it when we have user feedback. 

>Very visually complex, do we really need to introduce pagination immediately to our users? Especially with a height of 1 which makes it:
Not a reasonable default, meaning it's a property everyone will have to modify to get a "standard" behavior
Not the default you usually get from toggling pagination, leading to these kinds of interactions where toggling the checkbox actually changes things

The default is 10 rows (it was 1 originally because it was the value I was using for testing). The pagination buttons will also only be visible if there's more than 1 page.
Having pagination is basically the whole purpose of having the table component, because otherwise the form would be giant and the user would have to scroll all the way down to complete the form. (Another option would be limiting the height and having a vertical scroll, but we decided this is a bad UX because we would have a scroll area inside of a scroll area)

The other points are bugs, I'll fix them tomorrow

volodymyr-melnykc:
> Another thing that's a bit inconsistent, the values displayed in the editor aren't matching those of the viewer, which is new:

@Skaiir @vsgoulart I see what's the issue. What we can do is show generic content for a table in the Form Definition, even for the sample data.
Meaning instead of showing columns "Id", "Name" and "Address", have generic "Column 1", "Column 2", "Column 3".
The actual table content is to be shown only in the Form Preview.

image

---

> Usually we have no values at all, but this is also not great:

> I think we need to agree on some visual language practices for elements which are populated through a source. Something like the skeletons in carbon would work well I think.

In the long term, it would be really good to merge these 2 views (Form Definition and Form Preview) into one, from where the user can edit and preview at the same time. We also got several user requests for this.

Skaiir:
> In the long term, it would be really good to merge these 2 views (Form Definition and Form Preview) into one, from where the user can edit and preview a form at the same time. We also got several user requests for this.

Well they aren't the same views so we can't exactly merge them, however something pretty close to that would be to have a single panel with an easily accessible toggle (both via keyboard and UI) between edit mode and view mode. This is pretty common in a lot of modeling tools.

packages/form-js-viewer/src/render/components/form-fields/Table.js
```diff
@@ -1,36 +1,405 @@
...
+  }, [ rowCount, sortBy ]);
+
+  if (disabled) {
+    return <DisabledTable { ...props } />;
+  }
+
+  /** @param {string} key */
+  function toggleSortBy(key) {
+    setSortBy((current) => {
+      if (current === null || current.key !== key) {
```
Skaiir, line 82: I think we should sort ascending first, then descending. My first expectation when clicking names on a table would see something from A-Z not Z-A. 

@volodymyr-melnykc wdyt

packages/form-js-viewer/src/render/components/form-fields/Table.js
```diff
@@ -1,36 +1,405 @@
...
+      return {
+        key,
+        direction: 'asc',
+      };
+    });
+  }
+
+  return (
+    <div class={ formFieldClasses(type) }>
+      {isString(label) && label.length > 0 ? (
```
Skaiir, line 102: We don't usually have this check here, I think we should not have a unique implementation and either move this to the label renderer or take it as a follow-up.

packages/form-js-viewer/src/render/components/form-fields/Table.js
```diff
@@ -1,36 +1,405 @@
...
+                              )}
+                            </>
+                          ) : null}
+                        </span>
+                      </th>
+                    );
+                  })}
+                </tr>
+              </thead>
+              {currentChunk.length === 0 ? (
```
Skaiir, line 158: Nitpick, but we can move the check inside of the tbody. Up to you.

packages/form-js-viewer/src/render/components/form-fields/Table.js
```diff
@@ -1,36 +1,405 @@
...
-        key: 'name'
-      },
-      {
-        label: 'Date',
-        key: 'date'
-      }
-    ]
-  }),
+  create: (options = {}) => {
+    let { id, columnsExpression, columns, rowCount, ...remainingOptions } =
```
Skaiir, line 226: this can be const, for good measures

packages/form-js-viewer/src/render/components/form-fields/Table.js
```diff
@@ -1,36 +1,405 @@
...
+  const [ currentPage, setCurrentPage ] = useState(0);
+  const currentChunk = chunkedData[currentPage] || [];
+
+
+  useEffect(() => {
+    setCurrentPage(0);
+  }, [ rowCount, sortBy ]);
+
+  if (disabled) {
+    return <DisabledTable { ...props } />;
```
Skaiir, line 76: Disabled is an actual state we use (although deprecated). The proper way to create behavior unique to the table in the editor would be to create an EditorTable, this is more hijacking an existing concept for a different purpose which I'm not 100% fond of. See `EditorText` for reference, it's pretty simple to change though and would move some non-viewer code outside of the viewer project.

packages/form-js-viewer/src/render/components/form-fields/Table.js
```diff
@@ -1,36 +1,405 @@
...
+
+  /** @type {[(null|Sorting), import("preact/hooks").StateUpdater<null|Sorting>]} */
+  const [ sortBy, setSortBy ] = useState(null);
+  const evaluatedColumns = useEvaluatedColumns(
+    columnsExpression || '',
+    columns,
+  );
+  const columnKeys = evaluatedColumns.map(({ key }) => key);
+  const evaluatedDataSource = useExpressionEvaluation(
+    dataSource.startsWith('=') ? dataSource : `=${dataSource}`,
```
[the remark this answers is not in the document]
vsgoulart: yep, I'll remove it

packages/form-js-viewer/src/render/components/form-fields/Table.js
vsgoulart: I don't mind either, I just decided it by myself since we didn't go this deep in the definition

volodymyr-melnykc:
@vsgoulart The latest update looks great! 

I have a minor comment regarding the visual layout. 
Can we add vertical spacing between the laber and table? Currently, it's 0, to be 4px (e.g. via "`row-gap`").

image

vsgoulart:
@volodymyr-melnykc done

Skaiir (approved):

Skaiir:
Nice work :)

PR #921, opened 2023-11-30
fix(monkey-patch/performance): make useFilteredFormData a context

Skaiir:
Related to https://github.com/camunda/tasklist/issues/3758.


Skaiir:
Closing this one as it caused issues and we have a better fix in dynamic lists

PR #927, opened 2023-12-04
chore: Bump to diagram-js@13.0.0

vsgoulart:


nikku (commented):
Looks great. Thanks that you follow up with this swiftly :clap:

PR #934, opened 2023-12-08
fix: give tables rendered by text some simple styling

### LINKED ISSUE #869: Improve the default table styles
### Describe the Bug

Currently when rendering a table both in standard styling and with the carbonization layer, it looks pretty bland. With carbon it's a lot worse as we the paddings to zero, which is an actual rendering issue as there is no spacing at all. 

### Steps to Reproduce

1. create a text view
2. add the following text to the view

```
|*test*|test2| 
|---| --- |
|**test222**|**test2**|
|test|test2|
|test|test2|
|test|test2|
```

3. witness the bland styles

grafik

### Expected Behavior

3. Tables should look nice :)

### Extra context

Here's how it looks carbonized, different data but worse rendering, the table elements are touching. 

grafik

Skaiir:
Closes #869

grafik

Nice improvement, I decided to go with the same colors as groups, just to not overextend our palette too much.

### DIFF
--- a/packages/form-js-viewer/assets/form-js-base.css
+++ b/packages/form-js-viewer/assets/form-js-base.css
@@ -71,6 +71,7 @@
     var(--cds-border-strong-01, var(--color-grey-225-10-55))
   );
   --color-borders-group: var(--cds-border-subtle, var(--color-grey-225-10-85));
+  --color-borders-table: var(--color-borders-group);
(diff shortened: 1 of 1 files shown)


volodymyr-melnykc:
It's a nice table!

volodymyr-melnykc (approved):

PR #937, opened 2023-12-08
feat: Add Carbon compatibility styles

### LINKED ISSUE #936: Carbonise new table component

vsgoulart:
Closes #936

### DIFF
--- a/packages/form-js-carbon-styles/src/carbon-styles.scss
+++ b/packages/form-js-carbon-styles/src/carbon-styles.scss
@@ -492,8 +492,7 @@
     letter-spacing: var(--cds-label-01-letter-spacing);
   }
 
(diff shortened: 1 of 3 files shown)


vsgoulart:
Some preview

image
image

Skaiir (approved):
Looks good, as a follow up it would be good to bring a table into the example test cases so that they can easily be seen via `npm run start:carbon` (and also trigger our regression tests)

Maybe we could do this after https://github.com/bpmn-io/form-js/issues/942

Skaiir:
Ah only one thing missing I see actually, can you change the commit message to reference the base issue?

vsgoulart:
> Ah only one thing missing I see actually, can you change the commit message to reference the base issue?
> 
> We usually write `Related to #___` or `Closes #___`

I don't think it makes sense to link to the root issue if we have a split issue. The issue I linked is also linked to the root issue.

PR #955, opened 2023-12-19
fix: allow scripts in iframes by default 

### LINKED ISSUE #954: Enable script execution inside iframes by default (no config)
### Describe the Bug

Current iframes lack the ability to execute scripts which makes them overly restrictive. For now, let's enable the script execution by default and then offer an opt-out later via properties. 

### Steps to Reproduce

1. Add an iframe
2. Link to https://bpmn.io/
3. Try and open the navigation menu inside the iframe => can't

### Expected Behavior

3. This should work by default

### Additional context

Main security params issue: https://github.com/bpmn-io/form-js/issues/901

Skaiir:
Closes #954


christian-konrad (approved):
Works, lgtm 

image

PR #959, opened 2023-12-21
feat: Implement form-js viewer debounce

### LINKED ISSUE #958: Debounce input fields in form-js-viewer

At the moment, we do not debounce text entry in any capacity in form-js. This isn't ideal, especially when working with really big forms, as the latency can be felt on the text entry. 

We should most likely be debouncing textfields, textareas and numberfields. We can follow the configuration method to doing debouncing, as is already implemented for the properties panel / editor. 

n/a

Skaiir:
Closes #958


nikku (approved):
I manually tested it and it works.

PR #963, opened 2023-12-29
fix: properly handle all formats of options sources

### LINKED ISSUE #960: Dynamic options aren't reliable
**What should we do?**

We should ensure that dynamic options for all select-like components work as documented (see https://docs.camunda.io/docs/components/modeler/forms/configuration/forms-config-options/)

That means that values should be able to be defined as objects, for example. 

**Why should we do it?**

Because it's the original behavior. 

See https://app.incident.io/camunda/incidents/798 && https://jira.camunda.com/browse/SUPPORT-19807 for details

Skaiir:
Closes #960

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Radio.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Radio.js
@@ -1,4 +1,5 @@
 import { useRef } from 'preact/hooks';
+import isEqual from 'lodash/isEqual';
 import useOptionsAsync, { LOAD_STATES } from '../../hooks/useOptionsAsync';
 import useCleanupSingleSelectValue from '../../hooks/useCleanupSingleSelectValue';
 import classNames from 'classnames';
@@ -85,16 +86,17 @@ export default function Radio(props) {
       loadState == LOAD_STATES.LOADED && options.map((option, index) => {
 
         const itemDomId = `${domId}-${index}`;
+        const isChecked = isEqual(option.value, value);
 
         return (
           <Label
             id={ itemDomId }
             key={ index }
             label={ option.label }
-            class={ classNames({ 'fjs-checked': option.value === value }) }
+            class={ classNames({ 'fjs-checked': isChecked }) }
             required={ false }>
             <input
-              checked={ option.value === value }
+              checked={ isChecked }
               class="fjs-input"
               disabled={ disabled }
               readOnly={ readonly }
--- a/packages/form-js-viewer/src/render/components/form-fields/Taglist.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Taglist.js
@@ -3,6 +3,7 @@ import { useMemo, useRef, useState } from 'preact/hooks';
 import { useService } from '../../hooks';
 import useOptionsAsync, { LOAD_STATES } from '../../hooks/useOptionsAsync';
 import useCleanupMultiSelectValues from '../../hooks/useCleanupMultiSelectValues';
+import { useGetLabelCorrelation } from '../../hooks/useGetLabelCorrelation';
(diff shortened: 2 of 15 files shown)


packages/form-js-viewer/src/render/hooks/useLabelCorrelation.js
```diff
@@ -0,0 +1,17 @@
...
+import { isObject } from 'min-dash';
+
+/**
+ * This hook allows us to retrieve the label from a value in linear time by caching it in a map
+ * @param {Array} options
+ */
+export function useLabelCorrelation(options) {
+
+  // This allows us to retrieve the label from a value in linear time
+  const labelMap = useMemo(() => Object.assign({}, ...options.map((o) => ({ [_getValueHash(o.value)]:  o.label }))), [ options ]);
```
vsgoulart, line 11: Did you have any performance issues? It's not a good idea to abuse `useMemo`,`useCallback`, etc because of memory leaks

packages/form-js-viewer/src/render/hooks/useLabelCorrelation.js
```diff
@@ -0,0 +1,17 @@
+import { useCallback, useMemo } from 'preact/hooks';
+import { isObject } from 'min-dash';
+
+/**
+ * This hook allows us to retrieve the label from a value in linear time by caching it in a map
+ * @param {Array} options
+ */
+export function useLabelCorrelation(options) {
```
vsgoulart, line 8:
```suggestion
export function useGetLabelCorrelation(options) {
```

packages/form-js-viewer/src/render/components/form-fields/Taglist.js
```diff
@@ -64,8 +66,7 @@ export default function Taglist(props) {
     onChange: props.onChange
   });
 
-  // We cache a map of option values to their index so that we don't need to search the whole options array every time to correlate the label
-  const valueToOptionMap = useMemo(() => Object.assign({}, ...options.map((o, x) => ({ [o.value]:  options[x] }))), [ options ]);
+  const correlateLabel = useLabelCorrelation(options);
```
vsgoulart, line 69:
```suggestion
  const getLabelCorrelation = useLabelCorrelation(options);
```

packages/form-js-viewer/src/render/components/util/sanitizerUtil.js
```diff
@@ -19,6 +20,15 @@ export function sanitizeDateTimePickerValue(options) {
   return value;
 }
 
+export function isDeepIncluded(value, array) {
```
vsgoulart, line 23: The name is not very descriptive, I think we can improve it

```suggestion
export function hasEqualValue(value, array) {
```

packages/form-js-viewer/src/render/components/form-fields/Radio.js
```diff
@@ -1,4 +1,5 @@
 import { useRef } from 'preact/hooks';
+import isDeepEqual from 'lodash/isEqual';
```
vsgoulart, line 2: Let's not rename it unless it's necessary

```suggestion
import isEqual from 'lodash/isEqual';
```

vsgoulart (commented):
It worked fine while testing, I have some minor comments though

packages/form-js-viewer/src/render/hooks/useLabelCorrelation.js
Skaiir: Yeah, this is why it's here and it's always been that way. I've just brought this pattern into a hook because we use it a few places. Otherwise the complexity is n^2 and we want the components to deal with values of n > 1000

Issue #918, opened 2023-11-28
Convert defaults to named exports where possible.


Skaiir:
### What should we do?

Move to using named exports by default.

### Why should we do it?

To ensure consistent naming and reduce reassignments.

pinussilvestrus:
I have the feeling this one will be in backlog forever, as the benefits are not that high. Do you think it makes sense to keep it or rather to our well-known refactor-as-you-go strategy?

Skaiir:
This is now enforced via linting rules.

PR #978, opened 2024-01-10
Use named exports in playground

Skaiir:
Partial PR

Related to https://github.com/bpmn-io/form-js/issues/918

### DIFF
--- a/packages/form-js-playground/src/Playground.js
+++ b/packages/form-js-playground/src/Playground.js
@@ -29,7 +29,7 @@ import { PlaygroundRoot } from './components/PlaygroundRoot';
 /**
  * @param {FormPlaygroundOptions} options
  */
-export default function Playground(options) {
+export function Playground(options) {
 
   const {
     container: parent,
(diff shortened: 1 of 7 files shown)


Skaiir:
@vsgoulart 

I'll do PRs per project for this one, it's a lot of files but the changes are mostly straight-forward one-liners.

packages/form-js-playground/src/Playground.js
```diff
@@ -29,7 +29,7 @@ import { PlaygroundRoot } from './components/PlaygroundRoot';
 /**
  * @param {FormPlaygroundOptions} options
  */
-export default function Playground(options) {
+export function Playground(options) {
```
vsgoulart, line 32: I'm a big fan of doing exports only in the last line of the file like `export {Playground}`, that's what we use on Tasklist

This way it's faster everything that a file exports, it would be cool to have the same standards across both projects, but I'm also ok if you want to leave it as is

vsgoulart (approved):

packages/form-js-playground/src/Playground.js
Skaiir: We can do this in a second step if you like, if we can find a linting rule for it.

Skaiir: Didn't take long to find, https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/exports-last.md.

I can add another PR once we've got all 3 blocks of changes into a single PR

PR #981, opened 2024-01-11
chore: fix internal usePrevious hooks

Skaiir:
Related to https://github.com/bpmn-io/form-js/pull/972#discussion_r1448611855

### DIFF
--- a/packages/form-js-editor/src/render/hooks/usePrevious.js
+++ b/packages/form-js-editor/src/render/hooks/usePrevious.js
@@ -1,13 +1,15 @@
 import {
-  useEffect,
-  useRef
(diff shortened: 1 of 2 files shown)


douglasbouttell-camunda (approved):
👍 

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

vsgoulart (approved):

Skaiir:
Actually there's an issue here, we do depend on making the check in a non-reactive manner some places. I think stuff like deep comparison for example. I'll shelf this one for a bit and get back to it in a bit.

PR #982, opened 2024-01-11
fix: improve text prompts of empty group components

### LINKED ISSUE #965: Dynamic List - handle empty state better with resizing
### Describe the Bug

When resizing the dynamic list to be very little, the empty state starts looking very messy, we should find an alternative to this, either by hiding the state entirely or substituting it for a smaller, simpler text field (or even just an icon). 

grafik

### Steps to Reproduce

1. Add a dynamic list
2. Make it 2 columns wide
3. The placeholder text renders poorly

### Expected Behavior

- Update the placeholder text from "_Drag and drop components here to create a repeatable list item_" to "_**Drag and drop components here.**_":
  - The updated placeholder text to be shown for wider breakpoints.
  - No need to mention in the placeholder text that components will be repeated since there is a separate indicator below.
- When the width of the Dynamic list component is less than 150px then the placeholder text is not displayed.

### Context

- follow-up from https://github.com/bpmn-io/form-js/issues/796

Skaiir:
Closes #965

@volodymyr-melnykc I made the hiding threshold 100pixels instead because it still seems to look okay, let me know how it seems to you. 

@vsgoulart Opted for a mutationobserver for simplicity, since this is purely a visual thing

### DIFF
--- a/packages/form-js-editor/src/render/components/FormEditor.js
+++ b/packages/form-js-editor/src/render/components/FormEditor.js
@@ -65,24 +65,34 @@ function ContextPad(props) {
   );
 }
 
-function Empty(props) {
-  if (props.field.type === 'default') {
-    return <div class="fjs-empty-editor">
+function EmptyGroup() {
+  return (
+    <div class="fjs-empty-component">
+      <span class="fjs-empty-component-text">Drag and drop components here.</span>
+    </div>
+  );
+}
+
+function EmptyForm() {
+  return (
+    <div class="fjs-empty-editor">
       <div class="fjs-empty-editor-card">
         <EmptyFormIcon />
         <h2>Build your form</h2>
         <span>Drag and drop components here to start designing.</span>
         <span>Use the preview window to test your form.</span>
       </div>
-    </div>;
+    </div>
+  );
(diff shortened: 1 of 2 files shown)


packages/form-js-editor/src/render/components/FormEditor.js
```diff
@@ -65,24 +65,57 @@ function ContextPad(props) {
...
+    return () => {
+      if (currentElement) {
+        resizeObserver.unobserve(currentElement);
+      }
+    };
+  }, []);
+
+  return (
+    <div class="fjs-empty-component" ref={ emptyComponentRef }>
+      <span style={ { display: isHidden ? 'none' : 'inline' } }>Drag and drop components here.</span>
```
vsgoulart, line 94: Let's use classes instead of inlining it

vsgoulart (changes requested):
@Skaiir Even though this component won't `ResizeObserver` + refs are no that great for performance.

Let's try to use [container queries instead](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_container_queries) which allows use to avoid shipping more JS

PR #989, opened 2024-01-16
feat: add security properties for iFrame

### LINKED ISSUE #901: iFrame component: Support security attributes

Related to https://github.com/camunda/product-hub/issues/1897
Follow up of #898 

We support the `iFrame` component as sandbox per default without any possibility of providing [security related attributes ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Permissions_Policy#iframe_syntax)(example allow same origin, camera access, microphone access and so on).

This is to do our best to secure the embedded content from causing harm to our users (https://www.reflectiz.com/blog/iframe-security/). 

However, we would like to gather feedback on whether and what of these attributes are still relevant and safe to support. Some web content does not work without such permissions, so we reduce the possible use cases we support. 

Allow iFrame security attributes to be configured. Cf. to the design of the iFrame component.

https://www.figma.com/file/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?type=design&node-id=439%3A7751&mode=design&t=Y6Hr0ZInixtUSMYF-1

image

```json
{
  "type": "iframe",
  "label": "The bpmn-io web page",
  "url": "https://bpmn.io/",
  "height": 400,
  "security": {
    "allowSameOrigin": true,
    "fullscreen": true,
    "geolocation": true,
    "camera": true,
    "microphone": true,
    "allowForms": true,
    "allowModals": true,
    "allowPopups": true,
    "allowTopNavigation": true
  }
}
```

We say that we do not support such use cases.

Skaiir:
Closes #901

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/IFrame.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/IFrame.js
@@ -1,16 +1,12 @@
-import { useContext, useMemo } from 'preact/hooks';
+import { useEffect, useMemo, useState } from 'preact/hooks';
 
-import { FormContext } from '../../context';
-
-import { useSingleLineTemplateEvaluation } from '../../hooks';
+import { useSingleLineTemplateEvaluation, useSecurityAttributesMap } from '../../hooks';
 import { sanitizeIFrameSource } from '../Sanitizer';
 
 import { Label } from '../Label';
 
-import {
-  formFieldClasses,
-  prefixId
-} from '../Util';
+import { formFieldClasses } from '../Util';
+
 
 const type = 'iframe';
 
@@ -21,14 +17,15 @@ export function IFrame(props) {
   const {
     field,
     disabled,
-    readonly
+    readonly,
+    domId
   } = props;
 
   const {
     height = DEFAULT_HEIGHT,
-    id,
     label,
-    url
+    url,
+    security = {}
   } = field;
 
   const evaluatedUrl = useSingleLineTemplateEvaluation(url, { debug: true });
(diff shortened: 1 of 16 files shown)


vsgoulart:
@Skaiir Didn't @volodymyr-melnykc ask for the `show security attributes` toggle to be removed?

packages/form-js-viewer/src/render/components/form-fields/IFrame.js
```diff
@@ -37,10 +34,16 @@ export function IFrame(props) {
...
   const evaluatedLabel = useSingleLineTemplateEvaluation(label, { debug: true });
 
-  const { formId } = useContext(FormContext);
+  const [ sandbox, allow ] = useSecurityAttributesMap(security);
+  const [ iframeRefresh, setIframeRefresh ] = useState(0);
+
+  // forces re-render of iframe when sandbox or allow attributes change, as browsers do not do it automatically
+  useEffect(() => {
+    setIframeRefresh(count => count + 1);
+  }, [ sandbox, allow ]);
```
vsgoulart, line 43: Kind of hacky, but I can't think of an option myself right now

packages/form-js-viewer/src/render/components/form-fields/IFrame.js
```diff
@@ -51,8 +54,12 @@ export function IFrame(props) {
...
           height={ height }
           class="fjs-iframe"
-          id={ prefixId(id, formId) }
-          sandbox="allow-scripts"
+          id={ domId }
+          sandbox={ sandbox }
+          key={ 'iframe-' + iframeRefresh }
+
+          /* @Note: JSX HTML attributes do not include <allow> */
+          { ...{ allow: allow } }
```
vsgoulart, line 62:
```suggestion
          { ...{ allow } }
```

vsgoulart (commented):
@Skaiir everything looks fine to me, we just need to check the previous comment I made about the toggle

packages/form-js-viewer/src/render/components/form-fields/IFrame.js
Skaiir: I mean I can also ref the iframe, then call the refresh manually via script, it's a little more code, it's equally hacky, so let's go for the virtual DOM hack instead of the regular DOM one.

Skaiir:
> @Skaiir Didn't @volodymyr-melnykc ask for the show security attributes toggle to be removed?

Yeah the PR was made before those comments, sorting this out now.

Skaiir:
Made some adjustments. I'm still a bit apprehensive with regards to this configuration here (ignore validation error).

grafik

This is an XSS attack surface, because if you've got allow-scripts and allow-same-origin that's the equivalent of the iframe basically not being a sandbox anymore, and yeah if there's malicious scripts it can elevate its own priviledges to doing everything, plus it has access to the session cookies, which means it can escalate into session hijacking.

volodymyr-melnykc (approved):

vsgoulart (approved):
LGTM

Issue #894, opened 2023-11-09
`null` initial values break number inputs


vsgoulart:
### Describe the Bug

When initializing input of type numbers with the value `null`, these inputs show a `NaN` error. Value `null` should be accepted since they are valid output of a completed form.

### Steps to Reproduce

1. Render a form with a input type number with initial value `null`
2. The form is rendered with an error in the number input (like in the image below)

### Expected Behavior

The input should show an empty field (since `null` is a valid output on form submissions)

### Additional information

![image](https://github.com/bpmn-io/form-js/assets/1022916/4495b3cf-ad99-4e42-a7f8-348a82e9d07d)

Skaiir:
@vsgoulart What version was this on? Can't replicate on master atm. Maybe a tasklist only issue?

vsgoulart:
> @vsgoulart What version was this on? Can't replicate on master atm. Maybe a tasklist only issue? Maybe something fixed by accident.

There's a problem with the issue the description, but the root cause of this is indeed on Tasklist.

But investigating this highlights another potential issue on `form-js` which is that the initial validation is inconsistent across components like in the image below (some things are validated on the first render and some aren't):
![image](https://github.com/bpmn-io/form-js/assets/1022916/653a5348-9c95-468c-8d18-2337cb18efb5)

Skaiir:
Well so the way this works, there's first and sanitation that runs on every field on start, and then there's validation. Number has also a special validation that checks for NaN. Yeah it's inconsistent, I think we should always just sanitize and report that sanitation happened. But as of now, we don't have any mechanism for reporting that, so we would just sanitize and forget about "NaN" entirely. Also text null should not parse to "null".

@christian-konrad does that make sense to you? We have a ticket tracking sanitation reporting somewhere that we can get to some other time.

Skaiir:
Just wanted to be thorough here, I checked all the conversion behaviors

![grafik](https://github.com/bpmn-io/form-js/assets/17801113/ded3e86f-1539-48f7-8495-4a5caefa234c)

I'm thinking we get rid of NaN, it's a very unique way of doing things compared to all the other components / types. Just sanitation should be enough. It may still make sense to display NaN in the number field when pasting in text into it, but then again there are better ways of doing this, like short toast notifications or something, and we don't know how frequent of a problem that is. I don't know, this can be discussed.

Null for text entry components should sanitize to "", not "null".

[...]

PR #998, opened 2024-01-20
feat: implement number field debouncing

### LINKED ISSUE #958: Debounce input fields in form-js-viewer

Skaiir:
Closes #958
(again)

After a lot of reworks and problems I finally got this sorted.
For testing, make sure the debouncing is working, cleaning up when tabbing out. This should debounce when holding up or down arrow, when typing, and when clicking the arrows via the mouse. 

I've also as part of this removed some problem things, like the fact that we were still sending an onChange with 'NaN' (a concept we had removed, but not fully, via https://github.com/bpmn-io/form-js/issues/894), the test case is respectively changed. 

What we are keeping is the NaN state when pasting something that's not a valid number, this is only visual and is here to give users feedback, otherwise they might think pasting just doesn't do anything. In the future a toast element might be more appropriate. Anyways this state is also more robust now.

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Number.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Number.js
@@ -1,6 +1,7 @@
 import Big from 'big.js';
 import classNames from 'classnames';
-import { useCallback, useMemo, useRef, useState } from 'preact/hooks';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'preact/hooks';
+import { useFlushDebounce, usePrevious } from '../../hooks';
 
 import { Description } from '../Description';
 import { Errors } from '../Errors';
@@ -32,8 +33,7 @@ export function Numberfield(props) {
     onFocus,
     field,
     value,
-    readonly,
-    onChange
+    readonly
   } = props;
 
   const {
@@ -42,7 +42,6 @@ export function Numberfield(props) {
     appearance = {},
     validate = {},
     decimalDigits,
-    serializeToString = false,
     increment: incrementValue
   } = field;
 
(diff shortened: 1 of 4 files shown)


packages/form-js-viewer/src/render/components/form-fields/Number.js
```diff
@@ -55,94 +54,120 @@ export function Numberfield(props) {
...
+  const previousCachedValue = usePrevious(value);
 
-  }, [ stringValueCache, value, cacheValueMatchesState ]);
-
-  const arrowIncrementValue = useMemo(() => {
-
-    if (incrementValue) return Big(incrementValue);
-    if (decimalDigits) return Big(`1e-${decimalDigits}`);
-    return Big('1');
+  useEffect(() => {
```
douglasbouttell-camunda, line 66: I think you can take out the `useEffect` here and just have the if statement running in the render.

packages/form-js-viewer/src/render/components/form-fields/Number.js
```diff
@@ -55,94 +54,120 @@ export function Numberfield(props) {
...
+    setDisplayValue(stringValue);
+    setCachedValue(sanitize(stringValue));
+
+  }, [ sanitize ]);
+
+  // when external changes occur independently of the input, we update the display and cache values of the component
+
+  const previousValue = usePrevious(value);
+
+  useEffect(() => {
```
douglasbouttell-camunda, line 114: Same as above. Don't know if you need this wrapped in `useEffect`

Skaiir: Yeah probably not, but then again it does encapsulate the behavior in a nice block. I mean either way works for me, I've tested without the effects and it's fine as well. This is more of a style thing, since executing directly in the render is basically an effect that runs every time. In terms of efficiency I'd say no effect is better, since we're just doing a single check as opposed to the multi checks involved in effect eval. But that's not something we should factor in.

Skaiir: I'll follow the react guidelines on this one, and we'll go without the effects. Not particularly opinionated on this.

douglasbouttell-camunda (approved):
Looks fine. Just need to see why the screenshots are different.

Skaiir:
@douglasbouttell-camunda These are the cross-library screenshots and I'm not sure why but they always fail. Within form-js anyways `Visual regression tests` is the only thing we really rely on.

vsgoulart:
> @douglasbouttell-camunda These are the cross-library screenshots and I'm not sure why but they always fail. Within form-js anyways `Visual regression tests` is the only thing we really rely on.

@douglasbouttell-camunda @Skaiir We have to update the screenshots on the Tasklist repo, but I'm waiting for the next release because we have some breaking changes on `form-js` ui

PR #1000, opened 2024-01-22
fix: repeatable entry parameters now validate with errors

### LINKED ISSUE #966: Dynamic List - Show proper validation error when setting a default number of items above 20
### Describe the Bug

The default number of repetitions option is limited to 20 as a maximum, but the error message associated is only shown on hover and a bit obscure.

grafik

### Steps to Reproduce

1. Add a dynamic list
2. Set number of default repetitions to 2
3. See it reflected in the preview
4. Set it to 22 now
5. See it not reflected in the preview
6. It's not obvious that the input was refused, no error message is clear unless hovering for a long time

### Expected Behavior

6. The fact that we don't accept above 20 should be clearly communicated

### Context

- follow-up from https://github.com/bpmn-io/form-js/issues/796

Skaiir:
Closes #966

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/factories/simpleRangeIntegerEntryFactory.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/factories/simpleRangeIntegerEntryFactory.js
@@ -1,6 +1,9 @@
 import { get } from 'min-dash';
 import { useService } from '../../hooks';
-import { NumberFieldEntry, isNumberFieldEntryEdited } from '@bpmn-io/properties-panel';
+import { TextFieldEntry, isTextFieldEntryEdited } from '@bpmn-io/properties-panel';
+import { isValidNumber } from '../../Util';
+
+import Big from 'big.js';
 
 export function simpleRangeIntegerEntryFactory(options) {
   const {
@@ -9,8 +12,7 @@ export function simpleRangeIntegerEntryFactory(options) {
     path,
     props,
     min,
-    max,
-    defaultValue
+    max
   } = options;
 
   const {
@@ -26,9 +28,8 @@ export function simpleRangeIntegerEntryFactory(options) {
     editField,
     min,
     max,
-    defaultValue,
     component: SimpleRangeIntegerEntry,
-    isEdited: isNumberFieldEntryEdited
+    isEdited: isTextFieldEntryEdited
   };
 }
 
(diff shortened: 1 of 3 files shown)


volodymyr-melnykc (approved):

packages/form-js-editor/src/features/properties-panel/entries/factories/simpleRangeIntegerEntryFactory.js
```diff
@@ -39,31 +40,39 @@ const SimpleRangeIntegerEntry = (props) => {
...
     element: field,
-    step: 1,
-    min,
-    max,
     getValue,
     id,
-    setValue
+    setValue,
+    validate: (value) => {
+      if (value === undefined || value === null || value === '') return;
```
vsgoulart, line 71: If the minimum is 1 we shouldn't allow nil values

packages/form-js-editor/src/features/properties-panel/entries/factories/simpleRangeIntegerEntryFactory.js
```diff
@@ -39,31 +40,39 @@ const SimpleRangeIntegerEntry = (props) => {
...
+      if (!isValidNumber(value)) return 'Should be a valid number.';
```
vsgoulart, line 72: We already check for `Number.isInteger`, this check is not necessary

packages/form-js-editor/src/features/properties-panel/entries/factories/simpleRangeIntegerEntryFactory.js
```diff
@@ -39,31 +40,39 @@ const SimpleRangeIntegerEntry = (props) => {
...
+      if (!Number.isInteger(Number(value))) return 'Should be an integer.';
+      if (Big(value).cmp(min) < 0) return `Should be at least ${min}.`;
+      if (Big(value).cmp(max) > 0) return `Should be at most ${max}.`;
```
vsgoulart, line 75: I think it's better to add braces in each if case, so it's more resilient to mistakes if we ever want to change the logic

packages/form-js-editor/src/features/properties-panel/entries/factories/simpleRangeIntegerEntryFactory.js
Skaiir: No we do, because there's always the "not configured" state with our properties, which goes back to a certain default.

vsgoulart (approved):

PR #1005, opened 2024-01-25
fix: Prevent demo data from being created on edited tables

### LINKED ISSUE #997: Table component demo data is being added when importing an edited table
### Describe the Bug

When we added the table component we added feature of creating some initial table component, but with that we added the bug that creates this initial demo data on a table component that was edited via properties panel, we should avoid creating demo data on edited tables because we can't guarantee the table has compatible columns.

### Steps to Reproduce

1. Create a table, edit the `dataSource` via properties panel
2. Save this form schema
3. Reload the editor and import the saved schema
4. The initial demo data is created based on the `id` and not on the `dataSource`

### Expected Behavior

Create the initial demo data only if the `dataSource` has the same value as the `id` and only if the columns are compatible with the demo data

### Environment

- OS: [e.g. Windows 7]

vsgoulart:
With the table component we added the concept of adding demo data to the new component, this data was being added when importing forms into the modeler.
This PR prevents this from happening and only adds the demo data when the schema matches with the demo data

Closes #997

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Table.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Table.js
@@ -23,14 +23,16 @@ const type = 'table';
  * @property {string} label
  * @property {string} key
  *
+ * @typedef Field
+ * @property {string} id
+ * @property {Array<Column>} [columns]
+ * @property {string} [columnsExpression]
+ * @property {string} [label]
+ * @property {number} [rowCount]
+ * @property {string} [dataSource]
+ *
  * @typedef Props
- * @property {Object} field
- * @property {string} field.id
- * @property {Array<Column>} [field.columns]
- * @property {string} [field.columnsExpression]
- * @property {string} [field.label]
- * @property {number} [field.rowCount]
- * @property {string} [field.dataSource]
+ * @property {Field} field
  *
  * @param {Props} props
  * @returns {import("preact").JSX.Element}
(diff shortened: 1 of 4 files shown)


douglasbouttell-camunda (approved):

packages/form-js-viewer/src/render/components/form-fields/Table.js
```diff
@@ -263,11 +265,30 @@ Table.config = {
...
-    { id: 1, name: 'John Doe', date: '31.01.2023' },
-    { id: 2, name: 'Erika Muller', date: '20.02.2023' },
-    { id: 3, name: 'Dominic Leaf', date: '11.03.2023' }
-  ],
+
+  /**
+   *
+   * @param {Field} field
+   */
+  generateInitialDemoData: (field) => {
```
Skaiir, line 273: Generally, let's be careful with these kinds of changes. Our config object is tied in with the custom component framework and should be treated more or less as public API. Of course we're never obligated but changing how it's structured is something we should not do often. 

I don't think in this case it matters that much, because the whole initialDemoData stuff is a very unique definition here, but I'm concerned people will want to hook into this and we end up getting rid of it in the future.

Maybe it's worth annotating this as experimental, so that people seeing this implementation don't try to copy and rely on it.

Skaiir: Or moving it to an experimental sub-object, which should get the point across 🤔

vsgoulart: @Skaiir I added the `@experimental` tag

PR #1017, opened 2024-01-31
fix: resize textarea on input changes

### LINKED ISSUE #1011: Auto-resize of textarea should happen on visual text changes, not on debounce triggers
### Describe the Bug

Auto-resizing of textareas is debounced, which leads to delayed looking autosizing.

### Steps to Reproduce

1. create a new textarea
2. in the viewer, in the textarea, spam some text, enter, some more text, enter
3. the resizing is delayed

### Expected Behavior

3. the resizing should happen whenever a new line is created

Skaiir:
Closes #1011

We need to also recalculate the text-area size on input changes so that it doesn't wait for deboucing to kick in. 

This implementation ends up making the call twice (once on typing, once when the value change actually happens through debouncing), but we need to keep the value change one active so that the textarea still resizes when externals change its state.

Well it's only running twice if you type on character every 301ms. It's better to say it runs one extra time

Differentiating a self-triggered value change from an external one does not seem worth the effort. So I think this is the right fix.

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Textarea.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Textarea.js
@@ -33,7 +33,7 @@ export function Textarea(props) {
   const { required } = validate;
   const textareaRef = useRef();
 
-  const [ onInputChange, flushOnChange ] = useFlushDebounce(({ target }) => {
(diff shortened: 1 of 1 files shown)


Skaiir:
Visual regression failure is due to fonts not loading fully  before the screenshots.

vsgoulart (approved):

PR #1018, opened 2024-01-31
merge dev to master (prep for 1.7)

Skaiir:


Skaiir:
@vsgoulart Do you seen anything in there that shouldn't get released? If not, I would suggest that we do a quick alpha release of this for integration tests. I'll keep fixing stuff I find on my end.

vsgoulart (approved):

Issue #1029, opened 2024-02-08
Default rows of dynamic list values are cleared 


esraagamal6:
### Describe the Bug
When a form has a dynamic list with default rows, values added are cleared when user clicks on other fields

### Steps to Reproduce

1. In tasklist open a task that has a dynamic list with some fields and default number of rows (form attached)
2. Add a value on the default row and add a new row

https://github.com/bpmn-io/form-js/assets/129334285/6597ad18-c163-4070-a66c-3c60200e2662

[formdynamiclist.json](https://github.com/bpmn-io/form-js/files/14210298/formdynamiclist.json)

### Actual Behavior
Data is cleared

### Expected Behavior
Data is not cleared

### Environment
Tasklist 8.5.0-alpha1

Skaiir:
This should be fixed in the next tasklist release.

PR #1033, opened 2024-02-13
fix: ensure repeatRenderManager passes down props correctly

Skaiir:
Related to #1029


Skaiir:
@vsgoulart could you check if this indeed fixes #1029 ?

vsgoulart (approved):
Fixed the issue for me on Tasklist

PR #1035, opened 2024-02-13
fix: improve drag preview alignment

### LINKED ISSUE #596: Drag preview icon should display under cursor (or at least close to it)
### Describe the Bug

Currently, the drag preview does not appear under the cursor, but displays offset relative to where it was grabbed from, which makes it a little clunky to drag around (it becomes uncertain whether to use the mouse or the preview as the drop target). 

*Where the preview appears when grabbed at the rightmost point (red star)*
image

### Steps to Reproduce

1. add any field
2. start dragging it from its rightmost point
3. preview isn't under cursor, and remains offset until drop

### Expected Behavior

3. preview should remain under the mouse cursor

Skaiir:
Closes #596


Skaiir:
This has already been merged a long time, dead branch kept as a PR

PR #1046, opened 2024-02-15
fix: ensure simpleSelect mouseDown event deactivates normally

### LINKED ISSUE #1045: Simple select throws some errors due to unmounted ref
### Describe the Bug

Following [this change](https://github.com/bpmn-io/form-js/commit/53ca342bf18fc4bb8a45a378e0770d31b09fbde8), sentry is picking up a few errors: `Cannot read properties of undefined (reading 'focus' / 'blur')`

This looks like it 's trying to call functions on the component when it isn't mounted.

After a bit more investigation, this is a full blown bug as the mousedown event is trying to access the dropdown even when the component is disabled, and hence not rendered. 

### Steps to Reproduce

1. disable a simpleSelect
2. click around on it
3. an exception it thrown

### Expected Behavior

3. no exception  should be thrown

Skaiir:
Closes #1045

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/parts/SimpleSelect.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/parts/SimpleSelect.js
@@ -56,7 +56,13 @@ export function SimpleSelect(props) {
   }, [ disabled, isDropdownExpanded, loadState, readonly, value ]);
 
   const onMouseDown = useCallback((e) => {
+
     const input = inputRef.current;
+
+    if (disabled || !input) {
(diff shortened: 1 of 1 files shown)


Skaiir:
Turns out there was an actual bug bug on this one. This event shouldn't fire when the component is disabled. I've also just null checked the ref as well for good measure:

https://camunda.slack.com/archives/C043W5V88M7/p1705413165147039?thread_ts=1705411723.101669&cid=C043W5V88M7

vsgoulart (approved):

PR #1052, opened 2024-02-16
fix: Prevent blur event from having value of out sync

### LINKED ISSUE #1036: Validation is triggered incorrectly
### Describe the Bug

In forms with a required field, tabbing out of the field does not trigger validation properly

### Steps to Reproduce

1. Model a form with a required field
2. Add a value to the field a tab out (it needs to be very fast)
3. The validation error appears (this seems to be happen only when editing it from the initial state)

### Expected Behavior

The validation error should not appear if the field is valid

### Additional info

- [Support case](https://jira.camunda.com/browse/SUPPORT-20434)

vsgoulart:
Closes #1036

### DIFF
--- a/packages/form-js-viewer/src/render/components/FormField.js
+++ b/packages/form-js-viewer/src/render/components/FormField.js
@@ -98,6 +98,8 @@ export function FormField(props) {
 
   const onBlur = useCallback(() => {
 
+    const value = get(data, valuePath);
+
     if (initialValidationTrigger) {
       setInitialValidationTrigger(false);
       viewerCommands.updateFieldValidation(field, value, indexes);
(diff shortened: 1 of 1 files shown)


packages/form-js-viewer/src/render/components/FormField.js
```diff
@@ -98,14 +98,16 @@ export function FormField(props) {
...
+
     if (initialValidationTrigger) {
       setInitialValidationTrigger(false);
       viewerCommands.updateFieldValidation(field, value, indexes);
     }
 
     eventBus.fire('formField.blur', { formField: field });
 
-  }, [ eventBus, field, indexes, value, viewerCommands, initialValidationTrigger ]);
+  }, [ eventBus, field, indexes, viewerCommands, initialValidationTrigger, data, valuePath ]);
```
vsgoulart, line 110: In the end the proper fix was just 2 lines long, I had a more complicated fix that was longer 🫠

Skaiir: Yeah I considered doing this at some point but thought the value would repopulate through. Oh well, fair enough. This should ensure things don't get out of sync again.

Skaiir (approved):

douglasbouttell-camunda (approved):

PR #1065, opened 2024-02-21
fix: hide readonly option when disabled is active

### LINKED ISSUE #1063: Disabled should hide readonly property
### Describe the Bug

grafik

Disabled property takes precedence over readonly, so we should hide the options from the properties panel when we activate it, since it doesn't make sense to configure both.

### Steps to Reproduce

1. add a number field
2. set disabled to true
3. readonly option is still changable

### Expected Behavior

3. readonly options should be hidden

Skaiir:
Closes #1063


vsgoulart (approved):
We could add a test for that if it's not too much work

Skaiir:
CI being CI, but I've tested locally and it's all green.

PR #1068, opened 2024-02-22
fix: dynamic options properly filter input values

### LINKED ISSUE #1067: Radio group with dynamic values does not check initial value
### Describe the Bug

When populating a radio group via `valuesExpression` the initial value is not checked

### Steps to Reproduce

1. Render this [form](https://github.com/bpmn-io/form-js/files/14363703/formdynamiclist.json)
2. Try to initialize it with the payload below:
```json
{
  "radio_field_options": [
    { "label": "Radio label", "value": "radio_value_1" },
    { "label": "Radio label 2", "value": "radio_value_2" }
  ],
  "radio_field": "radio_value_1"
}
```
4. `Radio label` is not checked

### Expected Behavior

`Radio label` should be checked

### Environment

- Library version: possibly since 1.5.0

Skaiir:
Closes #1067


Skaiir:
Main changes here:
- Stopped doing early checks for feel-expression driven stuff, because it's kinda complicated to evaluate feel ahead of time like this, and we already have a mechanism to ensure no value / variable mismatch: useCleanupSingle/MultiSelectValue. We probably can bring it back by changing how sanitation works, but for now I'm happy with this (will raise an investigation ticket after).
- Changed how deep compare memoization works, I was way overcomplicating it, got rid of state and effect
- Changed how getOptions works, also overcomplicated, got rid of state and effect as well, useMemo now and it doesn't delay y one render cycle

packages/form-js-viewer/src/render/hooks/useDeepCompareMemoize.js
```diff
@@ -0,0 +1,18 @@
+import { useRef } from 'preact/hooks';
+import isEqual from 'lodash/isEqual';
+
+/**
+ * A custom hook to manage state changes with deep comparison.
+ *
+ * @param {any} value - The current value to manage.
+ * @returns {any} - Returns the current state.
```
vsgoulart, line 8:
```suggestion
 * @template T
 * @param {T} value - The current value to manage.
 * @returns {T} - Returns the current state.
```

packages/form-js-viewer/src/render/hooks/useOptionsAsync.js
```diff
@@ -32,44 +32,38 @@ export function useOptionsAsync(field) {
...
-
+      const keyedOptions = (initialData || {})[optionsKey];
       if (keyedOptions && Array.isArray(keyedOptions)) {
         options = keyedOptions;
       }
+    }
 
     // static options
-    } else if (staticOptions !== undefined) {
+    else if (staticOptions !== undefined) {
```
vsgoulart, line 51: better move the brace back next to the else :sweat_smile:

vsgoulart:
There's something wrong with puppeteer, it's also failing on my PR

packages/form-js-viewer/src/render/hooks/useOptionsAsync.js
Skaiir: I'll revert this one because it's historical but maybe something worth adding to style rules. It's inconsistent in the code base.

Personally, I much prefer the braces around each block, I find it reads more natural.

Skaiir: We could ask the bpmn.io folks about this one and get a vote going.

Skaiir:
> There's something wrong with puppeteer, it's also failing on my PR

Yup, but I've also gotten some green runs in between the broken ones. We'll see if it runs again. Locally, it's running fine.

vsgoulart (approved):

PR #1069, opened 2024-02-22
chore: add prettier

vsgoulart:


douglasbouttell-camunda:
Will this only run manually? Do we want to automatically run this as a pre-commit hook?

vsgoulart:
> Will this only run manually? Do we want to automatically run this as a pre-commit hook?

@douglasbouttell-camunda we just don't have the setup for that currently and I didn't want to add more deps
I think most people have it configured in their editor already and since we already check it on the CI I think we can try to go without it for now

douglasbouttell-camunda:
@vsgoulart Looks like windows is failing because it is trying to enforce a different set of line endings. Could we [set the git repo to define which line endings to use?](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings)

douglasbouttell-camunda (commented):
Looks this will fail cross platform since line endings are not configured in this repo and defaulting to the OS default. This differs between mac and windows.

vsgoulart:
> Looks this will fail cross platform since line endings are not configured in this repo and defaulting to the OS default. This differs between mac and windows.

@douglasbouttell-camunda I'll take a look later, [I think Prettier should handle it](https://prettier.io/docs/en/options.html#end-of-line)

vsgoulart:
@douglasbouttell-camunda End of line issues are solved

douglasbouttell-camunda (approved):

nikku:
I'm not a fan (personally) but if you feel that will speed up your development and assert decent code quality (that you cannot maintain otherwise), then let's go :+1:. 

We'll not adopt prettier at large across the bpmn.io code base in the forseeable future.

PR #1074, opened 2024-02-27
Implement hidden fields

### LINKED ISSUE #1070: Change the iframe preview site
### Describe the Bug

The bpmn.io website doesn't particularly like being rendered in iframes and throws a few errors, mostly around cross origin request, which can be disturbing during debugging. It makes sense thematically but maybe we would prefer not having errors in our most used debugging path. 

### Steps to Reproduce

1. `npm run start:playground`
2. look at the browser console, 7-11 errors are thrown all from the iframe

### Expected Behavior

2. our browser console should remain clean

### LINKED ISSUE #1073: Expression fields for data processing and output / Hidden fields

Sometimes, we want to produce some additional data alongside that produced by our regular fields. We also may want to produce some generic data which can be used as intermediate information to simplify further feel expressions down the line.

While the ideal solution for this would be to have some powerful data handling capabilities, a quick and easy way to fix this at least for the time being is to solve this via a hidden field approach.

The aforementioned data panel, but this would be very costly to implement. 

### Additional context

Initial discussions: https://github.com/camunda/camunda-modeler/issues/4123

Skaiir:
Closes #1073 
Closes #1070
Related to #1076


Skaiir:
Small issue I noticed is that in the editor when you change something in the properties panel, the state resets and doesn't re-initialize in the field. Probably have to hook into the reset event to fix this but not crucial right now.

Skaiir:
@vsgoulart The true PR size more like 500, I have one commit in there that's 1000 lines long where I made a change where we now have a single function call to verify the structure of the properties panel for a given field. There were a lot of these properties panel calls. Doesn't save a ton of lines but it should be easier to work with:

```
expectPanelStructure(container, {
      'General': [
        'Field label',
        'Field description',
        'Key',
        'Default value',
        'Disabled',
        'Read only'
      ],
      'Condition': [],
      'Validation': [
        'Required'
      ],
      'Custom properties': []
});
```

Skaiir:
Back to draft while I figure out what's playgroundroot's problem.

volodymyr-melnykc:
@Skaiir The expression component is placed second in the Input components section. Since this is a component that is less common than other basic input fields, can we move it to be the last one?

image

Skaiir:
@volodymyr-melnykc 

> can we move it to be the last one?

Done :)

vsgoulart (changes requested):
@Skaiir Sorry, but we'll have to split this PR...

Let's do the playground root refactor in one PR, [the refactor you mentioned here in another](https://github.com/bpmn-io/form-js/pull/1074#issuecomment-1970602546) and then the hidden field in another PR

Skaiir:
Merged via split PR

PR #1081, opened 2024-03-04
chore: explicitely define type button on non-submit-button elements

Skaiir:
Related to #1077


Skaiir:
Kind of a no brainer on this one, basically just adding type=button to all buttons to avoid them submitting forms.

PR #1086, opened 2024-03-06
fix: properly pass `this` variable context to dynamic list elems

### LINKED ISSUE #1085: Incorrect assignment of the `this` variable in dynamic lists
### Describe the Bug

The this variable in dynamic lists is assigned to the root of a dynamic list, but it should be set to the actually individual data item. 

### Steps to Reproduce

1. create a dynamic list
2. give it a textfield bound to key `variable1`
3. define a text view with feel expression `=this.variable1`
4. set textfield to `"A"`
5. text view shows `"[A]"`

### Expected Behavior

5. text view should show `A`

Skaiir:
Closes #1085

Issue was that restProps also had "value" passed from outer elements, but it was the value of the full dynamic list. So it was overriding the value obtained in the map. Simply changed the variable name to get around the override. Added a test scenario to ensure this never happens again.

### DIFF
--- a/packages/form-js-viewer/src/features/repeatRender/RepeatRenderManager.js
+++ b/packages/form-js-viewer/src/features/repeatRender/RepeatRenderManager.js
@@ -79,11 +79,11 @@ export class RepeatRenderManager {
 
     return (
       <>
-        {displayValues.map((value, index) =>
+        {displayValues.map((itemValue, itemIndex) =>
           <RepetitionScaffold
-            key={ index }
-            index={ index }
-            value={ value }
+            key={ itemIndex }
+            itemIndex={ itemIndex }
+            itemValue={ itemValue }
             parentExpressionContextInfo={ parentExpressionContextInfo }
             repeaterField={ repeaterField }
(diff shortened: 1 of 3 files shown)


packages/form-js-viewer/src/features/repeatRender/RepeatRenderManager.js
```diff
@@ -79,11 +79,11 @@ export class RepeatRenderManager {
 
     return (
       <>
-        {displayValues.map((value, index) =>
+        {displayValues.map((itemValue, itemIndex) =>
           <RepetitionScaffold
```
douglasbouttell-camunda, line 83: (nitpick, something to keep in mind for the future) 

I would have `{ ...restProps }` at the start do that it can't override any explicitly named props in future and help avoid this issue.

React will evaluate properties in order so you can define elements like so
```jsx
<SomeComponent
  overridableProperty={...}
  {...restProps}
  nonOverridableProperty={...}
/>

douglasbouttell-camunda (approved):
Took me a second to understand why this was broken. Good catch! Left you some advice for the future.

PR #1090, opened 2024-03-07
chore: bump feelers and feelin, allowing overrides of special keywords from context

### LINKED ISSUE #975: Variables named duration cannot be used in feel expressions
### Describe the Bug

When defining a variable named duration and using it in a feel expression or template, an error occurs.

grafik

This is most likely due to the fact that duration is a reserved keyword in feel

### Steps to Reproduce

1. Add a text field
2. Be in templating mode on the text property
3. Enter a template like `Duration = {{duration}}`
4. Set an input variable duration to some random value
5. The rendering is broken

### Expected Behavior

5. (unsure) the variable should overwrite the default feel function and operate properly

Skaiir:
Closes #975

Handled all the changes on a feelers update. Just requires a bump here, works as expected.

There are test cases covering this scenario in both the feelin and feelers repos. I don't think checking the feel evaluations is a responsibility of this library, especially considering expression languages are overridable in this project.


douglasbouttell-camunda (approved):
LGTM. Carbonization snapshot tests are failing though.

vsgoulart:
@douglasbouttell-camunda it needed to be fixed on the Tasklist side, I think I fixed
Let's see the result of the CI

PR #1098, opened 2024-03-18
Sub-PR: expression fields fix playground root

Skaiir:
Sub-pr of https://github.com/bpmn-io/form-js/pull/1074

### DIFF
--- a/packages/form-js-playground/src/components/PlaygroundRoot.js
+++ b/packages/form-js-playground/src/components/PlaygroundRoot.js
@@ -17,104 +17,77 @@ import { EmbedModal } from './EmbedModal';
 import { JSONEditor } from './JSONEditor';
 import { Section } from './Section';
 
-
 import './FileDrop.css';
 import './PlaygroundRoot.css';
 
-
-export function PlaygroundRoot(props) {
+export function PlaygroundRoot(config) {
 
   const {
-    additionalModules = [], // goes into both editor + viewer
-    actions: actionsConfig = {},
+    additionalModules, // goes into both editor + viewer
+    actions: actionsConfig,
     emit,
-    exporter: exporterConfig = {},
-    viewerProperties = {},
-    editorProperties = {},
-    viewerAdditionalModules = [],
-    editorAdditionalModules = [],
-    propertiesPanel: propertiesPanelConfig = {},
-    onInit: onPlaygroundInit,
-    onStateChanged
-  } = props;
+    exporter: exporterConfig,
+    viewerProperties,
+    editorProperties,
+    viewerAdditionalModules,
+    editorAdditionalModules,
+    propertiesPanel: propertiesPanelConfig,
+    apiLinkTarget
+  } = config;
 
   const {
     display: displayActions = true
-  } = actionsConfig;
+  } = actionsConfig || {};
 
-  const paletteContainerRef = useRef();
   const editorContainerRef = useRef();
-  const formContainerRef = useRef();
-  const dataContainerRef = useRef();
-  const resultContainerRef = useRef();
+  const paletteContainerRef = useRef();
   const propertiesPanelContainerRef = useRef();
+  const viewerContainerRef = useRef();
+  const inputDataContainerRef = useRef();
+  const outputDataContainerRef = useRef();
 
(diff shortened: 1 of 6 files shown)


packages/form-js-playground/src/components/PlaygroundRoot.js
```diff
@@ -193,59 +165,77 @@ export function PlaygroundRoot(props) {
...
-    });
-  }, [ onStateChanged, schema, data ]);
 
-  const handleDownload = useCallback(() => {
+    if (!apiLinkTarget) {
+      return;
+    }
+
+    apiLinkTarget.api.getState = () => ({ schema, data });
+    apiLinkTarget.api.load = load;
```
douglasbouttell-camunda, line 234: So we have two ways to load data: through the initial props and this api method.

packages/form-js-playground/src/components/PlaygroundRoot.js
```diff
@@ -193,59 +165,77 @@ export function PlaygroundRoot(props) {
...
+  // initialize data through props
   useEffect(() => {
-    dataEditorRef.current.setValue(toString(initialData));
-  }, [ initialData ]);
+    if (!config.initialSchema) {
+      return;
+    }
+
+    load(config.initialSchema, config.initialData || {});
+  }, [ config.initialSchema, config.initialData, load ]);
```
douglasbouttell-camunda, line 186: I would have these be the default values of the `useState` calls.

```js
const [ schema, setSchema ] = useState(config.initialSchema);
const [ data, setData ] = useState(config.initialData || {});

useEffect(() => {
  if (!schema) { 
    return; 
  }    
  formEditorRef.current.importSchema(schema, data);
  inputDataRef.current.setValue(toString(data));
  if (inputDataContainerRef.current) {
    const variables = getSchemaVariables(schema);
    inputDataRef.current.setVariables(variables);
  }
}, [ data, schema, getSchemaVariables ]);
```

This means you can git rid of the effect on 188-190, 192-197 and all `load` does it set the states.

douglasbouttell-camunda (commented):
Nothing major. Happy to discuss if any of them are useful.

packages/form-js-playground/src/components/PlaygroundRoot.js
Skaiir: Yes exactly. I agree this isn't great and not react-like, but I didn't want to make any changes to any of the logic.

packages/form-js-playground/src/components/PlaygroundRoot.js
```diff
@@ -193,59 +165,77 @@ export function PlaygroundRoot(props) {
...
+      getResultView: () => outputDataRef.current,
+      getSchema: () => formEditorRef.current.getSchema(),
+      saveSchema: () => formEditorRef.current.saveSchema(),
+      setSchema: setSchema,
+      setData: setData
+    };
 
+  }, [ apiLinkTarget ]);
+
+  // separate effect for state to avoid re-creating the api object every time
```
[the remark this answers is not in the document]
Skaiir: I mostly did this because the state changes constantly, while the rest of the API really only gets set up once ideally.

packages/form-js-playground/src/components/PlaygroundRoot.js
```diff
@@ -193,59 +165,77 @@ export function PlaygroundRoot(props) {
...
+      get: (name, strict) => formEditorRef.current.get(name, strict),
+      getDataEditor: () => inputDataRef.current,
+      getEditor: () => formEditorRef.current,
+      getForm: () => formViewerRef.current,
+      getResultView: () => outputDataRef.current,
+      getSchema: () => formEditorRef.current.getSchema(),
+      saveSchema: () => formEditorRef.current.saveSchema(),
+      setSchema: setSchema,
+      setData: setData
+    };
```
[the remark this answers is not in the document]
Skaiir: I guess it's mostly a preference thing. I just found the callback harder to read personally. 

In any case, this class isn't exported at the moment so it's mostly just an implementation detail.

douglasbouttell-camunda (approved):

PR #1103, opened 2024-03-21
feat: do not re-instantiate JSON editors on every change

### LINKED ISSUE #1101: JSON Editors scroll around when we don't want them to
### Describe the Bug

grafik

The following editors scroll a bunch whenever data changes. Since most data changes do no affect the structure, it would be better not to change the scroll position every time as it makes it quite difficult to watch a data-point change.

### Steps to Reproduce

1. start the playground demo
2. scroll halfway in the output data panel
3. modify some data through the preview form
4. the data editor jumps around

### Expected Behavior

4. the data editor would stay on the state we're currently having a look at

Skaiir:
Closes #1101

### DIFF
--- a/packages/form-js-playground/src/components/JSONEditor.js
+++ b/packages/form-js-playground/src/components/JSONEditor.js
@@ -1,5 +1,4 @@
 import mitt from 'mitt';
-
 import { basicSetup } from 'codemirror';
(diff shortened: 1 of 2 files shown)


Skaiir:
This is an incomplete solution especially with the addition of the scripting component. We need to be able to differentiate programmatic changes from user changes if we want to introduce a jump to feature.

PR #1117, opened 2024-04-04
fix: make expression result evaluation more reliable

### LINKED ISSUE #1114: Expression results get wiped during form editing
### Describe the Bug

Expression results are not recalculated every time when moving form elements around in the editor, see the [modeler issue](https://github.com/camunda/camunda-modeler/issues/4227) for details.

Skaiir:
Closes #1114


packages/form-js-viewer/src/render/components/form-fields/ExpressionField.js
```diff
@@ -1,12 +1,13 @@
 import { useCallback, useEffect } from 'preact/hooks';
-import { useExpressionEvaluation, useDeepCompareMemoize, useService, useEffectOnChange } from '../../hooks';
```
vsgoulart, line 2: `useEffectOnChange` is not used anymore, let's delete it

Skaiir: Nicely spotted, removed it :)

vsgoulart (approved):

barmac:
Hi, I noticed that this PR doesn't contain any tests. How do we verify that the feature doesn't break again?

Issue #1120, opened 2024-04-04
Refactor datetime component


Skaiir:
### What should we do?

We should clean up the datetime component and simplify it as much as possible.

### Why should we do it?

Because it is too much code, too hard to read and not really the most robust either.

PR #1121, opened 2024-04-04
fix: prevent datetime from validating early

### LINKED ISSUE #1099: Required date field shows "Field is required." on form exhibition
### Describe the Bug

When displaying a form that contains a required date field for user input, the date field displays a validation message "Field is required.", whereas the other fields do not.

### Steps to Reproduce

1 - Create a form with two fields: one of type text and the other of type date, both required.
2 - Display the form for user input.

### Expected Behavior

Both fields should show no validation message upon the initial display of the form.

### Environment

- Host Chrome 122
- OS: Windows 11
- Library version: @bpmn-io/form-js@1.7.1

Skaiir:
Closes #1099

The root cause issue for this is the fact that the datetime component has a lot of funky synchronization code that leads to onChange being called when it initializes. The simplest and quickest way to fix this, is to not call onChange when newValue = oldValue, and it makes sense generally. 

But ideally, we wouldn't want this logic branch to be called on init at all. Now I'm not sure I can fix this without a full refactor of datetime, and this component is a house of cards so my suggestion is, we take the easy fix for now, and I've raised [this issue](https://github.com/bpmn-io/form-js/issues/1120) to provide a refactor to the component in the future.

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Datetime.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Datetime.js
@@ -15,7 +15,7 @@ import { Timepicker } from './parts/Timepicker';
 
 import { formFieldClasses, prefixId } from '../Util';
 import { sanitizeDateTimePickerValue } from '../util/sanitizerUtil';
-import { parseIsoTime, serializeDate, serializeDateTime, serializeTime } from '../util/dateTimeUtil';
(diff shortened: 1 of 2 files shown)


Skaiir:
I've got some test coverage for this in https://github.com/bpmn-io/form-js/pull/1124

douglasbouttell-camunda (approved):

PR #1140, opened 2024-04-10
fix(styles): Don't clip radio buttons

douglasbouttell-camunda:
Closes https://github.com/camunda/tasklist/issues/3688

----

Do not clip Labels. This causes the Carbon radio button to clip.


vsgoulart:
closing if favor of https://github.com/bpmn-io/form-js/pull/1261

PR #1143, opened 2024-04-10
feat: implement formFieldInstanceRegistry

### LINKED ISSUE #1142: Create a better way to work with individual formField instances
### What should we do?

We should create a module which tracks individual formfield instances in the viewer.

### Why should we do it?

At the moment, it is incredibly difficult to programmatically interact with individual elements of a form outside of the preact render cycle. So things like validation force us to write rather complicated algorithms which basically just recomputes what preact gives us for free. We should, as part of the preact render cycle, expose an API which "saves" all that work rendering and can be used across the app for things like validation, expression evaluations, ect...

Skaiir:
Closes #1142

### DIFF
--- a/packages/form-js-viewer/src/Form.js
+++ b/packages/form-js-viewer/src/Form.js
@@ -204,64 +204,38 @@ export class Form {
    * @returns {Errors}
    */
   validate() {
-    const formFields = this.get('formFields'),
-          formFieldRegistry = this.get('formFieldRegistry'),
-          pathRegistry = this.get('pathRegistry'),
+    const formFieldRegistry = this.get('formFieldRegistry'),
+          formFieldInstanceRegistry = this.get('formFieldInstanceRegistry'),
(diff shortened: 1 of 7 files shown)


douglasbouttell-camunda (approved):

packages/form-js-viewer/src/Form.js
```diff
@@ -204,64 +204,38 @@ export class Form {
...
+    });
 
-    const workingErrors = {};
-    validateFieldRecursively(workingErrors, formFieldRegistry.getForm());
-    const filteredErrors = this._applyConditions(workingErrors, data, { getFilterPath: getErrorPath, leafNodeDeletionOnly: true });
-    this._setState({ errors: filteredErrors });
+    this._setState({ errors });
 
-    return filteredErrors;
+    // @ts-ignore
```
vsgoulart, line 237: @Skaiir Can we fix the types instead of ignoring them?

PR #1155, opened 2024-04-16
fix: update validation to use formfield instances and proper contexts

### LINKED ISSUE #1147: Validation feel expressions do not respect local context
### Describe the Bug

image

At the moment we evaluate feel expressions in the validation parameters not against the local context of the field being validated, but the root data context. This is incorrect and makes using feel validations in dynamic lists not work as expected.

### Steps to Reproduce

1. add a feel validation on a child of a dynamic list
2. it is not possible to use this.value, or any of the context related values here

### Expected Behavior

2. feel expressions should evaluate within the local context

Skaiir:
Closes #1147

### DIFF
--- a/packages/form-js-playground/test/custom/viewer.js
+++ b/packages/form-js-playground/test/custom/viewer.js
@@ -21,7 +21,6 @@ function RangeRenderer(props) {
 
   const onChange = ({ target }) => {
     props.onChange({
-      field,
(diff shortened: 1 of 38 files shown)


douglasbouttell-camunda (approved):
Looks good to me. Some minor comments around validation functions.

PR #1170, opened 2024-04-30
ci: add workflow to sync v2 with develop

vsgoulart:


.github/workflows/MERGE_DEVELOP_TO_V2.yml
```diff
@@ -0,0 +1,33 @@
+name: Merge develop to
```
douglasbouttell-camunda, line 1: Think this should read "Merge develop to v2"

.github/workflows/MERGE_DEVELOP_TO_V2.yml
```diff
@@ -0,0 +1,33 @@
...
+      - 'develop'
+
+jobs:
+  Merge_main_to_develop:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout develop
```
douglasbouttell-camunda, line 14: I think this should read "Checkout v2 branch"

.github/workflows/MERGE_DEVELOP_TO_V2.yml
vsgoulart: yep, fixed both errors

douglasbouttell-camunda (approved):

Issue #1166, opened 2024-04-25
`form-js-viewer` refactor, splitting view and state management


Skaiir:
### What should we do?

We need to split the `form-js-viewer` state management logic from its rendering in two different projects.

### Why should we do it?

There are many motivations for this change:

- Our reliance on preact to compute things like feel expressions, visibility, ect... is slowing us down more and more, many timing issues are created as a result of tethering render to computations and it also affects performance
- Our tests are unreliable and slow from their reliance on jsdom and waiting for UI event propagation
- Splitting the two would allow us to implement brand new renderers by interchanging
- ...

### How we are going to do it?

Internal discussions ongoing

Skaiir:
We are currently considering to build two sub-packages to handle the separation. These would be independent from each-other which means that you could either use the core package with a custom renderer, or the renderer package with any type of schema corresponding to the visual contract we want to build up between the two packages. 

I briefly sketched out how we see the core architecture's direction (early ideas anyways). But we could opt for a near-stateless transformation pipeline that would look something like this:

![Sketching render pipeline](https://github.com/bpmn-io/form-js/assets/17801113/a796f757-68ff-416f-a74d-06ed97b12b2b)

PR #1174, opened 2024-05-06
v2: bootstrap viewer render project

Skaiir:
Related to #1166

### DIFF
--- a/packages/form-js-viewer-core/package.json
+++ b/packages/form-js-viewer-core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@bpmn-io/form-js-viewer-core",
   "version": "1.8.6",
-  "description": "View forms - powered by bpmn.io",
+  "description": "Manage form state - powered by bpmn.io",
   "type": "module",
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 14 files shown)


packages/form-js-viewer-core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@bpmn-io/form-js-viewer-core",
   "version": "1.8.6",
-  "description": "View forms - powered by bpmn.io",
+  "description": "Manage form state - powered by bpmn.io",
```
douglasbouttell-camunda, line 4: We'll probably need to discuss what should go here but what is here just now is fine.

douglasbouttell-camunda (approved):

packages/form-js-viewer-render/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "name": "bpmn.io contributors",
+    "url": "https://github.com/bpmn-io"
+  },
+  "sideEffects": false,
+  "files": [
+    "dist"
+  ],
+  "devDependencies": {
+    "@preact/preset-vite": "^2.8.2",
+    "vitest": "^1.5.3"
```
vsgoulart, line 37: @Skaiir We can install this on the project root since we don't need them to run as a peer dependency this way we avoid users installing these packages when the user installs `@bpmn-io/form-js-viewer-render`

vsgoulart (commented):
@Skaiir I was going to mention on the meeting today, but I think we should focus more on the core for now and let it more library agnostic but it also doesn't hurt already bootstrapping this package

PR #1178, opened 2024-05-16
fix: revert to min-dom@4.1.0

vsgoulart:
Closes https://github.com/bpmn-io/min-dom/issues/22


vsgoulart:
There's an issue with the lockfile, reverting it back to draft to investigate

douglasbouttell-camunda (approved):

PR #1201, opened 2024-06-06
feat: enable passing of documentation link to form properties panel

Skaiir:
Related to https://github.com/camunda/team-hto/issues/630

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/PropertiesPanel.js
+++ b/packages/form-js-editor/src/features/properties-panel/PropertiesPanel.js
@@ -6,8 +6,10 @@ import { reduce, isArray } from 'min-dash';
 
 import { FormPropertiesPanelContext } from './context';
 
-import { PropertiesPanelHeaderProvider } from './PropertiesPanelHeaderProvider';
+import { getPropertiesPanelHeaderProvider } from './PropertiesPanelHeaderProvider';
 import { PropertiesPanelPlaceholderProvider } from './PropertiesPanelPlaceholderProvider';
+
+const EMPTY = {};
 
 export function PropertiesPanel(props) {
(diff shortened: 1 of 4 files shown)


packages/form-js-editor/src/features/properties-panel/PropertiesPanel.js
```diff
@@ -83,6 +85,17 @@ export function PropertiesPanel(props) {
...
+  const formFields = getService('formFields');
+
+  const PropertiesPanelHeaderProvider = useMemo(
+    () =>
+      getPropertiesPanelHeaderProvider({
+        getDocumentationRef: propertiesPanelConfig.getDocumentationRef,
+        formFields,
+      }),
+    [formFields, propertiesPanelConfig],
+  );
```
vsgoulart, line 97: Was this memo really necessary? If you didn't see any noticeable performance issue without it we should remove to avoid memory leaks

vsgoulart (approved):

packages/form-js-editor/src/features/properties-panel/PropertiesPanel.js
Skaiir: Nothing is necessary but not having this means we re-render the whole header every render cycle. Which isn't ideal. It won't lag, yet, but could cause problems down the line.

Basically, this header provider stuff is a config, and we need to keep the reference to the config stable, hence the memo.

vsgoulart: Just rerendering is not an issue, as long as rerendering doesn't take long. Also, not all rerenders are committed to the DOM which is usually the most expensive.

But overusing memoization might cause memory leaks as explained [here](https://schiener.io/2024-03-03/react-closures)

PR #1202, opened 2024-06-06
chore: removed deprecated `shouldSort` properties panel attributes

### LINKED ISSUE #1200: Clear out usages of shouldSort
### What should we do?

The `shouldSort` property of the properties panel has been deprecated, let's clean it out.

### Why should we do it?

To stay up to date with the standards of the bpmn.io project, and clear some console warnings ;)

Skaiir:
Closes #1200

See https://github.com/bpmn-io/properties-panel/commit/bb1cfb05a41f3e7632ace7f34927e198e43ad775 if you want to look into why this was done. But basically, list autoSorting was deemed an antifeature. It is completely gone from the properties panel and only throws a log warning. This PR clears those warnings.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@babel/core": "^7.24.5",
         "@babel/plugin-transform-react-jsx": "^7.23.4",
         "@babel/plugin-transform-react-jsx-source": "^7.24.1",
-        "@bpmn-io/properties-panel": "^3.18.2",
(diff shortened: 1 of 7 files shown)


vsgoulart (approved):
@Skaiir The changes are ok, but there's a failing test

Skaiir:
@vsgoulart Could you try `npm run all`ing the branch locally, would be quite a bit of help because I can't replicate, but the errors are happening every time.

PR #1211, opened 2024-06-14
fix: style pre blocks

### LINKED ISSUE #867: Code formatting isn't styled in the markdown renderer
### Describe the Bug

Whenever we use markdown formatting to display code blocks, the text appears pretty normal, with none of the expected code block styling.

image

### Steps to Reproduce

1. Add a text view
2. Write some markdown that should render to code blocks
3. The text is only mildly changed, but it's almost not recognizable that we're in a code block.

### Expected Behavior

We expect something like this:

image

douglasbouttell-camunda:
Closes #867 


Skaiir:
Nice one :D

Issue #1205, opened 2024-06-08
Right to Left support for elements


okaeiz:
### Is your feature request related to a problem? Please describe.

Yes. I am using form-js in Camunda Desktop Modeler. When the form is in Hebrew, Persian or Arabic, the placeholders, labels and values are rendered left to right. It is frustrating for RTL users. I request the RTL support for forms or form elements separately.

### Describe the solution you'd like

I think there are 2 ways to achieve this:
1. Add a direction option to the form properties.
2. Add a direction option to every form element such as text field, text area, datetime, etc.

### Describe alternatives you've considered

After the implementation of form-js HTML View element, the HTML tags for templating and representing data (such as h1, h2, a, li, etc.) could be modified and assigned CSS attributes. I managed to change the font and direction in HTML View. But apparently the use of javascript for passing input values in HTML View is disabled due to security reasons. In other words, no inputs can be used in HTML View. Therfore, I still have to use text field and text area which are not customizable. 

### Additional context

I have also raised issues in Camunda platform offical repositories. The RTL support seems to be a good idea and it definitely is; because it attracts a large number of users who speak an RTL language to the communtiy,

Skaiir:
So, in terms of how I see this implementation going:


The functionality should be provided through configuration, so you'd instantiate your form like this:

```
const form = new Form({ data, schema, i18n: { rtl: true } );
```

Then you can access this config via 

```
const config = useService('config');
const { i18n } = config || {};
const { rtl = false };
```

and set some conditions on the base fields to define this rtl behavior.

okaeiz:
@Skaiir Thank you. I played around with the code and this is what I achieved. I created a `StyleGroup` since the `Appearance` name was already used for another group. Then I added a `DirectionEntry` to the group using which I can select the direction.


https://github.com/bpmn-io/form-js/assets/100111649/0f6a5ac1-5736-4c31-80c3-f8c14474d2ac


As you can see, the `Button` component is rendered according to what the user has selected. I am implementing the same fetching logic for all the components. Based on the earlier discussions under [this issue](https://github.com/camunda/camunda-modeler/issues/4306), I intend to add an extra entry to the Text Field component which can overwrite the global settings of direction for possible LTR values such as emails and URLs. Am I on the right track?

Skaiir:
[...]

PR #1212, opened 2024-06-16
RTL direction support via properties panel (part 1)

okaeiz:
Has worked on #1205

- [ ] This PR adds a new form-js group named `StyleGroup` to the properties panel that is activated for the form component (`default` type).
- [ ] This PR also adds a `DirectionEntry` to the newly-added `StyleGroup` which is a global option that sets the direction of the form components.
- [ ] Within this PR, only the following components adhere to `DirectionEntry`:
  - Checkbox
  - Checkbox group
  - Button
  - Text Field (I will work on another entry for this component which can overwrite the `DirectionEntry` value)
 - [ ] If the direction is `rtl`, the font is changed to [Vazirmatn ](https://github.com/rastikerdar/vazirmatn)which has been discussed before in [Camunda ](https://github.com/camunda/camunda/issues/19102)main repository. The font change is only applied on the so-called form components.


okaeiz:
I made some changes before configuring my Git user; can we skip the CLA signature for `root`?

Skaiir:
@okaeiz We can, but major thing right now is to clean up the PR, I'll review the code afterwards. 

You might want to reset to bpmn-io:develop, and then cherry pick your changes into it. I see some duplicate commits, and also a tons of merge commits which we don't allow in PRs.

Skaiir:
If the PR isn't yet in a finished state, we usually make use of the draft functionality of github :)

PR #1219, opened 2024-07-03
fix: Prevent drag and drop from accessing prop when undefined

vsgoulart:
Closes https://github.com/bpmn-io/form-js/issues/1218

### DIFF
--- a/packages/form-js-editor/src/features/dragging/Dragging.js
+++ b/packages/form-js-editor/src/features/dragging/Dragging.js
@@ -78,8 +78,8 @@ export class Dragging {
     let targetParentId;
 
     if (formFieldNode) {
-      formField = this._formFieldRegistry.get(formFieldNode.dataset.id);
-      columns = (formField.layout || {}).columns;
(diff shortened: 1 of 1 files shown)


packages/form-js-editor/src/features/dragging/Dragging.js
```diff
@@ -78,8 +78,8 @@ export class Dragging {
     let targetParentId;
 
     if (formFieldNode) {
-      formField = this._formFieldRegistry.get(formFieldNode.dataset.id);
-      columns = (formField.layout || {}).columns;
+      formField = this._formFieldRegistry.get(formFieldNode.dataset.id) || { layout: {} };
+      columns = formField.layout.columns;
```
vsgoulart, line 82: @Skaiir this bug seems to be only happening on the Desktop modeler on Windows, I couldn't reproduce it locally but this is the issue if you follow the stack trace

Could you give it a try and reproduce it on your machine? (It should be ok to merge this anyway)

PR #1226, opened 2024-07-30
Offset Date value with timezone offset to set the correct Datetime component value when its subtype is Date

### LINKED ISSUE #1223: Date only field endless countdown when form initialized with data
### Describe the Bug

When initializing a form with some data and the form contains a date only field (a component of type `datetime` and subtype `date`) and the component value is passed in the form data (for example `2024-07-24`), an endless countdown of the dates begins and never stops.

All library versions after 1.7.1 behave like this. v1.7.1 stops the countdown at some date that is not correct.

**IMPORTANT DETAIL**: this bug is only (?) reproducible with time zone **`(UTC-06:00) Central Time (US & Canada)`**. Works OK with Bulgarian timezone (UTC+02:00) but all our clients are in the US and use the `Central Time` TZ

### Steps to Reproduce

1. Copy the following code in some `index.html` file:
```
<html>
<head>
  <link rel="stylesheet" href="https://unpkg.com/@bpmn-io/form-js@1.9.0/dist/assets/form-js.css">
</head>
<body>
  <div id="form"></div>
  <script src="https://unpkg.com/@bpmn-io/form-js@1.9.0/dist/form-viewer.umd.js"></script>
  <script>
    async function init() {
      const container = document.querySelector('#form');
      let form = await FormViewer.createForm({
        container,
        schema: { components: [], type: 'default' }
      });

      const schema = {
        "components": [
          {
            "subtype": "date",
            "dateLabel": "Date",
            "label": "Date time",
            "type": "datetime",
            "layout": {
              "row": "Row_0d18n2g",
              "columns": null
            },
            "id": "Field_1lwhnt4",
            "key": "datetime_a5ocm9"
          },
          {
            "action": "submit",
            "label": "Button",
            "type": "button",
            "layout": {
              "row": "Row_0a8960z",
              "columns": null
            },
            "id": "Field_190m4tx"
          }
        ],
        "type": "default",
        "id": "Form_0238ts0",
        "exporter": {
          "name": "form-js (https://demo.bpmn.io)",
          "version": "1.8.3"
        },
        "schemaVersion": 16
      };
      await form.importSchema(schema, { datetime_a5ocm9: '2024-07-24' });
    }

    init();
  </script>
</body>
</html>
```
2. Open the `index.html` file with a browser
3. Observe that the date input changes its value all the time counting down

### Expected Behavior

The input doesn't change its value all the time and it's correctly populated with the date that's passed in the `data` object.

### Environment

- Host (Browser/Node version), if applicable: Chrome 127
- OS: Windows 11
- Library version: 1.9.0 (but every version >= 1.7.1 has the bug)

vladislav-karamfilov:
Closes #1223

This PR fixes the above issue which is valid for TZs with negative offset. Previously, when the `Date` object was constructed it was for the previous date for TZ is with negative offset and this triggered the endless date countdown described in the linked issue. The fix takes into account the TZ offset of the populated date and adds it to the `Date` object always resulting in the correct date for TZs with negative and positive offsets.

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Datetime.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Datetime.js
@@ -86,7 +86,9 @@ export function Datetime(props) {
 
     switch (subtype) {
       case DATETIME_SUBTYPES.DATE: {
-        date = new Date(Date.parse(value));
+        const dateTimestamp = Date.parse(value);
+        const timezoneOffsetMinutes = new Date(dateTimestamp).getTimezoneOffset();
+        date = new Date(dateTimestamp + timezoneOffsetMinutes * 60 * 1000);
         break;
       }
       case DATETIME_SUBTYPES.TIME: {


Skaiir:
@vladislav-karamfilov We've got something high priority currently but we'll take a look at your contribution pretty soon, thanks for providing this fix, I never was quite able to reproduce this bug but aware of it :)

Skaiir:
Simplified it slightly in https://github.com/bpmn-io/form-js/pull/1234, and changed the commit messages.

vladislav-karamfilov:
> Simplified it slightly in #1234, and changed the commit messages.

Great! Thanks, @Skaiir! Do you know when will this fix be included in an upcoming release?

Skaiir:
@vladislav-karamfilov There are other defects I need to handle but it doesn't take much time for a release so expect 1.9.1 in 10-15 minutes, there will be a 1.9.2 in the coming weeks as well.

vladislav-karamfilov:
> @vladislav-karamfilov There are other defects I need to handle but it doesn't take much time for a release so expect 1.9.1 in 10-15 minutes, there will be a 1.9.2 in the coming weeks as well.

Great! Thanks, @Skaiir!

PR #1237, opened 2024-08-16
fix: disallow drop when no associated field is found

Skaiir:
Closes #1219


vsgoulart (approved):

vsgoulart:
@Skaiir Could we write a test case for this one?

Issue #1239, opened 2024-08-18
[Implement] Create file input component


vsgoulart:

Skaiir:
Closed via https://github.com/bpmn-io/form-js/pull/1246

PR #1240, opened 2024-08-19
feat: bootstrap filepicker

vsgoulart:
Related to #1239


packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+export function FilePicker() {
+  return null;
+}
+
+FilePicker.config = {
+  type: 'filepicker',
+  keyed: true,
+  label: 'File picker',
+  group: 'basic-input',
```
Skaiir, line 12: Unsure about if this is basic, but for now doesn't matter much to me.

Skaiir (approved):

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
vsgoulart: On the mocks it's basic, in the end it will be just simple HTML file input but we can always change it easily

PR #1244, opened 2024-08-21
feat: add filepicker properties panel

vsgoulart:
Related to #1239

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/entries/AcceptEntry.js
+++ b/packages/form-js-editor/src/features/properties-panel/entries/AcceptEntry.js
@@ -1,0 +1,65 @@
+import { get } from 'min-dash';
+
+import { useService, useVariables } from '../hooks';
+
+import { FeelTemplatingEntry, isFeelEntryEdited } from '@bpmn-io/properties-panel';
+
+export function AcceptEntry(props) {
+  const { editField, field } = props;
+
+  const entries = [];
+
+  entries.push({
+    id: 'accept',
+    component: Accept,
+    editField: editField,
+    field: field,
+    isEdited: isFeelEntryEdited,
+    isDefaultVisible: (field) => field.type === 'filepicker',
+  });
+
+  return entries;
+}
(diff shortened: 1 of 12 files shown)


packages/form-js-editor/src/features/properties-panel/entries/AcceptEntry.js
```diff
@@ -0,0 +1,65 @@
...
+  return FeelTemplatingEntry({
+    debounce,
+    element: field,
+    getValue,
+    id,
+    label: 'Supported file formats',
+    singleLine: true,
+    setValue,
+    variables,
+    description,
```
Skaiir, line 50: A regex validation would be good to have here. Especially if you plan to inject it into the HTML. I'm sure you're going to sanitize it somewhere down the line, but it may become confusing for a user if say they use a ; or something and it doesn't work without a properties panel error.

packages/form-js-editor/src/features/properties-panel/entries/MultipleEntry.js
```diff
@@ -0,0 +1,43 @@
...
+
+  const getValue = () => {
+    return get(field, path, '');
+  };
+
+  const setValue = (value) => {
+    return editField(field, path, value);
+  };
+
+  return ToggleSwitchEntry({
```
Skaiir, line 35: Wouldn't it make sense to feel enable this? It might help creating more generic forms.

vsgoulart: I'll do it

packages/form-js-editor/src/features/properties-panel/entries/AcceptEntry.js
Skaiir: The validation here is less about the risk and more about the UX in my view. If there is any risk to be honest it needs to be handled later anyways because of the feel variables.

Skaiir (approved):

PR #1245, opened 2024-08-23
feat: implement basic filepicker on viewer

vsgoulart:
Related to 1239 #

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
@@ -1,8 +1,86 @@
+import { formFieldClasses } from '../Util';
+import { Label } from '../Label';
+import { Errors } from '../Errors';
+import { useRef, useState } from 'preact/hooks';
+import { useSingleLineTemplateEvaluation } from '../../hooks';
+
+const type = 'filepicker';
+
 /**
+ * @typedef Props
+ * @property {(props: { value: string }) => void} onChange
+ * @property {string} domId
+ * @property {string[]} errors
+ * @property {boolean} disabled
+ * @property {boolean} readonly
+ * @property {boolean} required
+ * @property {Object} field
+ * @property {string} field.id
+ * @property {string} [field.label]
+ * @property {string} [field.accept]
+ * @property {boolean} [field.multiple]
+ *
+ * @param {Props} props
  * @returns {import("preact").JSX.Element}
  */
-export function FilePicker() {
-  return null;
+export function FilePicker(props) {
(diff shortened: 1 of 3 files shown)


packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
```diff
@@ -1,8 +1,86 @@
...
+    <div class={formFieldClasses(type, { errors, disabled, readonly })}>
+      <Label htmlFor={domId} label={label} required={required} />
+      <input
+        type="file"
+        className="fjs-hidden"
+        ref={fileInputRef}
+        id={domId}
+        name={domId}
+        multiple={evaluatedMultiple === false ? undefined : evaluatedMultiple}
+        accept={evaluatedAccept === '' ? undefined : evaluatedAccept}
```
Skaiir, line 48: One issue I see here is that this accept does not enforce the file type. You can still navigate to 'All files' and it'll accept it in the upload. Something that will probably need doing, although not crucial right now.

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
```diff
@@ -1,8 +1,86 @@
...
+ * @param {Props} props
  * @returns {import("preact").JSX.Element}
  */
-export function FilePicker() {
-  return null;
+export function FilePicker(props) {
+  /** @type {import("preact/hooks").Ref<HTMLInputElement>} */
+  const fileInputRef = useRef(null);
+  /** @type {[File[],import("preact/hooks").StateUpdater<File[]>]} */
+  const [selectedFiles, setSelectedFiles] = useState([]);
```
Skaiir, line 30: Can be handled later, but I can see that the file uploader caches the values during form import and reset. This might not matter hugely now because things are going to change once the files are properly handled, but this would clear it:

```
  eventBus.on('import.done', () => setSelectedFiles([]));
  eventBus.on('reset', () => setSelectedFiles([])));
```

Skaiir (approved):
Couple of things I noted, aside from that looks good. Fix now or later at your discretion.

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
vsgoulart: Yeah, but this is on the OS level. I researched it and it looks like we have no control over that

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
vsgoulart: I guess it doesn't hurt to add it

PR #1246, opened 2024-08-25
feat: implement file registry on filepicker

vsgoulart:
Related to  #1239


packages/form-js-viewer/src/Form.js
```diff
@@ -168,9 +168,12 @@ export class Form {
 
     const errors = this.validate();
 
+    const files = this.get('fileRegistry').getAllFiles();
```
Skaiir, line 171: I think this needs to be filtered based on the available instances.  Otherwise it'll submit hidden files as well. 

Lazy pseudocode:

```
const filteredInstances = instances
  .filter(instance => instance.type === 'filepicker')
  .map(instance => instance.value);

const filteredFiles = files.filter(file => filteredInstances.includes(file.value));
```

But **first** there's the issue of how these files are keyed (see other comment).

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
```diff
@@ -26,19 +26,28 @@ const type = 'filepicker';
...
   const eventBus = useService('eventBus');
+  /** @type {import('../../FileRegistry').FileRegistry} */
+  const fileRegistry = useService('fileRegistry', false);
   const { field, onChange, domId, errors = [], disabled, readonly, required } = props;
   const { label, multiple = '', accept = '', id } = field;
   const evaluatedAccept = useSingleLineTemplateEvaluation(accept);
   const evaluatedMultiple =
     useSingleLineTemplateEvaluation(typeof multiple === 'string' ? multiple : multiple.toString()) === 'true';
   const errorMessageId = `${domId}-error-message`;
+  const filesKey = `${id}_value_key`;
```
Skaiir, line 41: Keying needs to be done differently. It needs to include the full path otherwise dynamic list keys will overwrite eachother. The id is unique only at the design level. The true id of an element at the render level is the full path / key of the element.

image

My prefered fileKey would be something like `fileRef::${simpleHash(fullPath)}`. Hashing is optional but it's more about the fact that it's mental overhead for the users for it to be actually readable. Also I know some people will try to reverse engineer the path from it which is not something I'd like to encourage.

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
```diff
@@ -71,24 +80,29 @@ export function FilePicker(props) {
...
+          fileRegistry.setFiles(filesKey, files);
+
           setSelectedFiles(files);
         }}
       />
       <div className="fjs-filepicker-container">
         <button
           type="button"
-          disabled={disabled}
+          disabled={disabled || readonly || fileRegistry === null}
```
Skaiir, line 103: Just FYI, at the moment it's still possible to submit a file in the editor. I guess you're planning on adding a custom editor renderer for this anyways?

Skaiir (changes requested):

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
Skaiir: If not, needs fixing for sure. I don't know if it's related to the label or something but when I click around on it in the editor, it opens the window, which can lead to an exception.

packages/form-js-viewer/src/Form.js
vsgoulart: I deleted the files when the component unmounts, this the logic is encapsulated on the component

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
vsgoulart: As I mentioned through Slack I couldn't reproduce it, maybe we can check it together tomorrow

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
vsgoulart: I didn't do the hashing, I'm not sure if I agree with the mental overhead point.
Anyway, this can be easily fixed tomorrow.

packages/form-js-playground/test/spec/form.json
```diff
@@ -2,256 +2,37 @@
   "$schema": "../../../form-json-schema/resources/schema.json",
   "components": [
     {
-      "type": "expression",
```
Skaiir, line 5: This is the main start:playground schema, probably committed by error?

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
```diff
@@ -40,21 +40,29 @@ export function FilePicker(props) {
...
   const filesKey = `file::${calculateValuePath(props.fieldInstance.valuePath)}`;
+  const deleteFiles = useCallback(() => {
+    if (fileRegistry) {
+      fileRegistry.deleteFiles(filesKey);
+    }
+  }, [fileRegistry, filesKey]);
 
   useEffect(() => {
-    const reset = () => {
+    return () => {
```
Skaiir, line 52: The problem with this is that by design, hiding operations are not destructive, so that if someone hides something by accident they don't lose their progress. So I think we'll have to go the route I initially suggested as otherwise we orphan file references and have different behavior for the file picker and regular components.

image

packages/form-js-viewer/src/render/components/form-fields/FilePicker.js
Skaiir: I guess on this one we can defer to @volodymyr-melnykc for a decision.

Skaiir (changes requested):

packages/form-js-viewer/src/features/repeatRender/index.js
```diff
@@ -1,8 +1,10 @@
 import { RepeatRenderManager } from './RepeatRenderManager';
+import { FileRegistry } from '../../render/FileRegistry';
 
 export const RepeatRenderModule = {
   __init__: ['repeatRenderManager'],
   repeatRenderManager: ['type', RepeatRenderManager],
+  fileRegistry: ['type', FileRegistry],
```
Skaiir, line 7: Just for future reference, we're registering this twice which I guess was due to a timing issue, but the modules access eachother. So probably here you needed either a `__depends__` or to re-order some things.

Skaiir (approved):

PR #1247, opened 2024-08-25
feat: add filepicker carbon styles

vsgoulart:
Related to #1239


Skaiir (approved):
Looks good :ok_hand:

PR #1248, opened 2024-08-26
fix: dynamic list collapsing issue

### LINKED ISSUE #1230: Dynamic lists truncate when they aren't expanded
### Describe the Bug

There is currently an issue where dynamic lists will not propagate data if they have not been expanded. 

### Steps to Reproduce

1. open the playground
2. add a  dynamic list and some inner field (say, keyed to `myItems` & `itemName`)
3. set the input data to a list of more than 5 items:

```
{
  "myItems": [
    {
      "itemName": "a"
    },
    {
      "itemName": "b"
    },
    {
      "itemName": "c"
    },
    {
      "itemName": "d"
    },
    {
      "itemName": "e"
    },
    {
      "itemName": "f"
    },
    {
      "itemName": "g"
    }
  ]
}

4. (optional) if the dynamic list is not collapsed at this point, collapse it
5. change the value of the last item
6. any item beyond the 5th isn't expanded (excepted) but also isn't available in the output data (very wrong)

```

### Expected Behavior

6. the collapsed or expanded states of a dynamic list should be purely visual and have no bearing on the output data

### Related

internal: https://jira.camunda.com/browse/SUPPORT-23067?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D

Skaiir:
Closes #1230

### DIFF
--- a/packages/form-js-viewer/assets/form-js-base.css
+++ b/packages/form-js-viewer/assets/form-js-base.css
@@ -1189,6 +1189,10 @@
   color: var(--color-warning);
 }
 
+.fjs-container .fjs-repeat-row-collapsed {
+  display: none;
+}
+
 .fjs-container .fjs-repeat-render-footer {
   display: flex;
   flex-direction: row;
--- a/packages/form-js-viewer/src/features/repeatRender/RepeatRenderManager.js
+++ b/packages/form-js-viewer/src/features/repeatRender/RepeatRenderManager.js
@@ -59,6 +59,7 @@ export class RepeatRenderManager {
     const showRemove = repeaterField.allowAddRemove && hasChildren;
 
     const displayValues = isCollapsed ? values.slice(0, nonCollapsedItems) : values;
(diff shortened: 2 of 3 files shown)


packages/form-js-viewer/test/spec/render/components/form-fields/Dynamiclist.spec.js
```diff
@@ -75,6 +75,21 @@ describe('Dynamic List', () => {
...
+      nonCollapsedItems: 1,
+    };
+
+    // when
+    const { container } = createDynamicList({ field });
+
+    // then
+    const collapsedItems = container.querySelectorAll('.fjs-repeat-row-collapsed .fjs-form-field-textfield');
+    expect(collapsedItems).to.have.length(2);
+  });
```
vsgoulart, line 91: This test is not really guaranteeing that we're outputting the correct data, the ideal test would be to just check the output data. It it possible to write a test like that?

I would just delete this test if it's not possible since this test is just testing the implementation details of the dynamic list

vsgoulart (approved):

PR #1261, opened 2024-09-11
fix(styles): Don't clip radio buttons

vsgoulart:
Related to https://github.com/bpmn-io/form-js/pull/1140

Closes https://github.com/camunda/camunda/issues/19945

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Checklist.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Checklist.js
@@ -68,12 +68,12 @@ export function Checklist(props) {
     <div class={classNames(formFieldClasses(type, { errors, disabled, readonly }))} ref={outerDivRef}>
       <Label label={label} required={required} />
       {loadState == LOAD_STATES.LOADED &&
-        options.map((o, index) => {
+        options.map((option, index) => {
           const itemDomId = `${domId}-${index}`;
-          const isChecked = hasEqualValue(o.value, values);
+          const isChecked = hasEqualValue(option.value, values);
 
           return (
-            <div className="fjs-inline-label" key={index}>
+            <div className="fjs-inline-label" key={option.value}>
               <input
                 checked={isChecked}
                 class="fjs-input"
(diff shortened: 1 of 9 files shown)


Skaiir (changes requested):
The clipping is intentional for long labels. With this change you can totally break the structure unfortunately. And since it can be driven by variables we don't really want that. I'm not sure if it's possible to properly do clipping but also have the label wrap the checkbox in the DOM. 

I think we need to split them, just generally, even if it means adding a nesting node to flexbox them.

packages/form-js-viewer/src/render/components/form-fields/Checklist.js
```diff
@@ -73,13 +73,7 @@ export function Checklist(props) {
...
 
           return (
-            <Label
-              htmlFor={itemDomId}
-              label={o.label}
-              class={classNames({
-                'fjs-checked': isChecked,
-              })}
-              required={false}>
+            <div className="fjs-inline-label" key={index}>
```
vsgoulart, line 76: Do not use indexes as keys, it can cause problems.
Let's use some variation of the `domId`

packages/form-js-viewer/src/render/components/form-fields/Checklist.js
```diff
@@ -94,7 +88,15 @@ export function Checklist(props) {
                 aria-invalid={errors.length > 0}
                 aria-describedby={[descriptionId, errorMessageId].join(' ')}
               />
-            </Label>
+              <Label
+                htmlFor={itemDomId}
+                label={o.label}
```
vsgoulart, line 93: Let's use the opportunity to improve this code and renamed the variable to not use a single character name, let's use `option` instead of `o`

packages/form-js-viewer/src/render/components/form-fields/Radio.js
```diff
@@ -69,12 +69,7 @@ export function Radio(props) {
           const isChecked = isEqual(option.value, value);
 
           return (
-            <Label
-              htmlFor={itemDomId}
-              key={index}
-              label={option.label}
-              class={classNames({ 'fjs-checked': isChecked })}
-              required={false}>
+            <div className="fjs-inline-label" key={index}>
```
vsgoulart, line 72: Let's improve the key here too

vsgoulart (commented):
@Skaiir There are some problems with the Carbonization, there's no padding between radio items and the checked state is not working on checkboxes or radio buttons anymore

PR #1278, opened 2024-09-26
deps: bump puppeteer

vsgoulart:
Closes #


vsgoulart:
@Skaiir These are the tests failing
image

It looks like the the error is coming from the `moveDragging` util [here](https://github.com/bpmn-io/form-js/blob/437b63ad7b76cb13ee8fbf6e70bbcb6afe975027/packages/form-js-editor/test/spec/FormEditor.spec.js#L941)

I couldn't really find why these elements have no  `tagName`

PR #1283, opened 2024-09-29
Update lockfile preactversion to 10.15.1 and fix null default options serialization issue

### LINKED ISSUE #1282: Default value <none> is serialized as "<none>" in the schema when using preact@10.15.1
### Describe the Bug

Whenever using more recent versions of preact, the default value is serialized as "<none>" in the schema, we want it to be serialized as no value at all.

### Steps to Reproduce

1. update preact in the lockfile to 10.15.1
2. create a select component
3. set default value to none
4. inspect schema
5. "\<none\>" is serialized

### Expected Behavior

5. nothing should be serialized

Skaiir:
Closes #1282


Skaiir:
@vsgoulart I changed how we do UI tests with the userEvent library (instead of sending discrete events), because preact 10.6.3 had [this change](https://github.com/preactjs/preact/pull/3354) which made a ton of tests fail. 

I haven't removed all the instances of fireEvent, mostly those which failed. But I will try to get rid of fireEvent everywhere it makes sense in a follow-up PR.

packages/form-js-viewer/test/spec/render/components/form-fields/Taglist.spec.js
```diff
@@ -667,8 +673,8 @@ describe('Taglist', function () {
 
         // when
         const filterInput = taglist.container.querySelector('.fjs-taglist-input');
-        fireEvent.focus(filterInput);
-        fireEvent.keyDown(filterInput, { key: 'Enter', code: 'Enter' });
+        await userEvent.click(filterInput);
+        await userEvent.keyboard('{enter}');
```
vsgoulart, line 677: I think enter is `capitalized`, could you check if it's really working or if it doesn't make any difference?

```suggestion
        await userEvent.keyboard('{Enter}');
```

vsgoulart (approved):
We will have to check if nothing will break on the web modeler because of conflicts in the Preact version but you said it's within our update range so it should be ok

PR #1288, opened 2024-10-02
fix: set radio name to properly imply tabindexes in firefox

### LINKED ISSUE #1287: Radio button issues: tabbing inconsistency across browsers
### Describe the Bug

There are two issues:

- When we tab across radio buttons not in firefox, the focus jumps and skips consecutive radio buttons
- When we tab across radio buttons in firefox, the focus is handled for every radio element

### Steps to Reproduce

1. start the viewer
2. open it in firefox
3. tab across a radio component
4. each radio element gets selected
5. put two radio components in the viewer test file after one-another
6. start the viewer
7. open in in chrome
8. tab across them
9. the second (and all following) radio buttons gets skipped entirely

I think this is probably due to the fact that radio components aren't named in either case which is why I made this a single issue. 

### Expected Behavior

(4) => radio elements should not be tabbed through
(9) => only a single radio group should be tabbed over

Skaiir:
Closes #1287


packages/form-js-viewer/test/spec/Form.spec.js
```diff
@@ -1593,6 +1594,33 @@ describe('Form', function () {
...
+    it('should tab once into radio group from the incomming direction', async function () {
+      // given
+      await bootstrapForm({
+        container,
+        schema: radioTabbingSchema,
+      });
+
+      // then
+      const textarea1 = screen.getByLabelText('Textarea 1');
+      await userEvent.click(textarea1);
```
vsgoulart, line 1607:
```suggestion
      await userEvent.click(screen.getByLabelText('Textarea 1'));
```

packages/form-js-viewer/test/spec/Form.spec.js
```diff
@@ -1593,6 +1594,33 @@ describe('Form', function () {
...
+      await userEvent.tab();
+      expect(screen.getByLabelText('Radio button 1')).to.equal(document.activeElement);
+
+      await userEvent.tab();
+      expect(screen.getByLabelText('Radio button 4')).to.equal(document.activeElement);
+
+      const textarea2 = screen.getByLabelText('Textarea 2');
+      await userEvent.click(textarea2);
```
vsgoulart, line 1615:
```suggestion
      await userEvent.click(screen.getByLabelText('Textarea 2'));
```

vsgoulart (approved):

vsgoulart (approved):

PR #1292, opened 2024-10-03
fix: do not create a simple label in datetime components

### LINKED ISSUE #1268: Remove `label` property on the datetime component
### What should we do?

Currently we have the `label` property exported in the datetime component but this property is not used anywhere except on the properties panel with the static value `Date time`.

We should remove this property and use the existing ones `dateLabel` and `timeLabel`

### Why should we do it?

To keep the exported schema clean with only relevant data

Skaiir:
Closes #1268

### DIFF
--- a/packages/form-js-editor/src/features/properties-panel/PropertiesPanelHeaderProvider.js
+++ b/packages/form-js-editor/src/features/properties-panel/PropertiesPanelHeaderProvider.js
@@ -13,6 +13,9 @@ export function getPropertiesPanelHeaderProvider(options = {}) {
       if (headerlessTypes.includes(type)) {
         return '';
       }
+      if (type === 'datetime') {
+        return field.dateLabel || field.timeLabel;
+      }
       if (type === 'text') {
         return textToLabel(field.text);
       }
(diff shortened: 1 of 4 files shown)


packages/form-js-editor/src/features/properties-panel/PropertiesPanelHeaderProvider.js
```diff
@@ -13,6 +13,9 @@ export function getPropertiesPanelHeaderProvider(options = {}) {
       if (headerlessTypes.includes(type)) {
         return '';
       }
+      if (type === 'datetime') {
+        return field.dateLabel || field.timeLabel;
```
vsgoulart, line 17: I wonder if we shouldn't merge both labels if they're both filled

vsgoulart (approved):

packages/form-js-editor/src/features/properties-panel/PropertiesPanelHeaderProvider.js
Skaiir: I think just the one is good enough for quick identification, which at the end of the day is all this little subheading is about (making sure you're changing the think you think you're working on). Maybe we can but it makes the headers kinda long so I'm happy leaving it like that personally.

PR #1300, opened 2024-10-10
Soft deprecate config labels

### LINKED ISSUE #1291: Soft deprecate the concept of label in config and split its responsabilties

Currently, we have a unified concept of label that takes on two different responsibilities. It defines (in most cases) the default label a component is using. It also defines the "name" a component uses in the UI. Especially because of that second reason, we need to define it on every component, which is confusing for components that don't have a label.

So for clarity and allowing us to control those two behaviors independently, I would propose that we split it into these two configs:
- componentName or simply name
- defaultLabel

To keep backwards compatibility with potential custom components, but move forwards, I think we should:
- still use config.label as a fallback in all current usages
- work on the type definitions to deprecate the property in linters

Keep it as-is?

### Additional context

Somewhat related to or to be worked on with https://github.com/bpmn-io/form-js/issues/1290
This allows us to get rid of the hardcoded solution for datetime label added here https://github.com/bpmn-io/form-js/pull/1292

Skaiir:
Closes #1291

I ended up trying to keep it minimal. Still keeping the isNew flag, but there's only one place where it HAD to be used. We are safe to default the labels now because in any case, if it's not new, the label should be '' (not undefined).

### DIFF
--- a/packages/form-js-editor/test/spec/core/FieldFactory.spec.js
+++ b/packages/form-js-editor/test/spec/core/FieldFactory.spec.js
@@ -310,12 +310,12 @@ describe('core/FieldFactory', function () {
 
 // helpers //////////////
 
-function testCreate(options, isNew = true) {
+function testCreate(options, isNewField = true) {
   const { type, label, keyed = false, initial = {}, expected = {} } = options;
 
   return inject(function (fieldFactory) {
     // when
-    const field = fieldFactory.create({ type, ...initial }, isNew);
+    const field = fieldFactory.create({ type, ...initial }, isNewField);
 
     // then
     expect(field.id).to.exist;
--- a/packages/form-js-editor/test/spec/core/FieldFactory.spec.js
+++ b/packages/form-js-editor/test/spec/core/FieldFactory.spec.js
@@ -13,7 +13,7 @@ describe('core/FieldFactory', function () {
       testCreate({
         type: 'button',
         label: 'Button',
-        defaults: {
+        expected: {
           action: 'submit',
         },
       }),
@@ -34,7 +34,7 @@ describe('core/FieldFactory', function () {
         type: 'checklist',
         label: 'Checkbox group',
         keyed: true,
-        defaults: {
+        expected: {
           values: [
             {
               label: 'Value',
@@ -50,7 +50,7 @@ describe('core/FieldFactory', function () {
       testCreate({
         type: 'default',
         keyed: false,
-        defaults: {
+        expected: {
           components: [],
         },
       }),
@@ -71,7 +71,7 @@ describe('core/FieldFactory', function () {
         type: 'radio',
         label: 'Radio group',
         keyed: true,
-        defaults: {
+        expected: {
           values: [
             {
               label: 'Value',
@@ -88,7 +88,7 @@ describe('core/FieldFactory', function () {
         type: 'select',
         label: 'Select',
         keyed: true,
-        defaults: {
+        expected: {
           values: [
             {
               label: 'Value',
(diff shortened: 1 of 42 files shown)


packages/form-js-viewer/src/core/FieldFactory.js
```diff
@@ -60,14 +60,13 @@ export class FieldFactory {
...
-    // cf. https://github.com/bpmn-io/form-js/issues/1291
-    const shouldApplyLabel = applyDefaults && !(config.type === 'datetime') && config.label;
-
-    const field = config.create({
-      ...(shouldApplyLabel ? { label: config.label } : {}),
-      ...attrs,
-    });
+    const field = config.create(
+      {
+        ...(config.label ? { label: config.label } : {}),
```
Skaiir, line 65: applyDefaults removed here as some long time ago, we made the change to serialize empty labels to '' instead of undefined (so that they would persist across loads). Hence this wasn't doing anything to properly formatted schemas.

packages/form-js-viewer/src/render/components/form-fields/Datetime.js
```diff
@@ -210,14 +210,17 @@ export function Datetime(props) {
...
   group: 'basic-input',
   emptyValue: null,
   sanitizeValue: sanitizeDateTimePickerValue,
-  create: (options = {}) => {
+  create: (options = {}, isNew) => {
     const defaults = {};
     set(defaults, DATETIME_SUBTYPE_PATH, DATETIME_SUBTYPES.DATE);
-    set(defaults, DATE_LABEL_PATH, 'Date');
+
+    if (isNew) {
```
Skaiir, line 221: Here we need isNew so that date label doesn't get created for time components.

packages/form-js-viewer/src/render/components/form-fields/Datetime.js
```diff
@@ -210,14 +210,17 @@ export function Datetime(props) {
 Datetime.config = {
   type,
   keyed: true,
-  label: 'Date time',
+  name: 'Date time',
   group: 'basic-input',
   emptyValue: null,
   sanitizeValue: sanitizeDateTimePickerValue,
-  create: (options = {}) => {
+  create: (options = {}, isNew) => {
```
vsgoulart, line 217: I just think we should give a more descriptive name

```suggestion
  create: (options = {}, is NewLabel) => {
```

packages/form-js-viewer/src/core/FieldFactory.js
```diff
@@ -13,7 +13,7 @@ export class FieldFactory {
     this._formFields = formFields;
   }
 
-  create(attrs, applyDefaults = true) {
+  create(attrs, isNew = true) {
```
vsgoulart, line 16:
```suggestion
  create(attrs, isNewLabel = true) {
```

packages/form-js-editor/test/spec/core/FieldFactory.spec.js
```diff
@@ -302,12 +310,12 @@ describe('core/FieldFactory', function () {
 
 // helpers //////////////
 
-function testCreate(options, applyDefaults = true) {
-  const { type, label, keyed = false, defaults = {} } = options;
+function testCreate(options, isNew = true) {
```
vsgoulart, line 313:
```suggestion
function testCreate(options, isNewLabel = true) {
```

vsgoulart (approved):

packages/form-js-viewer/src/render/components/form-fields/Datetime.js
Skaiir: I'll rename it all to isNewField (the fact that the label is new is only an implementation detail of the fact that the field is new).

PR #1309, opened 2024-10-24
fix: fix filepicker component merge error

vsgoulart:


packages/form-js-playground/src/components/PlaygroundRoot.js
```diff
@@ -140,7 +140,8 @@ export function PlaygroundRoot(config) {
 
     // pipe viewer changes to output data editor
     formViewer.on('changed', updateOutputData);
-    formViewer.on('formFieldInstanceRegistry.changed', updateOutputData);
+    formViewer.on('formFieldInstance.added', updateOutputData);
```
Skaiir, line 143: I think these were reverted?

Skaiir (approved):

Issue #1327, opened 2024-12-05
[Implement] Create document preview component


vsgoulart:
Epic https://github.com/camunda/product-hub/issues/2592, part of the Document handling https://github.com/camunda/product-hub/issues/2555
Design mocks: [link](https://www.figma.com/design/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?node-id=1078-729)

### What should we do?

Following up on #1239 would like to also add the possibility to preview and download files in a form

- Create a new file preview component according to [the mocks](https://www.figma.com/design/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?node-id=1078-729)
  - This new component will have the following configurations on the properties panel: `Label`, `Document reference` and optionally configure the `Endpoint`
    - `Label`: a title shown on the top of the list documents
    - `Document reference`: the variable in the context which `form-js` will use to render the files
    - `Endpoint`: the key in the context which will have the endpoint URL which will be used to fetch each file. By default, it is set to the Camunda API endpoint.
  - Each file should be downloadable
  - PDFs and images should be previewable for any other format we will show the filename

### Why should we do it?

So users can consume files uploaded in other forms or services

volodymyr-melnykc:
Design for the document preview component in forms:

https://www.figma.com/design/J78Djf3PgCEAPxFD86floX/Form-Editor-Components?node-id=1078-729&t=sC5vhpadMvDHwpWT-1

marcosgvieira:
@vsgoulart I believe this one is only missing documentation, correct?

vsgoulart:
@marcosgvieira The PR is open: https://github.com/camunda/camunda-docs/pull/4773

marcosgvieira:
@vsgoulart I believe we can close this one, is that correct?

PR #1329, opened 2024-12-09
feat: bootstrap document preview element

vsgoulart:
This PR bootstraps the new new `Document preview` element. It adds the new component to the palette, updates the JSON schema definition and sets up the the properties panel for it.

The next PR will include the component runtime logic.

Related to #1327


vsgoulart:
@barmac @nikku Could you or someone from the modelling team help with code reviews since we're a bit short staffed on the HTO team?

nikku:
@vsgoulart I'll hopefully get to this PR tomorrow :crossed_fingers:

nikku (commented):
Technical feedback:

* From the looks of it this feature integrates nicely with the way other features (components) are built
  * Integrates with editor, viewer + schema
  * Built in the "standard" way
  * Schema version bumped
* I see a lack of test coverage on the editor, i.e. don't see where we verify this works properly (configuration yields right form schema)
  * :arrow_right: At this point I assume you tested the E2E workings manually, rigidly.
* No test coverage in the viewer
  * :arrow_right: At this point I assume you tested the E2E workings manually, rigidly.
  
UI / UX feedback:

I verified if this works in the playground via

```
npm run start form-playground
```

* It seems not to be properly integrated, I can pull it in, but don't see an (editor) preview, I don't see a viewer preview either:

image

nikku:
I ask @Skaiir to also have a look, as you are a more knowledgable person in the realm of `form-*`.

vsgoulart:
@nikku This is just the first half of the feature to make it easier to review. For now it's just the modelling part and then the next PR will have the runtime implementation on a second PR.

That's why I'm targeting a feature branch too instead of develop too

Skaiir (approved):
One small thing, we don't need a new `title` there is already a way to change the name of the label property based on file types.

See `function getLabelText(type)`

Aside from that, great work :rocket:

vsgoulart:
Custom `title` schema prop was removed via the last the commit

PR #1332, opened 2024-12-12
feat: add runtime implementation of document preview component

### LINKED ISSUE #1327: [Implement] Create document preview component

vsgoulart:
Add runtime implementation of the document preview component and missing design fixes

Closes #1327

### DIFF
--- a/packages/form-js-carbon-styles/src/carbon-styles.scss
+++ b/packages/form-js-carbon-styles/src/carbon-styles.scss
@@ -1395,3 +1395,24 @@
     }
   }
 }
+
+// Document preview ////////////
+
+.fjs-container .fjs-form-field-documentPreview {
+  .fjs-documentPreview-single-document-container {
+    border-radius: 0;
+  }
+
+  .fjs-documentPreview-download-button {
+    border-radius: 0;
+    background-color: var(--cds-background);
+    border-color: var(--cds-button-tertiary);
+    color: var(--cds-button-tertiary);
+    cursor: pointer;
+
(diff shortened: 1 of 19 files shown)


Skaiir:
Something potentially to follow up on:

These error states are quite tiny and could benefit of some iconography or something.
image

Skaiir (approved):
I think the visuals can be polished but this delivers a bunch of value already and is plenty consistent. Code quality is good as always.

Issue #1341, opened 2024-12-19
Remove content hash injection from document preview


vsgoulart:
### What should we do?

On https://github.com/bpmn-io/form-js/issues/1340 we created a feature to inject a content hash into the document's URL.
This was created as a temporary fix to make the document preview work on Camunda Tasklist with the document service.

Because this is a Camunda specific feature we should try to find a way to remove it, an option would be to change Camunda Tasklist to have a BFF to manage this hash from the backend.

### Why should we do it?

To keep `form-js` as generic as possible

PR #1342, opened 2024-12-19
feat: Inject content hash into document preview URL

### LINKED ISSUE #1340: Inject content hash into URL into document preview URL
### What should we do?

In order for the document preview to work together with Camunda Tasklist we must inject a content hash as query param in the document URL for security reasons.
This content hash will be provided via the metadata and must respect the following shape:
```ts
{
  documentId: string,
  metadata: {
    contentType: string,
    fileName: string,
    customProperties?: {
      contentHash?: string
    }
  }
}
```

The query param will respect the follow pattern: `http://<endpoint with documentId>?contentHash=<hash>`

This solution should be a temporary solution because it's specific to Camunda's APIs and we should try to keep `form-js` as generic as possible. The best solution would be for Camunda Tasklist to have a BFF (backend for frontend) to manage this for the UI.

### Why should we do it?

So the document preview works with Camunda Tasklist

vsgoulart:
This PR adds the content hash injection into the document preview URL so it works with Camunda Tasklist.
This is a temporary fix and we should try to remove it as soon as we can for the reasons explained here: https://github.com/bpmn-io/form-js/issues/1341

Closes #1340

### DIFF
--- a/packages/form-js-playground/test/spec/Playground.spec.js
+++ b/packages/form-js-playground/test/spec/Playground.spec.js
@@ -88,6 +88,9 @@ describe('playground', function () {
           metadata: {
             fileName: 'My document.pdf',
             contentType: 'application/pdf',
+            customProperties: {
+              contentHash: '1234567890',
+            },
           },
         },
         {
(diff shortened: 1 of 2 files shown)


Skaiir:
Due to time constraints, it makes sense to approve this as-is, but the solution long term should involve creating an extension point via modules for the consuming library to override.

Skaiir (approved):

PR #1356, opened 2025-02-11
fix: form input autocomplete

vsgoulart:
This PR fixes the autocompletion on the form input, before we were just completing the variable names which resulted in invalid JSON.

Closes https://github.com/camunda/camunda-modeler/issues/4758


Skaiir (approved):
Since we don't touch completions that often, I think this is fine, but in my view having a proper separation of getting the cursor's "postional state", and then having a map of related completions would be nicer and more readable.

Feel free to implement or not depending on time, but if you see future changes it would be nice to set this structure in place from the start.

PR #1375, opened 2025-03-18
feat: remove Camunda special document endpoint logic

vsgoulart:
When we added the document preview component we added some Camunda specific logic for building the document URLs.
With this PR we're adding the option to pass a custom module which the user can use to add their own logic to build URLs

You can find an example of the Tasklist custom module [here](https://github.com/camunda/camunda/pull/29762/files#diff-7d178fc7a72885f5438db7d665d0ac026719884722229839bcac08233d3ace78)

Closes https://github.com/bpmn-io/form-js/issues/1341


packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
```diff
@@ -3464,27 +3464,6 @@ describe('properties panel', function () {
         });
       });
     });
-
-    describe('documentPreview', function () {
```
Skaiir, line 3468: Any reason this was removed altogether? One one of the properties was cleared, no?

Skaiir: Nevermind I get it.

Skaiir (approved):

nikku:
Great improvement! `#architecture`

PR #1393, opened 2025-04-11
feat: add description for multiple toggle

### LINKED ISSUE #1391: When using the File picker component, inform the user that the variable always returns an array
### What should we do?

When using the File picker component, the user should be informed that the result variable always returns an array of files, either the file picker is configured to support a single file upload or upload of multiple files.

A message is displayed in the Properties panel under the switcher "Support multiple files". Depending on the switcher state it describes that an array returns a single file or multiple files.

Image

### Why should we do it?

The File picker component always returns an array of files, even when configured for single file upload.
It can be not clear to users and has led to confusion in the past.

vsgoulart:
Closes https://github.com/bpmn-io/form-js/issues/1391

### DIFF
--- a/packages/form-js-editor/assets/form-js-editor-base.css
+++ b/packages/form-js-editor/assets/form-js-editor-base.css
@@ -839,3 +839,7 @@
 .gu-mirror.fjs-drag-row-move .cds--row {
   flex-wrap: nowrap !important;
 }
+
+.bio-properties-panel-description .bio-properties-panel-strong {
+  font-weight: bold;
+}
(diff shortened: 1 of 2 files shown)


vsgoulart:
@volodymyr-melnykc I made some small changes in the text, you can check it [here](https://demo-add-multiple-description--camunda-form-playground.netlify.app/)

volodymyr-melnykc:
@vsgoulart Thanks for the update. Having a message under the switcher looks good!

When a switcher is off, the message says "_The picker variable will return an array with single_" missing the wording "file" at the end. Can we update it to "_The picker variable will return an array with a single file_"?

Also, I noticed strange behavior with the switcher. Sometimes it doesn't change the state when clicking on it, or automatically switches even when not clicking on it. Here's a video recording:

https://github.com/user-attachments/assets/b0507941-612e-46b6-bb97-e8f4772fbb93

PR #1400, opened 2025-04-24
fix: explictly pass textfield change event target

### LINKED ISSUE #1399: Textfield in a form instantiated inside shadow dom
### Describe the Bug

When creating a form instance inside a shadow DOM node typing a value in a textfield won't work properly. The value typed simply disappears after some time.
This is due to some differences in the event handling inside shadow dom noes + React events + Preact events.

### Steps to Reproduce

1. Model a form with a textfield
2. Instantiate the form inside a shadow dom node ([repro here](https://codesandbox.io/p/devbox/cpvt3k))
3. Try to type inside the textfield
4. The value disappears after some time

### Expected Behavior

The textfield should work normally

vsgoulart:
This fixes an issue when a form is instantiated inside a shadow DOM node together with React. When typing inside a textfield the value would disappear after a while (after the debounce function runs)

I couldn't dig 100% to the root reason of this, but I believe this is due to the differences in the event management between shadow DOM together with the React/Preact event handling

This fix was suggested by customers on this [support ticket](https://jira.camunda.com/browse/SUPPORT-26988)

Closes #1399

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/Textfield.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/Textfield.js
@@ -19,11 +19,18 @@ export function Textfield(props) {
 
   const { required } = validate;
 
-  const [onInputChange, flushOnChange] = useFlushDebounce(({ target }) => {
+  const [onChange, flushOnChange] = useFlushDebounce(({ target }) => {
     props.onChange({
       value: target.value,
     });
   });
+
+  /**
+   * @param {import('preact').JSX.TargetedEvent<HTMLInputElement, Event>} event
+   */
+  const onInputChange = (event) => {
+    onChange({ target: event.target });
+  };
 
   const onInputBlur = () => {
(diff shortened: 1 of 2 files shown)


packages/form-js-viewer/src/render/components/form-fields/Textfield.js
```diff
@@ -19,12 +19,19 @@ export function Textfield(props) {
...
+  const [onChange, flushOnChange] = useFlushDebounce(({ target }) => {
     props.onChange({
       value: target.value,
     });
   });
 
+  /**
+   * @param {import('preact').JSX.TargetedEvent<HTMLInputElement, Event>} event
+   */
+  const onInputChange = (event) => {
```
vsgoulart, line 31: I couldn't really think of a good way to test this that wouldn't be too cumbersome since the errors happens with shadow dom + react

Skaiir (approved):

Issue #1207, opened 2024-06-10
Jalali calendar support for form date-time element


okaeiz:
### Is your feature request related to a problem? Please describe.

Yes. The Gregorian calendar - which is currently being used in the datetime component - is not useful in some regions. I request implementing a new component, the Jalali datetime.

### Describe the solution you'd like

I recommend using this eye-catching [Jalai datetime component](https://github.com/persian-tools/persian-mobile-datepicker) which is written in Javascript and can be implemented as a new input component.

### Describe alternatives you've considered

Another way to implement this functionality would be to add extra configuration to the existing datetime component. Maybe adding a dropdown to the General section of the properties panel and letting the users choose between Gregorian and Jalali dates.

### Additional context

This is the datepicker I intend to implement. I am not sure about the rendering procedure yet though.

![theme](https://github.com/bpmn-io/form-js/assets/100111649/a6aa8209-e63a-442e-bc84-6eb4c3d2129e)

okaeiz:
I took a look at [flatpickr](https://github.com/flatpickr/flatpickr) which is the datepicker library that is used in datetime component, and I found out there's a [Jalali support PR](https://github.com/flatpickr/flatpickr/pull/2623) already that is made by someone else and it seems to require a review. But the library seems inactive for months.

Is that what we are looking for here? Shall we contribute to that project? As an alternative, I was wondering maybe we could create a locale environment variable for modeler. So when it is set to `fa` (Persian), the default flatpickr datetime component would be inaccessible and a Jalali component would be enables and ready to use within the form. That's because most users may find it confusing if we let them choose between Gregorian and Jalali calendars. So that should be activated according to the locale. Does that sound like a good idea?

@Skaiir @volodymyr-melnykc

volodymyr-melnykc:
@Skaiir How would you recommend handling it? ☝️

okaeiz:
Any updates? @volodymyr-melnykc @Skaiir

PR #1406, opened 2025-07-04
feat: add jalali calendar

mohammad-moa:
In this PR, Jalali calendar has been added using a side library related to flatpickr called flatpickr-jalali-support.

To use it, you must use `locale: 'fa'` in the Datetime.js file in the path
packages/form-js-viewer/src/render/components/form-fields/Datetime.js

Similarly, in the Datepicker.js component in the path
packages/form-js-viewer/src/render/components/form-fields/parts/Datepicker.js
import Persian
```
import { Persian } from 'flatpickr-jalali-support/dist/l10n/fa';
```
the config variable must be used as follows
```
    let config = {
      allowInput: true,
      dateFormat: getLocaleDateFlatpickrConfig('fa'),
      static: true,
      clickOpens: false,
      locale: Persian,
      // TODO: support dates prior to 1900 (https://github.com/bpmn-io/form-js/issues/533)
      minDate: disallowPassedDates ? 'today' : '01/01/1900',
      errorHandler: () => {
        /* do nothing, we expect the values to sometimes be erroneous and we don't want warnings polluting the console */
      },
    };
```

Also to display the input placeholder, you should use
```
getLocaleReadableDateFormat('fa')
```

Related to [issue](https://github.com/bpmn-io/form-js/issues/1207)


AlekseyManetov:
Hey @mohammad-moa, thanks for the contribution!

Could you help me understand the intended usage — how would a user activate Jalali calendar mode? I believe it should be configurable, and doesn't break our current default behavior.

Also, could you add tests covering the expected configuration and behavior? That would help us understand the API surface and make sure things work end-to-end.

PR #1422, opened 2025-07-16
ci: refresh lockfile to fix memory leak

vsgoulart:
We were getting some memory leaks when building with Lerna. This PR refreshes the lockfile to fix the leaky dep


packages/form-js-viewer/test/spec/render/components/form-fields/Taglist.spec.js
```diff
@@ -570,24 +570,22 @@ describe('Taglist', function () {
...
-
         // then
-        let focusedItem = dropdownList.querySelector('.fjs-dropdownlist-item.focused');
-        expect(focusedItem.innerText).to.equal('Tag4');
+        const tag4Item = screen.getByText('Tag4');
+        expect(tag4Item.classList.contains('focused')).to.be.true;
 
         await userEvent.keyboard('{arrowdown}{arrowdown}');
 
         await waitFor(() => {
```
Skaiir, line 579: `findByText` could eliminate the need for the `waitFor`, right? I'm sure there's a few places like this so we don't need to fix it atomically, but maybe just a good practice to try and do this everywhere. Although I'm not interested in breaking everything right now :sweat_smile:

PR #1474, opened 2026-02-28
fix(number): address number field precision edge cases

### LINKED ISSUE #1021: Increment breaks when too many decimals are used in number fields
### Describe the Bug

When the number of decimals is very high, the number field increment and decrement breaks and gets stuck due to rounding errors. We ideally want to still let users control this, so we need to consider a solution that is more reliable.

Additionally, we are able to write a number much more complex than the data will allow for, which can be confusing UX. 

firefox_1CDQKAAjAo

This problem only exists when we don't use string based number fields, and are due to floating point precision limits.

### Steps to Reproduce

1. add a new number field
2. give it 100 decimal points
3. increment / decrement it
4. works fine (we should get x*e-100)
5. set the field to 1
6. increment and decrement
7. increment and decrementing is stuck **(bug)**
8. reset to 1
9. .111111111, keep adding more precision
10. we are able to define a number that is much more precise than the underlying value will allow for **(bug)**

### Expected Behavior

unsure yet

Skaiir:
Closes #1021

Adds:
- Warning when exceeding maximum numeric precision (:warning: we need to confirm this new warning pattern)
- Disables the stepper arrows when they'd break due to numeric precision issues  (like adding a 1e-6 step to 1e12)

### DIFF
--- a/packages/form-js-viewer/assets/form-js-base.css
+++ b/packages/form-js-viewer/assets/form-js-base.css
@@ -31,6 +31,8 @@
   --color-red-360-100-95: hsl(360, 100%, 95%);
   --color-red-360-100-97: hsl(360, 100%, 97%);
 
+  --color-orange-33-100-40: hsl(33, 100%, 40%);
+
   --color-white: hsl(0, 0%, 100%);
   --color-black: hsl(0, 0%, 0%);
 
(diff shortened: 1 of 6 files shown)


packages/form-js-viewer/assets/form-js-base.css
```diff
@@ -31,6 +31,8 @@
   --color-red-360-100-95: hsl(360, 100%, 95%);
   --color-red-360-100-97: hsl(360, 100%, 97%);
 
+  --color-orange-33-100-40: hsl(33, 100%, 40%);
```
vsgoulart: @Skaiir We need to check the Carbonization of this on Tasklist

nikku (commented):
@Skaiir what is the status here? If you cannot follow-up, then let's move to backlog - and ensure the issue is clearly linked + captures what is broken.

PR #1497, opened 2026-03-20
fix: allow dynamic-list item shorthand in text templates

saschabuehrle:
## Bug
Fixes #1496. Inside dynamic list items, `{{field}}` currently resolves only against top-level form data. Item properties are only available via `{{this.field}}`.

## Fix
- In `buildExpressionContext`, spread the current `this` item object into the expression context.
- Keep top-level form data precedence to avoid breaking existing forms where the same key exists globally.
- Added regression tests for both shorthand item access and conflict precedence behavior.

## Validation
- Built viewer package successfully (`npm run build --workspace @bpmn-io/form-js-viewer`).
- Ran viewer test suite in this environment; existing unrelated locale/time-based failures remain, while the new util/expressions assertions pass.

Greetings, saschabuehrle

### DIFF
--- a/packages/form-js-viewer/src/util/expressions.js
+++ b/packages/form-js-viewer/src/util/expressions.js
@@ -10,8 +10,11 @@ import { wrapObjectKeysWithUnderscores } from './simple';
 export function buildExpressionContext(context) {
   const { data, ...specialContextKeys } = context;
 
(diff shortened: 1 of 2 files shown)


barmac:
Hi @saschabuehrle, we'd need you to sign the CLA before we can review & merge your changes.

philippfromme:
@saschabuehrle could you sign the CLA so we can move forward?

PR #1504, opened 2026-04-21
fix(playground): drop unused second arg from domClasses calls

Skaiir:
## Summary

`min-dom`'s `classes` helper accepts a single element. The `document.body` second argument passed in four `domClasses(container, document.body)` calls inside `JSONEditor.js` is silently ignored — dead code since introduction.

## Root cause

Likely a misremembering of a different `min-dom` helper signature. The helper's implementation uses only the first arg, so behavior is unchanged; this is a cleanup rather than a visible bug fix.

## Test plan

- [ ] `npm test` in `packages/form-js-playground` stays green
- [ ] Manual: open the playground, edit JSON — editor class toggling (`fjs-json-editor`, `fjs-cm-no-lint`) still works


nikku:
Looks ready to merge :star:

PR #1511, opened 2026-05-11
fix(form): render number field with value 0 set after initial render

### LINKED ISSUE #1510: Number field renders empty when value updates to `0` after initial render
### Describe the Bug

When a number field's value transitions to `0` *after* the initial render (for example when input data is re-imported in the form playground, or when async data populates the form), the field displays as empty and the submitted output becomes `null`.

The root cause is a falsy-value check in the outer-value sync path: `(value && value.toString()) || ''` evaluates to `''` for `value === 0`, which then sanitizes to `null`.

### Steps to Reproduce

1. Open the playground with a form containing a number field with key `amount`.
2. Leave the input data empty initially.
3. Update the input data to `{ "amount": 0 }`.
4. Observe that the number field renders empty.
5. Observe that the output data shows `"amount": null`.

### Expected Behavior

The number field should render `0` and the output data should contain `"amount": 0`.

The same expectation applies to the `serializeToString` mode for arbitrary precision: the displayed and submitted value must preserve the incoming string (e.g. `"0"` or `"123.233333333333333333333"`).

### Environment

- Library version: 1.21.3

Skaiir:
Closes #1510

The number field used `(value && value.toString()) || ''` to mirror outer value changes into local state, which collapses `0` to `''` and then sanitizes to `null`. The bug only surfaces when the value transitions to `0` *after* the initial render (e.g. async data load or playground re-import), since `useState(value)` captures the initial value directly and bypasses the outer-sync path.

Extracted the conversion into `serializeToInputString` to keep the contract explicit: `null`/`undefined` map to `''`, everything else passes through `.toString()`.

### DIFF
--- a/packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
+++ b/packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
@@ -3,6 +3,7 @@ import * as sinon from 'sinon';
 import { act, fireEvent, render, screen, waitFor } from '@testing-library/preact/pure';
 
 import { query as domQuery } from 'min-dom';
+import { set } from 'min-dash';
 
 import { OPTIONS_SOURCES, OPTIONS_SOURCES_DEFAULTS } from '@bpmn-io/form-js-viewer';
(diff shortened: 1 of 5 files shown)


nikku (approved):
Thanks! Solid fix + test coverage.

Issue #1169, opened 2024-04-28
How to translate form-js


kimatelier622:
How to support  Internationalization (i18n) ?

### Environment

- Host (Browser/Node version), if applicable: [e.g. MS Edge 18, Chrome 69, Node 10 LTS]
- OS: [e.g. Windows 7]
- Library version: [e.g. 2.0.0]

Ken-Scofield:
@Skaiir, Can you provide a demo?

Skaiir:
@Ken-Scofield fyi Niklas no longer maintains the project officially (though of course this is open source so he's still free to contribute :D).

So, depends if you want to internationalize the app or the form itself.

For the app, we currently don't support internationalization. See https://github.com/bpmn-io/form-js/issues/1055 for discussions. We're also in the middle of a pretty large re-architecturing moving towards the next major so it's probably not the right time to start. But we have it in mind for the future. We would mimic how it is done in [bpmn-js](https://github.com/bpmn-io/bpmn-js).

itzhouze:
> We're also in the middle of a pretty large re-architecturing moving towards the next major so it's probably not the right time to start. But we have it in mind for the future. We would mimic how it is done in [bpmn-js](https://github.com/bpmn-io/bpmn-js).

Hello everyone, we are also interested in a translation solution and are about to consider creating a fork of the form-js project in order to incorporate a translation logic.

This would of course make little sense if you are considering incorporating a translation solution at the form-js level and are also working on a re-architecturing.

Is there any news about the major restructuring and when it will be ready for use?

Issue #1055, opened 2024-02-18
Request for Internationalization (i18n) Support to Facilitate Localization


JhonRiveroLinktic:
### Is your feature request related to a problem? Please describe.

Yes, the current limitation is that form-js does not support internationalization (i18n), making it challenging to localize forms for users in non-English speaking regions. This situation can be frustrating for developers aiming to create multilingual applications that serve a global audience. The lack of i18n support restricts the usability and accessibility of forms created with form-js for users who do not speak English.

### Describe the solution you'd like

I propose the addition of internationalization (i18n) support in form-js, which would include:

The ability to translate all user interface texts (field labels, error messages, buttons, etc.) into different languages.
A simple way for developers to add their own language files, making it easy to customize and extend the available translations based on their application's needs.
An option to change the language at runtime, allowing users to select their preferred language, enhancing the user experience and accessibility.
This feature would allow form-js to be more inclusive and accessible to a broader audience, significantly improving the user experience in multilingual settings.

### Describe alternatives you've considered

As an alternative, developers could manually override UI texts after the form has been rendered, but this approach is cumbersome and not scalable. It requires additional coding and maintenance effort, especially when dealing with complex forms or multiple languages.

Another alternative is to fork the form-js project and implement i18n support independently. However, this approach would mean diverging from the main project and missing out on future updates and improvements made by the original form-js team.

### Additional context

The addition of i18n support would align form-js with other modern web libraries and frameworks that prioritize accessibility and global usability. This feature would not only benefit developers looking to create more accessible and user-friendly applications but also support the global vision of web inclusivity.

christian-konrad:
Hi @JhonRiveroLinktic , thank you for opening this issue!

We are currently looking into i18n support from a broader lens.
I am curious about your current or targeted setup.

Do you use or plan to use form-js inside of Camunda, or in a custom application?
Do you already use i18n libraries? What library integration would you favor to be supported by form-js?

Skaiir:
My personal take on this, and thanks for raising this we definitely will be looking into this eventually, but we shouldn't stray for from something that has already been solved and we should simply follow bpmn-js's solution to this. 

What this means for this library would be to provide a translation module. Basically, wrap every user-facing string in a translate(string) call. And then allow for language packs to be supplied to form-js, and the currently active language to be configured.

[...]

PR #1514, opened 2026-05-12
Add translations support for viewer and editor

Senior-Master-Merger:
## Summary

This PR adds initial translations support for the form-js viewer and editor.

It allows passing translation-related configuration when creating viewer and editor instances.

Additionally, a new `language` configuration option was added to `Form.js` and `FormEditor.js`, allowing the language to be configured during initialization, e.g.:

```js
language: 'en'
```

A new method was also introduced:

```js
setLanguage(language)
```

This is my first attempt at integrating translations support into form-js. Feedback and suggestions for improvements or adjustments are very welcome.

## Follow-up

Some translations are still missing in the properties panel, as they originate from the separate properties panel repository.

This includes:
- the tooltip for the dropdown arrow icon (`Toggle section`)
- the tooltip shown for edited entries (`Section contains edits`)
- the tooltip for the list group item count (`List contains x items`)

These translations should likely be handled in a dedicated follow-up issue in the corresponding properties panel repository.


Skaiir:
@Senior-Master-Merger Hey there, thanks for the contribution. I had a quick look and it's a decent start, but I expect a few changes before we can introduce this. 

**Relevant issues:**
https://github.com/bpmn-io/form-js/issues/1055
https://github.com/bpmn-io/form-js/issues/1169

So what I would suggest to you is following the i18n approach taken in [bpmn-js](https://github.com/bpmn-io/bpmn-js).

Namely:
- **Don't bundle the dictionaries/translations proper inside of the application**, let that be externally provided. We don't tend to keep actual languages in the core tooling as it is costly to maintain and not something we can promise, so we separate it out into clearly "community-maintained" repos.
- **Don't make language a global variable**. This ties in with the following point but right now two form-js's living in the same page read from the same variable, which isn't well isolated.
- **No runtime API surface out of modules**. No `setLanguage` in the core API surface, the goal within the form-js library is to enable internationalization, not formally providing it, so we don't want to expose this kind of API in the core. Same as bpmn-js, the expectation is that these are supplied at module configuration time.
- There are some typos that say ge instead of de, not that relevant within this library though.

If you want to provide actual translations in the bpmn-io libraries, we might be able to set up a community-maintained repository like we did here: https://github.com/bpmn-io/bpmn-js-i18n.

The success metric to use here is, how can you make this possible to configure with the minimum impact on the current library, again I point to bpmn-js for examples as it does it in a very minimal way. As far as I've read from this, we're not too far. 

Anyways, thanks for the start on this, I think we're all interested to bring this in, so with a few iterations we should be able to bundle this.

Skaiir:
@nikku Can we confirm if we'd be willing to open up an i18n repo for community translations for forms if this contribution comes to maturity?

barmac:
The community translations could be collected in a similar way as for bpmn-js: https://github.com/bpmn-io/bpmn-js-i18n
Perhaps we want to have them in the monorepo so that there is no need for [cross-repository updates](https://github.com/bpmn-io/bpmn-js/blob/develop/package.json#L26).

Senior-Master-Merger:
> @Senior-Master-Merger Hey there, thanks for the contribution. I had a quick look and it's a decent start, but I expect a few changes before we can introduce this.
> 
> **Relevant issues:** #1055 #1169
> 
> So what I would suggest to you is following the i18n approach taken in [bpmn-js](https://github.com/bpmn-io/bpmn-js).
> 
> Namely:
> 
>     * **Don't bundle the dictionaries/translations proper inside of the application**, let that be externally provided. We don't tend to keep actual languages in the core tooling as it is costly to maintain and not something we can promise, so we separate it out into clearly "community-maintained" repos.
> 
>     * **Don't make language a global variable**. This ties in with the following point but right now two form-js's living in the same page read from the same variable, which isn't well isolated.
> 
>     * **No runtime API surface out of modules**. No `setLanguage` in the core API surface, the goal within the form-js library is to enable internationalization, not formally providing it, so we don't want to expose this kind of API in the core. Same as bpmn-js, the expectation is that these are supplied at module configuration time.
> 
>     * There are some typos that say ge instead of de, not that relevant within this library though.
> 
> 
> If you want to provide actual translations in the bpmn-io libraries, we might be able to set up a community-maintained repository like we did here: https://github.com/bpmn-io/bpmn-js-i18n.
> 
> The success metric to use here is, how can you make this possible to configure with the minimum impact on the current library, again I point to bpmn-js for examples as it does it in a very minimal way. As far as I've read from this, we're not too far.
> 
> Anyways, thanks for the start on this, I think we're all interested to bring this in, so with a few iterations we should be able to bundle this.

@Skaiir Hi, thanks for the detailed feedback.

I've tried to incorporate several of the points you raised and would appreciate another review when you have time.

Regarding translations themselves, I think a community-maintained repository would be very beneficial. I'd be happy to maintain the German translations there and contribute additional improvements over time.

I'd be interested to hear whether the current direction is closer to what you had in mind and if there are any remaining concerns before this could be considered for inclusion.

Thanks again for the guidance and for taking another look.

mstein11:
Very cool addition. :-) 
Is there a timeline for a release of this feature?

Senior-Master-Merger:
@mstein11 Thanks! Unfortunately, I don't know yet.

I sent my response to the requested changes a few weeks ago and am currently waiting for the next update. Hopefully we'll hear something soon.

AlekseyManetov:
@Senior-Master-Merger Thanks for the follow-up, this is much closer. The main points from @Skaiir's review are addressed: no bundled dictionaries, no global language variable, no `setLanguage` in the public API.  Coverage is broad and consistent, `aria-label` and `title` included.

A few things to rework.

### Must fix

**1. `Dragging.js` — don't translate `'horizontal'` / `'vertical'`.** These are dragula API values, not user text. A German dictionary would translate the words and break drag & drop. Keep them as plain literals. Also, `translate` is passed in through `createDragulaInstance(options)`, but `Dragging` already has it injected — that plumbing can go.

**2. `Palette.js` — don't translate `entry.type` in the search filter.** The type (`textfield`, `checkbox`) is an identifier, not user text. Once a dictionary has an entry for it, searching by the technical name stops working. Keep `simplifyString(entry.type)` as is.

**3. Translation keys must be static literals, with runtime values as placeholders.** A template literal can't be grepped out of the source, which is how these dictionaries get built — and with a filename in the key, no dictionary can cover it at all. The PR already does this correctly everywhere else, e.g. in `Validator.js` and `FieldFactory.js`:

```js
translate(`Download ${fileName}`)              // before
translate('Download {fileName}', { fileName }) // after
```

Two places need it: `DocumentPreview.js` (`Download …`) and `DeleteActionProvider.js` (`Remove …`). The palette entry title is a third — see the open question below.

**4. `Validator.js` — one full sentence per key.** `runNumberValidation` builds one message from three `translate` calls and adds an English `'s'` outside the translation. Other languages can't assemble a correct sentence from fragments:

```js
if (decimalDigits === 0) {
  errors.push(translate('Value is expected to be an integer.'));
} else if (decimalDigits === 1) {
  errors.push(translate('Value is expected to have at most 1 decimal digit.'));
} else {
  errors.push(translate('Value is expected to have at most {value} decimal digits.', { value: decimalDigits }));
}
```

**5. `OPTIONS_SOURCES_LABELS` — breaking API change.** It's a public export of `@bpmn-io/form-js-viewer`, so object → function means consumers doing `OPTIONS_SOURCES_LABELS[source]` silently get `undefined`.  Keep the object untranslated and translate in `OptionsSourceSelectEntry`.

**6. Tests are missing.** All test changes are mock plumbing — nothing checks that a custom `translate` changes anything. One viewer spec and one editor spec would be enough, and would have caught points 1, 2 and 4.

### Should fix

- Unrelated changes, please drop: `.idea/` in `.gitignore`, the Prettier reflow of the CI workflows and `eslint.config.mjs`, stray blank-line deletions, and 69 removed lines in `package-lock.json`.
- `Validator.js`: `.toString()` on `min` but not `max`, and `` translate(`Field must match pattern {value}.`) `` uses backticks with no interpolation.
- `getService` threading is half applied — in `GeneralGroup` some entries get it, some don't. Pick one way (`useService('translate')` looks like the rule).

### Open question

`PaletteEntry.js` has `` translate(`Create ${getIndefiniteArticle(type)} ${label} element`) ``. The static-key rule applies here too, but the indefinite article makes it more than a mechanical change. An `{article}` placeholder doesn't help gendered languages, since German needs agreement (*ein Textfeld* vs *eine Checkbox*). Translating article and label as one key does, but needs two entries per field type. Dropping the article gives one clean key and matches bpmn-js (`Create start event`), but changes the English text. Flagging it so we settle the phrasing before it's changed.

### Housekeeping

- The branch conflicts with `develop`. A rebase and squash into a conventional commit would help — 66 commits and 5 merges right now.
- The description still documents `language: 'en'` and `setLanguage()`, both removed.

Thanks again for sticking with this — the shape is right, and with the above sorted I think we can get it in.

dhionson2:
Hi @Senior-Master-Merger and maintainers — I went through the remaining review feedback and prepared a clean branch rebased onto the current develop.

I kept Alexander's original contribution attributed separately and added my fixes in a second commit on top.

The remaining objective review items are addressed, including:

keeping technical drag/palette identifiers untranslated;
using static translation keys with placeholders;
complete Validator sentence keys;
preserving the OPTIONS_SOURCES_LABELS public object API;
viewer/editor coverage using custom translate.

I also fixed a max: 0 interpolation edge case where {value} could remain unresolved.

I ran an additional browser acceptance pass with 36/36 scenarios passing, including isolated instances with different translators, real drag & drop with deliberately hostile horizontal/vertical translations, palette technical search, Remove {label}, options-source translations, download interpolation and fallback behavior.

I intentionally left the PaletteEntry “Create …” wording unchanged since that was explicitly left as an open decision.

The prepared branch is here:
https://github.com/dhionson2/form-js/tree/contrib/pr-1514-public-ready

I haven't opened a competing PR against develop. Would you prefer that I open a PR against Senior-Master-Merger:translation-support, or another workflow?

AlekseyManetov:
@dhionson2 Thanks for your changes! I've created a separate PR that incorporates your and @Senior-Master-Merger's work: https://github.com/bpmn-io/form-js/pull/1549. 

@dhionson2 Could you please sign the CLA in the new PR? (See https://github.com/bpmn-io/form-js/pull/1549#issuecomment-5412653622)

I'll follow up with some additional improvements to create a community translations package, similar to what we have for bpmn-js (https://github.com/bpmn-io/bpmn-js-i18n).

I'm closing this PR in favor of #1549.

PR #1529, opened 2026-07-06
Fix drag and drop with newest preact(>10.16.0) versions

### LINKED ISSUE #760: Draggle (dragula as well) don't play nice with preact v11.16.0
### Describe the Bug

We found a bug in the web modeler integration of form-js, where a forced update to preact 11.16 (via resolutions) made certain drag and drop operations lead to components disappearing from the editor. I was able to replicate this locally as well.

firefox_k3h2hnVXmR

While forced resolutions of dependencies is always something that can go wrong, this remains a problem for us to adopt newer versions of preact in the future. 

The specific version of preact made some modifications to how children are diff'd. My guess is, that the draggle interaction with the DOM interferes with preact's new optimized diffing algo. 

### Steps to Reproduce

1. put two components in a row next to eachother
2. drag the right one to the left
3. 🪄  field disappears

### Expected Behavior

3. field should be moved properly

### Environment

Preact 11.16.0 only

AlekseyManetov:
Closes https://github.com/bpmn-io/form-js/issues/760

Dragula physically moves the dragged node in the DOM. Once the schema updates, Preact re-renders — but its 10.16+ diffing gets confused by the externally moved node and can drop the field from the DOM.

On drop we now use `drake.cancel(true)` (revert dragula's DOM change) instead of `drake.remove()`, so the DOM matches Preact's virtual DOM before the schema-driven re-render. This fix compatebility issues and allows `preact` bump (`^10.19.3`), aligning it with `@bpmn-io/properties-panel`.

Also flushes deferred auto-focus in the properties-panel tests, which became timing-sensitive under the newer Preact (behavior is fine in the real UI).

https://github.com/user-attachments/assets/3b695d24-3864-4a06-83dc-c10070f079eb


packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js
```diff
@@ -3908,6 +3922,14 @@ function findTextbox(id, container) {
   return container.querySelector(`[name=${id}] [role="textbox"]`);
 }
 
+// advances the event loop one tick inside act, flushing deferred (setTimeout)
+// work and the resulting renders before we assert
+function nextTick() {
+  return act(async () => {
```
nikku, line 3928: Odd that we need to essentially "act" twice. Need to clean this up, one day.

nikku (approved):
Good stuff. ⭐

nikku (commented):
Please add CHANGELOG update - so we don't forget on release!

PR #1535, opened 2026-07-13
feat(playground): use cm-theme highlighting in data and result editors

barinali:
Apply the shared `@bpmn-io/cm-theme` (`feelLight`) highlighting to the playground JSON data (input) and result (output) editors so all bpmn-io CodeMirror editors share one theme.

Related to camunda/form-playground#64

The screenshot from camunda/form-playground to demonstrate the change.

Without this revision
image

With this revision
image


nikku (approved):
Great stuff ⭐

PR #1538, opened 2026-07-16
fix(viewer): avoid redundant document preview fetches

### LINKED ISSUE #1520: Document preview re-fetches and blinks on every parent re-render
### Describe the Bug

The PDF and image renderers inside the Document Preview field re-fetch the document on every parent re-render. The fetched blob is swapped into an object URL, which tears down the `<embed>` / `` and reattaches it. The result is a visible blink whenever anything in the surrounding form causes a render (typing, validation, expression re-evaluation in the editor).

### Steps to Reproduce

1. Add a Document Preview field to a form pointing at a PDF.
2. Open the form in the editor or playground.
3. Type or interact with any other field that triggers form-level re-renders.
4. Observe the PDF preview blink on each render tick.

### Expected Behavior

The Document Preview should only re-fetch and remount when the document URL or the request configuration actually changes.

### Environment

- Library version: 1.21.3

### Findings

- `requestInit` is rebuilt on every render in `DocumentPreview.js` via `getDocumentRequestInit`, producing a new object reference each time.
- `onError` is passed as an inline arrow (`() => setHasError(true)`) to `PdfRenderer` / `ImageRenderer`.
- Both end up in the `useEffect([url, onError, requestInit])` dependency array that drives the `fetch`, so the effect fires on every parent render.

---

_Issue drafted with AI assistance._

oiahoon:
Closes #1520

This keeps the request configuration and renderer error callback stable when the document is unchanged. Parent rerenders therefore no longer restart PDF or image fetch effects or replace object URLs.

The focused regression forces the preview into view, rerenders the same document, and verifies both request construction and preview fetching happen once. It also verifies that changing the document still triggers a new fetch.

Try it out:

1. Add a PDF or image Document Preview with a custom endpoint builder.
2. Trigger an unrelated form rerender.
3. Confirm the preview does not blink or issue another request.

### DIFF
--- a/packages/form-js-viewer/src/render/components/form-fields/DocumentPreview.js
+++ b/packages/form-js-viewer/src/render/components/form-fields/DocumentPreview.js
@@ -4,7 +4,7 @@ import { Errors } from '../Errors';
 import { formFieldClasses } from '../Util';
 import { isString } from 'min-dash';
 import DownloadIcon from './icons/Download.svg';
-import { useEffect, useRef, useState } from 'preact/hooks';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'preact/hooks';
 import { Label } from '../Label';
 
 const type = 'documentPreview';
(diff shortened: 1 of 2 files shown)


nikku:
Thanks for the contribution. Reasonable fix for a common react / refs oversight.

nikku (approved):

nikku:
Released as `v1.24.1`

PR #1546, opened 2026-08-14
fix: wrap long checkbox, checkbox group and radio group option labels

### LINKED ISSUE #1545: Checklist item text does not wrap for long labels
### Describe the Bug

Long text in a checklist form field item does not wrap in the rendered form. The container has a fixed max width, so overflowing text gets cut off with no way to read the full item in the UI. The only way to see the full content is via the browser Inspect panel.

Image

### Steps to Reproduce

1. Create a checklist field in a form with an option label long enough to exceed the checklist item's width
2. Render the form (e.g. in Tasklist)
3. The label text is cut off instead of wrapping to a new line, and there is no way to widen the view to read it

### Expected Behavior

Long checklist item labels should wrap onto multiple lines instead of being cut off.

### Environment

- Host (Browser/Node version), if applicable: Chrome Version 151.0.7922.109
- OS: MacOS Tahoe 26.5.1
- Library version: Tasklist, cluster version8.9.14
### Context

https://camunda.slack.com/archives/C08MRKHJ0CD/p1786544091921849

barmac:
image

Closes #1545

---

### Problem

Option labels reuse the shared `.fjs-form-field-label` rule, which carries `white-space: nowrap; overflow: hidden` — truncation intended for *field* labels. Applied to checkbox / checkbox group / radio group option text there is no ellipsis and no way to widen the view, so long labels are simply unreadable. Reproducible in the viewer and in the editor.

### Fix

* `.fjs-inline-label` → `align-items: flex-start`, so the input stays next to the *first* line of a wrapping label. The class is used only by checkbox, checkbox group and radio group, so the scope is precise.
* option labels → `white-space: normal; overflow: visible; overflow-wrap: anywhere`.

`overflow-wrap: anywhere` rather than `break-word` is load-bearing: the label is a flex item whose `min-width: auto` equals its min-content size, and `break-word` does not reduce that — a long URL still overflows with `break-word` and wraps correctly with `anywhere`.

The label text and the required asterisk are now wrapped in a single `<span class="fjs-label-content">`. As separate flex items the asterisk detached to the far right of a wrapped label, which is visible on a required checkbox with a long label.

### Verification

Base and Carbon themes, rendered before / after: checked, unchecked, readonly, disabled, long sentence, long URL, and narrow columns (words break rather than clip). Viewer and editor checked against a real build.

### Notes

* :warning: **Visual regression will fail** until the snapshots are regenerated — `align-items: flex-start` shifts single-line option text a few pixels, and both `form.json` and `complex.json` contain checkbox groups and radio groups. Needs the `update-snapshots` label.
* Unrelated, pre-existing: Carbon's `.fjs-form-field-checklist .fjs-form-field-label:not(:first-of-type)` (`carbon-styles.scss:645`) no longer matches option labels — under the current markup each option label is the only `<label>` in its wrapper, so it *is* `:first-of-type`, and the Carbon option font sizing is silently not applying.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nikku (approved):

barmac:
Released in v1.25.0

PR #1549, opened 2026-08-25
Add translations support for viewer and editor

### LINKED ISSUE #1055: Request for Internationalization (i18n) Support to Facilitate Localization

AlekseyManetov:
Closes https://github.com/bpmn-io/form-js/issues/1055, https://github.com/bpmn-io/form-js/issues/1169

Supersedes https://github.com/bpmn-io/form-js/pull/1514, rebased onto current `develop`. Original work by @Senior-Master-Merger, review follow-ups by @dhionson2 — commits keep their authorship.

### Proposed Changes

**Translation support** — [d739d0a4](https://github.com/bpmn-io/form-js/commit/d739d0a46ca6e280fc060d92bb1e8aaaf25f9675), [63e7add8](https://github.com/bpmn-io/form-js/commit/63e7add860a431c6e9a14700fc1ea9e14b3e6e0e), [60a3b6bf](https://github.com/bpmn-io/form-js/commit/60a3b6bf0c1abef26546b9f71d600ea14e6ef410), [7dfd5364](https://github.com/bpmn-io/form-js/commit/7dfd5364e45531372fc5ebea78de35e1ce9370a0)

The viewer and editor register diagram-js' `TranslateModule` and route user-facing strings through `translate`. Nothing is translated by default — a consumer supplies a `translate` at module configuration time, like bpmn-js. No runtime API is added.

**`packages/form-js-i18n`** — [febf2900](https://github.com/bpmn-io/form-js/commit/febf29005cc39ed8a225715bd8399b9975494793)

A new package holding the community dictionaries, starting with English, and the tooling around them. `npm run collect-translations` runs the test suites with a `translate` that records every template it is asked for, and writes the resulting key list into the package. `npm test` there checks each dictionary against that list and reports missing or outdated entries.

**Key coverage** — [aefd7992](https://github.com/bpmn-io/form-js/commit/aefd7992a103ba08eb6b0228ee8cd6a7d08d92db)

Collection only sees what the tests execute, so this adds the tests needed for the key list to be complete.

**Palette label overflow** — [191d26eb](https://github.com/bpmn-io/form-js/commit/191d26eb2212a60966c501c9efeb67d8c19c507e)

Palette entry labels had a fixed height and no wrapping, so a label longer than the English one ran outside its tile. Labels now break, and anything that still does not fit is clipped; the entry title carries the full label.


AlekseyManetov:
## 🤖 Review: bpmn-io/form-js#1549 — Add translations support for viewer and editor

### Summary

The viewer and editor register diagram-js' `TranslateModule` and route user-facing strings through the injected `translate`. Nothing is translated by default — a consumer supplies its own `translate` at module configuration time. A new `packages/form-js-i18n` holds the English dictionary, the tooling that collects translatable keys by running the suites with a recording `translate`, and a coverage check comparing the two. 126 files, +3452 / −496.

### Verdict

**READY TO MERGE ✅**

### Definition of Done

| Criterion | Status | Notes |
|-----------|--------|-------|
| Usable / design principles | ✅ | Default English output is preserved at every composed key; opt-in, so existing embedders see no change |
| Discoverable | ✅ | `packages/form-js-i18n/README.md` documents the opt-in and the contribution flow; the documented import resolves under Node ESM |
| Accessible | ✅ | `aria-label` and `title` strings are covered, not just visible text; palette tiles no longer clip a label longer than the English one |
| Fits existing functionality | ✅ | No new public API — `translate` is supplied via `additionalModules`, matching bpmn-js |
| Simple solution | ✅ | Module-scope constants stay constant; only labels are translated at point of use |
| Understandable | ✅ | The one non-obvious rule (stringify numeric replacements) is documented where it is first needed and back-referenced elsewhere |
| Well tested | ✅ | Dedicated translation specs in both packages; `codecov/patch` 93.70% of diff hit, project 91.23% (+1.39%) |
| Clean code | ✅ | No dead code, debug output or leftovers in the diff |
| Code standards (lint / style) | ✅ | `eslint packages` clean; `tsc --pretty` clean |
| Architectural consistency | ✅ | Mirrors the bpmn-js i18n toolchain — `TranslateModule` registration, karma collector reporter, `docs/translations.json`, dictionary coverage check |
| CI passing | ✅ | Green on this tree |

### Blockers

None.

### Concerns

**The dictionary coverage check reports but never gates.** In `packages/form-js-i18n/test/test.js`, `errors` is only appended at line 97, when a dictionary fails to import; missing and unknown keys warn at lines 85 and 89 and the process still exits 0. That is the right call for community dictionaries — going red the moment someone adds a string would block the string author on a translator. The real gap is adjacent and worth its own issue: nothing verifies `docs/translations.json` is itself current, so a newly added string silently never enters the key list. A CI step running `collect-translations` and failing on a diff would close it, at the cost of running both suites twice.

### Suggestions

- **`NumberEntries.js:122` lacks the zero-case test its sibling has.** Both it and `DefaultValueEntry.js:173` stringify `decimalDigits`, but only the latter is covered at `decimalDigits: 0`. The path is reachable in `NumberEntries.js` too — `decimalDigits: 0` with input `1.5` clears the `Big(value).cmp(minimumValue)` guard on line 118 and lands on line 122. Same shape as the existing test.

- **Two message pairs differ only by a full stop**, and collection turns each into a separate translator-facing key: `Should not contain more than {value} decimal digits` (`DefaultValueEntry.js:173`) vs `… digits.` (`NumberEntries.js:122`), and `Should be a valid number` (`DefaultValueEntry.js:170`) vs `… number.` (`NumberEntries.js:107`). The inconsistency predates this change and unifying it alters user-visible English, so it belongs on its own — but it is now a maintenance cost carried by every dictionary.

- **`isValidDocument` admits an empty filename.** `DocumentPreview.js:140` checks `'fileName' in document.metadata` but not that it is non-empty, so `fileName: ''` renders `Download {fileName}` in the download button's `aria-label`. Data-driven rather than config-driven, so low priority, but a `typeof … === 'string' && length` check would close it along with the empty title it already produces.

- **Key collection hooks a private method.** `test/globals.js:13-15` in both packages wraps `Form.prototype._createInjector` / `FormEditor.prototype._createInjector`. It is the smallest hook available and the intent is commented, but a rename would silently stop collection rather than fail — the dictionary would just stop growing. A `console.warn` when the property is absent would make that loud.

- **The collector runs ahead of consumer modules**, so a spec supplying its own `translate` is excluded from collection by design; its keys have to be exercised elsewhere in the suite. This is documented in the comment, and the coverage commit compensates — worth keeping in mind when adding specs that stub `translate`.

### What's good

- **Minimal blast radius on the libraries.** The viewer and editor gain one module registration each; everything else is call sites. `diagram-js` was already a bundled dependency of both, so the import costs nothing at the package boundary.

- **Keys are static literals with placeholders throughout** — `translate('Download {fileName}', { fileName })`, never a template literal. That is what makes the key list extractable at all, and it holds across ~285 keys with no exceptions.

- **Technical identifiers stay untranslated.** Palette search still matches on `entry.type` (`Palette.js`), dragula's `horizontal`/`vertical` are untouched, and select option `value`s remain literals while only their labels go through `translate`. Getting this wrong is the classic i18n regression — searching by technical name breaking, or drag & drop breaking under a German dictionary.

- **Full sentences per key, not assembled fragments.** `Validator.js` splits the decimal-digits message into three complete keys rather than concatenating a translated stem with an English plural `s`. Languages that inflect can actually produce a correct sentence.

- **The stringify rule is applied uniformly and explained once.** `translate` falls back to the raw placeholder for a falsy replacement, so `{ value: 0 }` would render the literal `{value}` — reachable with `decimalDigits: 0` and input `1.5`. Every numeric replacement is stringified, with the rationale on `Validator.js` and a back-reference from `Table.js`. The remaining raw replacements are provably safe: guarded strings (`label || key` in `Table.js:100`, the truthy `pattern` check), literals in `EditorTable.js`, `Big` instances, and `RepeatRenderManager.js:187` where the collapse guard forces the count above the default of five.

- **The translation specs test what actually breaks in i18n**, not that a mock was called: a placeholder moved to a different position by the target language, a placeholder dropped entirely, `max: 0` interpolating rather than leaking, and two forms on one page with different dictionaries staying isolated.

- **Key collection runs the real suites**, so the list reflects what the code renders rather than what a grep can find — and the accompanying test work closes the coverage gap that approach leaves rather than shipping a knowingly partial list.

packages/form-js-editor/src/features/properties-panel/entries/RowCountEntry.js
```diff
@@ -45,36 +47,37 @@ function RowCount(props) {
...
-const validate = (value) => {
+const validate = (value, translate) => {
   if (isNil(value)) {
     return null;
   }
 
   if (!isNumber(value)) {
-    return 'Must be number';
+    return translate('Must be number');
   }
```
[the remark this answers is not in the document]
AlekseyManetov: They were absent from the collected keys because both branches are unreachable. `NumberFieldEntry` calls `onInput(target.value ? parseFloat(target.value) : undefined)`, so `validate` only ever receives a number (possibly `NaN`) or `undefined`:

- `RowCountEntry` — `isNil(value)` returns first, and `isNumber(NaN)` is `true`, so `!isNumber(value)` can never hold
- `MaxHeightEntry` — the `undefined` / `''` guard returns before `typeof value === 'string'` could be true

Reusing an existing key wouldn't have helped: collection records what the suite actually passes through `translate`, so an unreachable branch contributes nothing regardless of which key it holds.

Removed both branches instead. `NaN` still produces an error via the `Number.isInteger` check that follows, and the key list is now complete — 0 of 226 literals uncollected.

packages/form-js-editor/src/features/properties-panel/entries/MaxHeightEntry.js
```diff
@@ -38,37 +41,38 @@ function MaxHeight(props) {
...
-const validate = (value) => {
+const validate = (value, translate) => {
   if (value === undefined || value === '') {
     return null;
   }
 
   if (typeof value === 'string') {
-    return 'Value must be a number.';
+    return translate('Value must be a number.');
   }
```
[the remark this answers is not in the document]
AlekseyManetov: same as https://github.com/bpmn-io/form-js/pull/1549#discussion_r3871192871

AlekseyManetov:
> Two message pairs differ only by a full stop, and collection turns each into a separate translator-facing key: Should not contain more than {value} decimal digits (DefaultValueEntry.js:173) vs … digits. (NumberEntries.js:122), and Should be a valid number (DefaultValueEntry.js:170) vs … number. (NumberEntries.js:107). The inconsistency predates this change and unifying it alters user-visible English, so it belongs on its own — but it is now a maintenance cost carried by every dictionary.

Fixed. The rest of the suggestions don't look worth it.

packages/form-js-editor/src/core/FormLayoutValidator.js
```diff
@@ -9,23 +9,25 @@ export class FormLayoutValidator {
    *
    * @param { import('./FormLayouter').FormLayouter } formLayouter
    * @param { import('./FormFieldRegistry').FormFieldRegistry } formFieldRegistry
+   * @param { Function } translate
```
barmac: We want to use a proper type like we do in bpmn-js: https://github.com/bpmn-io/bpmn-js/blob/230692f7218e713571938ca315cd6f30b774b420/lib/features/modeling-feedback/ModelingFeedback.js#L6

packages/form-js-editor/src/features/palette/components/Palette.js
```diff
@@ -180,6 +181,7 @@ export function collectPaletteEntries(formFields) {
         group: fieldConfig.group,
         icon: fieldConfig.icon,
         iconUrl: fieldConfig.iconUrl,
+        create: (options) => fieldConfig.create(options),
```
barmac: This seems to be net-new?

packages/form-js-editor/src/features/properties-panel/entries/HeadersSourceSelectEntry.js
```diff
@@ -3,26 +3,25 @@ import { AutoFocusSelectEntry } from '../components';
...
-    static: {
-      label: translate('List of items'),
-      value: 'static',
-    },
-    expression: {
-      label: translate('Expression'),
-      value: 'expression',
-    },
-  };
+const OPTIONS = {
```
barmac, line 6: Good one

barmac (approved):
Let's squash it. Great teamwork :star:

PR #1554, opened 2026-09-04
fix(form-js-integration): resolve `@bpmn-io/form-js` from the monorepo

barmac:
Closes #

Supersedes #1553 — the macOS / Node 24 CI job failed in the integration type smoke test because `@bpmn-io/form-js-integration` resolved a published `@bpmn-io/form-js@1.24.1` tarball instead of the monorepo package.

## Why it was stale

`packages/form-js-integration` is `private`, and `@bpmn-io/release@0.4.0` leaves private packages out of the release run **entirely** — `findPackages()` does `if (pkg.private) continue`, so integration is absent from both the version-bump plan and the `repinScope` used to rewrite workspace dependency ranges.

Under lerna this worked: the last lerna release stamped it at `1.24.1`. Lerna was dropped in 9c9e1fe1, and the two releases since (v1.25.0, v1.26.0) never touched integration. Its exact pin — `"@bpmn-io/form-js": "1.24.1"`, made exact deliberately in 2c240805 — then became unsatisfiable by the `1.26.0` workspace, so npm materialised a nested published tree under `packages/form-js-integration/node_modules/`. The `tsc` smoke test stopped exercising the branch under test.

This is a different mechanism from `@bpmn-io/form-json-schema` sitting at `1.18.0`: that one is change-gated and simply hasn't been re-released.

## Changes

- `version` → `1.26.0`
- `@bpmn-io/form-js` → `^1.26.0` (caret, not exact)
- regenerated `package-lock.json`, dropping the nested tarball entries

The caret is the load-bearing part. An exact pin re-arms the same trap at v1.27.0; a caret range is satisfied by any newer workspace version, so npm links locally regardless of whether the release tooling repins it.

## Follow-up

`@bpmn-io/release@0.5.0` versions and repins private packages by default (`if (pkg.private && excludePrivate) continue`). The bump is sitting on `renovate/bpmn-io-release-0.x` with no PR opened yet — worth landing so integration's metadata moves with the monorepo instead of only being range-compatible with it.

## Verification

- `packages/form-js-integration/node_modules/` is gone; `node_modules/@bpmn-io/form-js` is a symlink to `packages/form-js`
- `grep -c 'packages/form-js-integration/node_modules' package-lock.json` → `0`
- `npm run build && npm test -w @bpmn-io/form-js-integration` → passes against local types
- `npm run lint:types` → passes

CI runs `npm run all` (`lint build lint:types test test:distro`), so the build precedes the smoke test and the now-local types are generated first.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


barmac:
Superseded by #1555 (same change, fresh branch).

PR #1559, opened 2026-09-11
fix: stop styling every button as read-only

AlekseyManetov:
Closes #

### Proposed Changes

Every form-js button currently renders grey with black text, because this rule overrides the accent and disabled styling for all of them:

```css
.fjs-container .fjs-button[type='submit']:read-only, … {
  color: var(--text-light);
  background-color: var(--color-background-readonly);
  border-color: var(--color-borders-readonly);
}
```
Introduced in 6bc96350 ("feat(viewer): support `readonly` attribute").
`:read-only` is the negation of `:read-write` — it matches every `<button>`, not just buttons in a readonly form. The rule is the last button background rule and one pseudo-class more specific than the accent rule above it, so it always wins. Its `color` also references `--text-light`, which does not exist; an invalid `var()` computes to `unset`, which for `color` means inherit, so the text does not fall back to the white set earlier.

Removing it restores what the existing rules already intend: accent for `submit`/`button`, bordered transparent for `reset`, and real disabled styling.

| | before | after |
|---|---|---|
| submit background | `rgb(241, 242, 244)` | `rgb(0, 119, 204)` |
| submit text | `rgb(0, 0, 0)` | `rgb(255, 255, 255)` |
| disabled text | `rgb(0, 0, 0)` | `rgb(129, 135, 152)` |

**Before**

image

**After**

image

### DIFF
--- a/packages/form-js-viewer/assets/form-js-base.css
+++ b/packages/form-js-viewer/assets/form-js-base.css
@@ -744,14 +744,6 @@
   border-color: var(--color-borders-disabled);
 }
 
-.fjs-container .fjs-button[type='submit']:read-only,
-.fjs-container .fjs-button[type='button']:read-only,
-.fjs-container .fjs-button[type='reset']:read-only {
-  color: var(--text-light);
-  background-color: var(--color-background-readonly);
-  border-color: var(--color-borders-readonly);
-}
-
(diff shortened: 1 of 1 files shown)


barmac (approved):
Good fix. It will have a noticable impact on the forms consumers.

PR #1560, opened 2026-09-11
feat: adopt @bpmn-io/theme

AlekseyManetov:
Related to https://github.com/bpmn-io/internal-docs/issues/1364

### Proposed Changes

form-js reads the shared `--bio-*` tokens from [@bpmn-io/theme](https://github.com/bpmn-io/shadcn-theme) instead of carrying its own colour palette. A consumer restyles form-js by binding those tokens once, rather than overriding form-js internals.

Carbon is removed from the project entirely. `@bpmn-io/form-js-carbon-styles` is deleted along with the Carbon e2e pages and the test harness that only existed to preview them — 43 files, −4864 lines. Tasklist was the only consumer and is migrating to C4, so nothing is left to keep it alive for.

Corner radii are bound the same way: `--input-border-radius`, `--button-border-radius` and `--surface-border-radius` are declared on `.fjs-container` and resolve through the `--bio-radius-*` scale, so the editor inherits them from the viewer.

### Breaking changes

The base stylesheets no longer resolve Carbon colours. Previously every colour in `form-js-base.css` deferred to Carbon with a form-js fallback, so an app that had Carbon loaded got form-js themed automatically, with no opt-in:

```css
/* form-js-base.css, before */
--color-background: var(--cds-field, var(--cds-field-01, var(--color-white)));
--color-text: var(--cds-text-primary, var(--color-grey-225-10-15));
```

```css
/* form-js-base.css, after */
--color-background: var(--bio-surface);
--color-text: var(--bio-text);
```

`@bpmn-io/form-js-carbon-styles` receives no further releases and `@bpmn-io/form-js` no longer depends on it or re-exports `./dist/assets/carbon-styles.scss`. To theme a form, bind the form-js CSS variables to your design system.

The bundled flatpickr styles are scoped to `.fjs-container` — see the date picker section below for why. Scope your own flatpickr overrides the same way, e.g. `.fjs-container .flatpickr-day`, and load them after `form-js.css`.

`--color-text-lightest` is removed, as it duplicated `--color-text-lighter`.

### What actually changes colour

Comparing `develop` against this branch in the default theme: 344 of 362 colour declarations resolve to the same value, 18 change, 4 are added, none are dropped. Every editor, palette and properties-panel variable resolves unchanged.

1. **The text scale loses a step.** form-js had four steps (15% / 35% / 45% / 55%); the theme has three. `--color-text-lightest` duplicated `--color-text-lighter`, so it is removed and the lightest step is now reserved for disabled content.

| | before | after |
| --- | --- | --- |
| field label | hsl(225, 10%, 35%) | hsl(225, 10%, 15%) |
| field description | hsl(225, 10%, 45%) | hsl(225, 10%, 35%) |
| input adornment | hsl(225, 10%, 45%) | hsl(225, 10%, 35%) |
| select placeholder / readonly display | hsl(225, 10%, 45%) | hsl(225, 10%, 35%) |
| ::placeholder, disabled options | hsl(225, 10%, 55%) | hsl(225, 10%, 35%) |
| dropdown empty state | hsl(225, 10%, 55%) | hsl(225, 10%, 35%) |
| disabled inputs, disabled buttons, empty table | hsl(225, 10%, 35%) | hsl(225, 10%, 55%) |

2. **Input borders follow the theme.** `--color-borders` goes from `hsl(225, 10%, 55%)` to `hsl(225, 10%, 75%)`, the border the Modeler's own forms already use. The most visible change — every input, select, textarea and taglist border gets lighter.

3. **Icons stop being pure black.** `--color-icon-base` and `--color-icon-inverted` were literally `--color-black`; the theme has no pure black, so the datetime adornment, repeat-row remove and taglist remove icons go from `rgb(0, 0, 0)` to `hsl(225, 10%, 15%)`.

4. **Shadows use a translucent colour.** `--color-shadow` fell back to `hsl(225, 10%, 85%)` — an opaque light grey sitting in a shadow slot. An opaque colour cannot fade, so the dropdown rendered a flat grey halo rather than a shadow.

| | before | after |
| --- | --- | --- |
| dropdown list | rgb(213, 215, 221) | rgba(0, 0, 0, 0.3) |
| drag mirror (.gu-mirror) | rgba(0, 0, 0, 0.15) | rgba(0, 0, 0, 0.3) |

5. **Focus and validation.** The ring widens to 2px to carry a translucent colour, and the focused control now also takes the focus border. Because the invalid state only overrides `outline-color`, it inherits that width — a solid red at 2px reads as a slab next to the 1px red border, so it takes the danger ring instead.

| | before | after |
| --- | --- | --- |
| --outline-definition | 1px solid hsl(225, 10%, 55%) | 2px solid hsla(205, 100%, 40%, 0.35) |
| focused control border | (unset) | hsl(205, 100%, 40%) |
| invalid outline | rgb(230, 0, 0) | rgba(230, 0, 0, 0.35) |

### Corner radii

form-js mixed `2px` and `3px` with no rule behind which was which — inputs were `3px` while the buttons sharing their layout rule were `2px`. The radii now resolve per element kind:

| | before | after |
| --- | --- | --- |
| inputs, selects, textareas, input groups, adornments, taglist | 3px | 2px |
| .fjs-button[type=submit\|button\|reset] | 2px | 3px |
| .fjs-taglist-tag, .fjs-repeat-row-remove | 2px | 3px |
| .fjs-dropdownlist, .fjs-table-middle-container | 3px | 4px |

One rule groups the text inputs together with the submit, button and reset variants, so the buttons used to inherit the input radius. They now take theirs from the button-only rule further down.

### The date picker

`assets/index.scss` vendors `flatpickr/dist/themes/light.css`, which hard-codes the picker's palette across 41 selectors. `form-js-base.css` overrode four day states (`today`, `selected`, and their hover/focus) and nothing else — the calendar surface, month header, nav arrows, weekdays, adjacent and disabled days, ranges, the time input and the spinner arrows were all literals.

That went unnoticed because Carbon's DatePicker is also built on flatpickr: `@carbon/styles` ships 75 flatpickr rules that covered the vendored theme in any Carbon app. With Carbon gone, the literals show through — the picker was unreadable in a dark theme and ignored the host palette in a light one.

A `--color-datepicker-*` family now covers the whole picker and binds to `--bio-*`:

```css
--color-datepicker-background: var(--bio-surface);
--color-datepicker-border: var(--bio-border);
--color-datepicker-header-background: var(--bio-surface-medium);
--color-datepicker-header-color: var(--bio-text-subtle);
--color-datepicker-day-color: var(--bio-text);
--color-datepicker-day-hover-background: var(--bio-surface-strong);
--color-datepicker-day-disabled-color: var(--bio-text-subtlest);
--color-datepicker-range-background: var(--bio-surface-medium);
--color-datepicker-focused-day: var(--bio-text-subtlest);
```

The vendored theme is also scoped, because its selectors were bare (`.flatpickr-day`) and so are Carbon's — equal specificity, so whichever stylesheet loaded last won. In an app with Carbon loaded that produced a mixed picker: form-js colours with Carbon's geometry. All 236 flatpickr rules now carry `.fjs-container`, and none are emitted unscoped:

```scss
// index.scss
.fjs-container {
  @include meta.load-css('flatpickr/dist/themes/light.css');
}
```

It is emitted before `form-js-base.css` so the theming above still wins the tie. The calendar uses `static: true`, so it always renders inside `.fjs-container` and the scope always matches.

### Other fixes
- **Long lines in code blocks wrap.** `pre` set no `white-space`, and neither do the text and html fields, so a long line escaped the field instead of reflowing.
- **The table focus outline stays inside the cell.** The container radius clipped the square outline at the corner, and `border-collapse: collapse` disabled the cell radius that would have followed it.
- **The bpmn.io attribution stops deferring to Carbon.** It resolved `--cds-*` directly; both Camunda products hide it with `display: none`, so this affects neither.
- **Icons are centred independent of their display value.**
- **The playground no longer relies on user agent styling.**

### Impact on consumers

| | affected | why |
| --- | --- | --- |
| Tasklist | yes | the only consumer of `@bpmn-io/form-js-carbon-styles`; covered by the migration to C4 |
| Camunda Modeler & Web Modeler  | yes | imports `@carbon/styles` globally with no `--cds-*` reset, so its form editor resolved Carbon today and now follows the bpmn.io theme |

Neither Modeler imports `@bpmn-io/form-js-carbon-styles`, so removing the package breaks no build.

### Documentation follow-up

[Form styling](https://docs.camunda.io/docs/apis-tools/frontend-development/forms/customize-and-extend/form-styling/) documents the old contract and goes stale with this PR. Source: `docs/apis-tools/frontend-development/03-forms/03-customize-and-extend/01-styling.md`, plus the same file under `versioned_docs/version-8.7`, `8.8` and `8.9`.

Three things need updating:

- **The CSS variables listing** (`### CSS variables`) prints every variable in its old Carbon-deferring form, e.g. `--color-background: var(--cds-field, var(--cds-field-01, var(--color-white)));`. All of these now resolve through `--bio-*`.
- **`--color-text-lightest` is documented but removed** — it duplicated `--color-text-lighter`.
- **The `### Example` section links to `packages/form-js-carbon-styles`**, a directory this PR deletes, so the link 404s once merged. The surrounding claim that "forms rendered in Tasklist appear in this design system by default" also stops being true — theming is now explicit, via the `--bio-*` tokens.

The page should instead describe binding the `--bio-*` tokens once, which is the same mechanism every bpmn.io editor now uses.

### Downstream: `@camunda/form-playground`

Both Modelers import the form CSS from `@camunda/form-playground`, not from `@bpmn-io/form-js`:

```scss
@import '~@camunda/form-playground/dist/assets/form-js.css';
```

form-playground copies form-js's assets into its own `dist` at **its** build time:

```js
copy({ targets: [{ src: 'node_modules/@bpmn-io/form-js/dist/assets/**/*', dest: 'dist/assets' }] })
```

So bumping `@bpmn-io/form-js` in an app updates the JS — it is a peer dependency, resolved from the app's tree — but leaves the CSS frozen at whatever form-playground shipped. form-playground has to be rebuilt and released against this version before either Modeler can pick it up. Its peer range is `"@bpmn-io/form-js": "^1.10.3"` and needs widening to admit `2.x`.

### Steps to try out

```bash
npm install
npm start
```

To check it inside the Camunda Modeler form editor — the consumer whose appearance changes — install the alpha in `camunda-modeler/client`:

```bash
npm install @bpmn-io/form-js@2.0.0-alpha.0
```


packages/form-js-viewer/assets/form-js-base.css
```diff
@@ -1,81 +1,76 @@
...
-  --color-text-light: var(--cds-text-secondary, var(--color-grey-225-10-35));
-  --color-text-lighter: var(--cds-text-secondary, var(--color-grey-225-10-45));
-  --color-text-lightest: var(--cds-text-placeholder, var(--color-grey-225-10-55));
-  --color-text-inverted: var(--cds-text-inverse, var(--color-text));
-  --color-text-disabled: var(--cds-text-disabled, var(--color-text-light));
-
-  --color-borders: var(--cds-border-strong, var(--cds-border-strong-01, var(--color-grey-225-10-55)));
-  --color-borders-group: var(--cds-border-subtle, var(--color-grey-225-10-85));
+
+  --color-background: var(--bio-surface);
```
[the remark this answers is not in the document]
AlekseyManetov: Not doing this — `src/carbon-styles.js` is deprecated and scheduled for removal, so mirroring the variable block there would add a second copy of the Carbon token mapping to hand-sync on every future change, in a file we intend to delete.

The legacy export isn't losing Carbon wholesale: it still paints the components it restyles with `--cds-*` directly. What it loses are base variables it never declared and inherited from the base stylesheet resolving `--cds-*` — the implicit coupling this PR removes.

Documented instead: the Breaking Changes entry now states the mapping lives in `src/carbon-styles.scss` and that the styled-components export no longer themes a form on its own. If we'd rather not ship a half-Carbon deprecated export, I'd support deleting it here instead of reviving it.

nikku:
Note: this PR removes Carbon theming from the core stylesheets, but a couple of Carbon touchpoints remain in core packages:

* `packages/form-js-viewer/src/render/components/PoweredBy.js` still defers to `var(--cds-icon-primary, ...)` / `var(--cds-text-primary, ...)` for the logo and caption colors.
* `@bpmn-io/form-js-viewer` still depends on `@carbon/grid` (`assets/grid.scss` does `@use '@carbon/grid'`).

Since we are already shipping a breaking change here, we may as well fully remove Carbon from the core in the same release line rather than carrying these leftovers forward.

nikku:
Same for the umbrella package: `@bpmn-io/form-js` still depends on `@bpmn-io/form-js-carbon-styles` and re-exports `dist/assets/carbon-styles.scss`.

Carbon should be *optional* — installed and imported by the consumer, not shipped with core. Drop the dependency and the re-export; consumers import the stylesheet from `@bpmn-io/form-js-carbon-styles` directly.

AlekseyManetov:
> Note: this PR removes Carbon theming from the core stylesheets, but a couple of Carbon touchpoints remain in core packages:
> 
> * `packages/form-js-viewer/src/render/components/PoweredBy.js` still defers to `var(--cds-icon-primary, ...)` / `var(--cds-text-primary, ...)` for the logo and caption colors.
> * `@bpmn-io/form-js-viewer` still depends on `@carbon/grid` (`assets/grid.scss` does `@use '@carbon/grid'`).
> 
> Since we are already shipping a breaking change here, we may as well fully remove Carbon from the core in the same release line rather than carrying these leftovers forward.

`PoweredBy.js` is fixed.

@carbon/grid is the layout engine — it emits only layout rules (widths, margins, breakpoints), no visual styling. It's compiled into `form-js.css` at build time, so it's self-contained and doesn't need the rest of Carbon present.

I'd keep it for now. Replacing it is a bigger rework — we'd need to find or build a replacement grid — and since it carries no visual styles it doesn't conflict with making the theming Carbon-free.

> Same for the umbrella package: `@bpmn-io/form-js` still depends on `@bpmn-io/form-js-carbon-styles` and re-exports `dist/assets/carbon-styles.scss`.
> 
> Carbon should be _optional_ — installed and imported by the consumer, not shipped with core. Drop the dependency and the re-export; consumers import the stylesheet from `@bpmn-io/form-js-carbon-styles` directly.

Fixed.

packages/form-js-editor/package.json
barmac, line 1: This one is `deps` as it's a production dep.

packages/form-js/CHANGELOG.md
```diff
@@ -6,8 +6,17 @@ All notable changes to [form-js](https://github.com/bpmn-io/form-js) are documen
...
+* `FEAT`: move the Carbon color mapping into `@bpmn-io/form-js-carbon-styles` ([#1560](https://github.com/bpmn-io/form-js/pull/1560))
 * `FIX`: align option labels with their checkbox or radio ([#1561](https://github.com/bpmn-io/form-js/pull/1561))
 * `FIX`: Prevent styling every button as read-only, which resulted in submit buttons looking like primary instead of gray ([#1559](https://github.com/bpmn-io/form-js/pull/1559))
+* `DEPS`: update to `@bpmn-io/properties-panel@3.55.0`
+
+### Breaking Changes
+
+- The base stylesheets no longer resolve Carbon colors from `--cds-*`. Loading `form-js-base.css` in a Carbon application no longer themes the form on its own — import `@bpmn-io/form-js-carbon-styles`, which now carries the color mapping next to the component styles it already provided. The color mapping lives in `src/carbon-styles.scss`; the deprecated `src/carbon-styles.js` styled-components export does not carry it and no longer themes a form on its own.
+- The `fjs-no-theme` selector is removed. It reset the `--cds-*` variables the base stylesheets no longer reference; not loading `@bpmn-io/form-js-carbon-styles` is now the way to opt out of Carbon, and it is the default.
```
barmac:
```suggestion
```

packages/form-js/CHANGELOG.md
```diff
@@ -6,8 +6,17 @@ All notable changes to [form-js](https://github.com/bpmn-io/form-js) are documen
...
 _**Note:** Yet to be released changes appear here._
 
+* `FEAT`: source color variables from `@bpmn-io/theme` ([#1560](https://github.com/bpmn-io/form-js/pull/1560))
 * `FIX`: align option labels with their checkbox or radio ([#1561](https://github.com/bpmn-io/form-js/pull/1561))
 * `FIX`: Prevent styling every button as read-only, which resulted in submit buttons looking like primary instead of gray ([#1559](https://github.com/bpmn-io/form-js/pull/1559))
+* `DEPS`: update to `@bpmn-io/properties-panel@3.55.0`
+
+### Breaking Changes
+
+- The base stylesheets no longer resolve Carbon colors from `--cds-*`. Loading them in a Carbon application no longer themes the form to match Carbon.
```
barmac: Let's have a migration hint in here. Something like:

```
Override the bpmn.io theme semantic tokens to use Carbon colors. [How to theme a bpmn.io library](link to a guide https://github.com/bpmn-io/themes).
```

barmac (approved):
The change itself works correctly. I still think we can frame the breaking changes better: https://github.com/bpmn-io/form-js/pull/1560/changes#r4045791483

PR #1561, opened 2026-09-14
fix: align option labels with their checkbox or radio

AlekseyManetov:

### Proposed Changes

Checkbox, checklist and radio option labels sit above their control.

A regression from [#1545](https://github.com/bpmn-io/form-js/pull/1545), released in 1.25.0, which changed the option row from `align-items: center` to `align-items: flex-start` so a wrapping label keeps the input on its first line. The input's `margin: 6px 10px 6px 4px` used to be spacing only — with `flex-start` the top margin became its position as well, and the resulting margin box is taller than the label's line box.

The control's layout is unchanged; the label follows it down instead.

`form-js-carbon-styles` resets the offset: Carbon replaces the native control with its own, sized to match the label's line box, so it is already aligned there.

**Before**

image

**After**

image

image


barmac (approved):
Approved but there is a conflict to resolve

packages/form-js-viewer/assets/form-js-base.css
```diff
@@ -290,6 +290,12 @@
   display: flex;
 }
 
+.fjs-container .fjs-form-field-checkbox .fjs-inline-label > .fjs-form-field-label,
+.fjs-container .fjs-form-field-checklist .fjs-inline-label > .fjs-form-field-label,
+.fjs-container .fjs-form-field-radio .fjs-inline-label > .fjs-form-field-label {
```
barmac: Do we need so much specificity?

AlekseyManetov: Simplified to `.fjs-container .fjs-inline-label > .fjs-form-field-label`

PR #1562, opened 2026-09-14
fix: fix security issue in DocumentPreview.js

anupamme:
## Summary
Fix high severity security issue in `packages/form-js-viewer/src/render/components/form-fields/DocumentPreview.js`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | HIGH |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `packages/form-js-viewer/src/render/components/form-fields/DocumentPreview.js:115` |
| **Assessment** | Likely exploitable |

**Description**: The DocumentPreview component fetches documents using URLs that can be controlled through document metadata. The isValidDocumentEndpoint function only validates URL syntax using URL.canParse() without restricting internal IP addresses, cloud metadata endpoints (169.254.169.254), or private network ranges. An attacker who controls the document endpoint URL can make the application fetch internal resources.

## Evidence

**Exploitation scenario**: An attacker who controls document metadata sets the endpoint to http://169.254.169.254/latest/meta-data/iam/security-credentials/ to access cloud metadata, or http://localhost:8080/internal-api/ to.

**Scanner confirmation**: multi_agent_ai rule `V-001` flagged this pattern.

**Production code**: This file is in the production codebase, not test-only code.

## Changes
- `packages/form-js-viewer/src/render/components/form-fields/DocumentPreview.js`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


barmac:
Hi,

I looked into this automated PR. First, please notice that the bpmn-io org [has a security policy](https://github.com/bpmn-io/form-js/security/policy), so please don't open pull requests before you contact the team. Otherwise, you basically disclose a vulnerability before anybody has a chance to fix it.

However, I believe there's no vulnerability in this case. The assumption made in the PR is that an attacker controls the form that a user opens.

> An attacker who controls the document endpoint URL can make the application fetch internal resources.

All of the code for the form viewer is ran on the client side, including the document preview component. So the _internal resources_ would be the ones that a user has access to. The document preview component will display an image or a PDF, but it does not execute any code from the endpoint. Since the code is executed in the browser, there's also no question of SSRF present in the branch name. I fail to see any vulnerability in this case.

The story would be completely different if we executed any code fetched from an endpoint, but that is not the case.
Alternatively, we'd have a SSRF vulnerability in case the code were executed on the server side, but it's not.

anupamme:
Thanks for taking the time to review this. You’re right about the execution model: DocumentPreview performs the request in the browser, so my original SSRF classification was incorrect. I was treating the URL-controlled fetch as a server-side request, which it is not.

I withdraw the security finding.

The Copilot review also highlighted that such a blacklist would not be a reliable SSRF boundary in the browser anyway (redirects, IPv6, DNS resolution, etc.), which reinforces your point.

barmac:
Hi, this PR triggered a thorough review of the DocumentPreview component which allowed to identify and fix an actual vulnerability: https://github.com/bpmn-io/form-js/security/advisories/GHSA-763f-qrmm-5w8g. Thanks for a nudge :)

